PR #19, opened 2023-02-02
Initial Allocator Impl

briandowns:
Signed-off-by: Brian Downs <brian.downs@gmail.com>


pkg/controller/cluster/cidr_allocation.go
```diff
@@ -0,0 +1,118 @@
...
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/galal-hussein/k3k/pkg/apis/k3k.io/v1alpha1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	cidrAllocationClusterPoolName = "k3k-cidr-allocation-cluster-pool"
```
galal-hussein, line 14: can we slightly change the name to match the k3s flags: cluster-cidr and service-cidr instead of cidr-allocation-cluster

galal-hussein (approved):

Issue #41, opened 2023-06-15
Persist k3k clusters state


galal-hussein:
The k3k clusters should have a persistent feature to enable the persistent of cluster store state on disk, the cluster state here would include:

- Database
- server dir
- agent dir

PR #55, opened 2023-07-18
Add cluster persistence with statefulsets

galal-hussein:
The PR adds the following:

- New fields to clutster spec:
```
                persistence:
                  type: object
                  properties:
                    type:
                      type: string
                      default: "ephermal"
                    storageClassName:
                      type: string
                    storageRequestSize:
                      type: string
```
- A feature to use dynamic storage provisioning using storatge class for both server and agent pods for the following dirs:
```
/var/lib/rancher/k3s
/var/lib/kubelet
```
If "dynamic" type is chosen for the `spec.persistence.type` then k3k will run the servers and agents as statefulsets

- https://github.com/rancher/k3k/issues/41


pkg/controller/cluster/agent/agent.go
```diff
@@ -38,15 +39,91 @@ func Agent(cluster *v1alpha1.Cluster) *apps.Deployment {
...
 			},
 		},
 	}
 }
 
-func agentPodSpec(image, name string, args []string) v1.PodSpec {
+func StatefulAgent(cluster *v1alpha1.Cluster) *apps.StatefulSet {
+	image := util.K3SImage(cluster)
+
+	const name = "k3k-agent"
```
briandowns, line 51: Would this be useful at package level anywhere else?

pkg/controller/cluster/agent/service.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/rancher/k3k/pkg/apis/k3k.io/v1alpha1"
+	"github.com/rancher/k3k/pkg/controller/util"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func StatefulAgentService(cluster *v1alpha1.Cluster) *v1.Service {
+	name := "k3k-agent"
```
briandowns, line 11: Looks like this can be declared at package level and used above and here.

pkg/controller/persist/persist.go
```diff
@@ -0,0 +1,6 @@
+package persist
```
briandowns, line 1: These values look to only be used in the cluster package. They don't need their own package and should be moved to cluster.go unexported.

pkg/apis/k3k.io/v1alpha1/types.go
```diff
@@ -40,6 +41,13 @@ type ClusterList struct {
 	Items []Cluster `json:"items"`
 }
 
+type PersistenceConfig struct {
+	// Type can be ephermal, static, dynamic
+	Type               string `json:"type"`
```
briandowns, line 46: It might be nice to attach a `.validate() error` method to this type to check if the given value in the `Type` field is a valid one and to short circuit if it's not.

briandowns (approved):

PR #68, opened 2023-11-30
[controller] HA stabilization and fix rejoining ephermal nodes

galal-hussein:
- Remove etcd member when server pod gets deleted only in ephermal HA mode
- Fix recursion bug that has been happening due to re-updating cluster spec


pkg/apis/k3k.io/v1alpha1/types.go
```diff
@@ -74,7 +74,9 @@ type NodePortConfig struct {
...
 
 type ClusterStatus struct {
-	ClusterCIDR string `json:"clusterCIDR,omitempty"`
-	ServiceCIDR string `json:"serviceCIDR,omitempty"`
-	ClusterDNS  string `json:"clusterDNS,omitempty"`
+	ClusterCIDR string             `json:"clusterCIDR,omitempty"`
+	ServiceCIDR string             `json:"serviceCIDR,omitempty"`
+	ClusterDNS  string             `json:"clusterDNS,omitempty"`
+	TLSSANs     []string           `json:"tlsSANs,omitempty"`
+	Persistence *PersistenceConfig `json:"persistence,omitempty"`
```
briandowns, line 81: I wish we had a config package. :/

pkg/controller/cluster/cluster.go
```diff
@@ -86,15 +141,34 @@ func (c *ClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request
...
+	listOpts := &client.ListOptions{Namespace: util.ClusterNamespace(&cluster)}
+	matchingLabels.ApplyToList(listOpts)
 
-		return reconcile.Result{}, c.createCluster(ctx, &cluster)
+	if err := c.Client.List(ctx, &podList, listOpts); err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+	for _, pod := range podList.Items {
+		if controllerutil.ContainsFinalizer(&pod, etcdPodFinalizerName) {
+			controllerutil.RemoveFinalizer(&pod, etcdPodFinalizerName)
```
briandowns, line 160: Is there an error check supposed to happen here?

pkg/controller/cluster/cluster.go
```diff
@@ -105,10 +179,11 @@ func (c *ClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request
...
 
-	if cluster.Spec.Persistence == nil {
-		// default to ephermal nodes
-		cluster.Spec.Persistence = &v1alpha1.PersistenceConfig{
-			Type: server.EphermalNodesType,
+	if cluster.Spec.Persistence != nil {
+		cluster.Status.Persistence = cluster.Spec.Persistence
+		if cluster.Spec.Persistence.StorageRequestSize == "" {
+			// default to 1G of request size
+			cluster.Status.Persistence.StorageRequestSize = "1G"
```
briandowns, line 186: Is there a constant defined anywhere for this value?

pkg/controller/cluster/cluster.go
galal-hussein: no not really, the removeFinalizer and AddFinalizer both return a boolean that will only suggests if the finalizer has being deleted, but it will be redundant since I added ContainsFinalizer() check

briandowns (approved):

Issue #71, opened 2023-12-07
Controller fails to create ingress


maxsokolovsky:
When I create a cluster whose `expose` option is `ingress`, I pass it the `ingressClassName`. But the controller fails to create the ingress.
It cannot get the node address from the node object.

galal-hussein:
Tested with: `v0.1.1-rc2`

Creating a cluster object with the name "system" will result in the following error message:
```
I0105 23:43:51.014334       1 cluster.go:145] enqueue cluster [system]
E0105 23:43:51.014341       1 cluster.go:184] Invalid cluster name system, no action will be taken
```

using the k3kcli with create cluster with name system will result in:
```
k3kcli-amd64 cluster create  --servers 1 --agents 0 --name system --token test
FATA[0000] invalid cluster name
``

Issue #70, opened 2023-12-06
Controller panics on cluster create


maxsokolovsky:
The nil pointer dereference occurs on cluster create: https://github.com/rancher/k3k/blob/dd618e580ae635c75be8f5bee7923b4f436747e4/pkg/controller/cluster/cluster.go#L144

Earlier in the method, the controller checks if spec.persistence is nil and sets it to a non-nil value if yes. However, the call to update the object somehow resets the value back to nil: https://github.com/rancher/k3k/blob/dd618e580ae635c75be8f5bee7923b4f436747e4/pkg/controller/cluster/cluster.go#L114

galal-hussein:
Fixed in v0.1.1-rc1

PR #80, opened 2024-01-03
Fix append to empty slice

galal-hussein:
- Fix the initialization of the addresses slice so that it doesn't add an empty item
- FIx a panic that was happening for the persistence type

- https://github.com/rancher/k3k/issues/71
- https://github.com/rancher/k3k/issues/70

### DIFF
--- a/pkg/controller/util/util.go
+++ b/pkg/controller/util/util.go
@@ -58,7 +58,7 @@ func Addresses(ctx context.Context, client client.Client) ([]string, error) {
 		return nil, err
 	}
 
-	addresses := []string{}
+	var addresses []string
 	for _, node := range nodeList.Items {
 		addresses = append(addresses, nodeAddress(&node))
 	}
--- a/pkg/controller/util/util.go
+++ b/pkg/controller/util/util.go
@@ -58,8 +58,7 @@ func Addresses(ctx context.Context, client client.Client) ([]string, error) {
 		return nil, err
 	}
 
-	addresses := make([]string, len(nodeList.Items))
-
+	addresses := []string{}
 	for _, node := range nodeList.Items {
 		addresses = append(addresses, nodeAddress(&node))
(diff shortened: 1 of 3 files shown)


pkg/controller/util/util.go
```diff
@@ -58,8 +58,7 @@ func Addresses(ctx context.Context, client client.Client) ([]string, error) {
 		return nil, err
 	}
 
-	addresses := make([]string, len(nodeList.Items))
-
+	addresses := []string{}
```
briandowns, line 61: Why are you doing this?

galal-hussein: so basically when you initialize the addresses with one "empty" item, and then append to it we will end up with a slice with two items, one of them is empty, I just figured this should be easier

briandowns (approved):

pkg/controller/util/util.go
```diff
@@ -58,8 +58,7 @@ func Addresses(ctx context.Context, client client.Client) ([]string, error) {
 		return nil, err
 	}
 
-	addresses := make([]string, len(nodeList.Items))
-
+	var addresses []string
```
maxsokolovsky, line 61: You can still pre-allocate the memory for the slice knowing its capacity and then using `append` safely.

`addresses := make([]string, 0, len(nodeList.Items))` rather than

`addresses := make([]string, len(nodeList.Items))`.

In the first example, the size would be zero, so there won't be empty elements when you `append` to the slice. And that code would be more efficient because the runtime won't need to keep copying the data when it needs to grow the slice, since you'd already pre-allocated the memory by specifying the capacity via the third argument to `make`.

maxsokolovsky: So the only change you need to fix this is to add a 0 as the second argument to `make` and keep `len(nodeList.Items)` as the third.

Issue #69, opened 2023-12-06
Controller allows cluster deployment in the k3k-system namespace if the cluster name is "system"


maxsokolovsky:
If I create a cluster named "system", the k3k controller tries to create a namespace called "k3k-system", which succeeds whether it already exists or not. The cluster is deployed in this special k3k-system namespace. The problem is that when the chart if uninstalled, Helm will also take down the namespace with the k3k cluster with it, which may be undesirable.

A possible solution may be to not allow a k3k cluster whose name is "system" or change the namespace naming scheme

galal-hussein:
Tested with: `v0.1.1-rc2`

Creating a cluster object with the name "system" will result in the following error message:
```
I0105 23:43:51.014334       1 cluster.go:145] enqueue cluster [system]
E0105 23:43:51.014341       1 cluster.go:184] Invalid cluster name system, no action will be taken
```

using the k3kcli with create cluster with name system will result in:
```
k3kcli-amd64 cluster create  --servers 1 --agents 0 --name system --token test
FATA[0000] invalid cluster name
``

PR #81, opened 2024-01-05
Add validation for system cluster name for both controller and cli

galal-hussein:
- https://github.com/rancher/k3k/issues/69

### DIFF
--- a/cli/cmds/cluster/create.go
+++ b/cli/cmds/cluster/create.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/rancher/k3k/cli/cmds"
 	"github.com/rancher/k3k/pkg/apis/k3k.io/v1alpha1"
+	"github.com/rancher/k3k/pkg/controller/cluster"
 	"github.com/rancher/k3k/pkg/controller/cluster/server"
 	"github.com/rancher/k3k/pkg/controller/util"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


cli/cmds/cluster/create.go
```diff
@@ -210,6 +210,9 @@ func validateCreateFlags(clx *cli.Context) error {
 	if name == "" {
 		return errors.New("empty cluster name")
 	}
+	if name == "system" {
```
briandowns, line 213:
```suggestion
	if name == clusterInvalidName {
```

pkg/controller/cluster/cluster.go
```diff
@@ -178,6 +179,11 @@ func (c *ClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request
 }
 
 func (c *ClusterReconciler) createCluster(ctx context.Context, cluster *v1alpha1.Cluster) error {
+
```
briandowns, line 182:
```suggestion
```

pkg/controller/cluster/cluster.go
```diff
@@ -178,6 +179,11 @@ func (c *ClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request
...
+	if cluster.Name == clusterInvalidName {
+		klog.Errorf("Invalid cluster name %s, no action will be taken", cluster.Name)
```
briandowns, line 184: It's fine to return nil here?

galal-hussein: yes I don't want the controller to reconcile again, just show an error and stop processing this cluster object, in the future, we need to add the status of the cluster in the cluster object as well

briandowns (approved):

Issue #83, opened 2024-01-08
[cli] Add kubeconfig subcommand


galal-hussein:
k3kcli should have a kubeconfig subcommand which should be able to generate kubeconfig on demand for any cluster created in the system

galal-hussein:
Tested with v0.2.0-rc1, generate with custom config works fine as well as default configuration

PR #87, opened 2024-01-09
Add kubeconfig subcommand

galal-hussein:
Adding kubeconfig subcommand, that will contain the following:

- `kubeconfig generate`: which will generate a custom defined kubeconfig for a specific cluster, currently it only uses cert client authentication since k3s is using that by default, and you can customize the following options:
  - expiration date: by default certificate will last 365
  - altNames: any alt names that needed to be added to the cert
  - common name: by default it uses admin common name `"system:admin"`
  - org: by default it sets `"system:masters"`

This way the user can create a kubeconfig for the k3k clusters with a custom CN and ORG that corresponds to certain privileges for said users.

**Core Changes**

This pr will change the behavior of the controller from creating "<cluster>-kubeconfig" secret to "<cluster>-bootstrap" secret which will contain the bootstrap data for the k3k cluster created.

- https://github.com/rancher/k3k/issues/83


pkg/controller/util/util.go
```diff
@@ -65,3 +86,144 @@ func Addresses(ctx context.Context, client client.Client) ([]string, error) {
 
 	return addresses, nil
 }
+
+func ExtractKubeconfig(ctx context.Context, client client.Client, cluster *v1alpha1.Cluster, hostServerIP string, cfg *KubeConfig) ([]byte, error) {
```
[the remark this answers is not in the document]
galal-hussein: will do, thanks!

cli/cmds/kubeconfig/generate.go
```diff
@@ -0,0 +1,147 @@
...
+func generateKubeconfig(clx *cli.Context) error {
+	var cluster v1alpha1.Cluster
+	ctx := context.Background()
+
+	restConfig, err := clientcmd.BuildConfigFromFlags("", cmds.Kubeconfig)
+	if err != nil {
+		return err
+	}
+
+	ctrlClient, err := client.New(restConfig, client.Options{
```
briandowns, line 92: Where is this error checked?

cli/cmds/kubeconfig/generate.go
```diff
@@ -0,0 +1,147 @@
...
+		AltNames:   certAltNames,
+	}
+	var kubeconfig []byte
+	if err := retry.OnError(backoff, apierrors.IsNotFound, func() error {
+		kubeconfig, err = cfg.Extract(ctx, ctrlClient, &cluster, host[0])
+		if err != nil {
+			logrus.Infof("waiting for cluster to be available: %v", err)
+			return err
+		}
+		return nil
```
briandowns, line 127: Looks like the an error value would be lost if the backoff is reached.

cli/cmds/kubeconfig/generate.go
```diff
@@ -0,0 +1,147 @@
...
+	pwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	if configName == "" {
+		configName = cluster.Name + "-kubeconfig.yaml"
+	}
+
+	logrus.Infof(`You can start using the cluster with:
```
briandowns, line 140: Does this get formatted and look alright being multiline and a log message?

cli/cmds/kubeconfig/generate.go
galal-hussein: I don't think so, this return belongs to the internal function which is passed to the retry function, it will only return nil if there is no error for each retrial, at the end if the backoff steps (5) are reached then the last value will be err and the program will exit

cli/cmds/kubeconfig/generate.go
galal-hussein: it does, yes, it looks like this at the end:
```
# k3kcli-amd64 kubeconfig generate --name demo2 --config-name test2.kubeconfig 
INFO[0001] certificate CN=system:admin,O=system:masters signed by CN=k3s-client-ca@1704915126: notBefore=2024-01-10 19:32:06 +0000 UTC notAfter=2024-12-31 19:36:15 +0000 UTC 
INFO[0001] You can start using the cluster with: 

	export KUBECONFIG=/home/hussein/shed/go/src/github.com/rancher/k3k/test2.kubeconfig
	kubectl cluster-info
	 
```

briandowns (approved):

PR #89, opened 2024-01-12
Cluster type validations

galal-hussein:
preventing the change of certain spec in the cluster type after the initial creation, that includes:
- ClusterCIDR
- ServiceCIDR
- Token
- ClusterDNS

### DIFF
--- a/pkg/controller/cluster/cluster.go
+++ b/pkg/controller/cluster/cluster.go
@@ -577,16 +577,16 @@ func (c *ClusterReconciler) getETCDTLS(cluster *v1alpha1.Cluster) (*tls.Config, 
 
 func (c *ClusterReconciler) validate(cluster *v1alpha1.Cluster) error {
 	if cluster.Name == ClusterInvalidName {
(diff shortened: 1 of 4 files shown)


pkg/controller/cluster/cluster.go
```diff
@@ -574,3 +574,19 @@ func (c *ClusterReconciler) getETCDTLS(cluster *v1alpha1.Cluster) (*tls.Config,
 		Certificates: []tls.Certificate{clientCert},
 	}, nil
 }
+
+func (c *ClusterReconciler) validate(cluster *v1alpha1.Cluster) error {
+	if cluster.Name == ClusterInvalidName {
+		return errors.New("Invalid cluster name " + cluster.Name + " no action will be taken")
```
briandowns, line 580: Error strings shouldn't be capitalized.

galal-hussein: fixing now

briandowns (approved):

PR #104, opened 2024-05-13
Fix code generation

galal-hussein:


MbolotSuse (approved):
LGTM thanks for fixing the codegen.

PR #105, opened 2024-05-15
Move to Github Action

galal-hussein:


ops/index-chart
```diff
@@ -18,7 +18,7 @@ cr index  --token ${GITHUB_TOKEN} \
           --package-path ./deploy/ \
           --index-path index.yaml \
           --git-repo k3k \
-          -o rancher
+          -o galal-hussein
```
brandond, line 21: is this correct?

galal-hussein: fixing, it was done for local testing purposes, reverting now to rancher

brandond (approved):

briandowns (approved):

PR #110, opened 2024-09-04
Changing the cluster to be namespaced

MbolotSuse:
Changes the cluster type to be namespaced (and changes the various controllers to work with this new feature). Also adds crd generation and docs to the core cluster type.

Note that this is technically a breaking change - existing clusters won't work and will need to be re-provisioned if this change is merged.


galal-hussein (approved):
LGTM, one change is to remove the createNamespace function since its not used anymore @MbolotSuse

vitorsavian (approved):

dereknola (approved):
LGTM

Issue #114, opened 2024-09-30
Isolation: network policy automatic addition


galal-hussein:
Currently none of the workloads generated within a k3k cluster can reach other pods in a different k3k cluster which is the correct behavior, however the pods still can reach the host's cluster network, the controller should automatically add a layer of isolation on the k3k cluster pods, for example by setting a network policy on the cluster's namespace so that it can't reach any other pods on the host's clusters

PR #125, opened 2024-10-01
Adding Networkpolicy to ClusterSets

galal-hussein:
The PR:
- Adds clusterset type by @MbolotSuse 
- Adds a default networkpolicy for each clusterset which does the following:
  - Accepts all ingress traffic
  - allow all egress except for:
    - podCIDRS of the current nodes in the system with an exception for the pods in the same namespace of the clusterset and the coredns pod
- Adds a node controller to track new nodes and update the netpolicies with their pod CIDR.

Issue:
- https://github.com/rancher/k3k/issues/114
- https://github.com/rancher/k3k/issues/128


main.go
```diff
@@ -4,32 +4,45 @@ package main
...
 
-var Scheme = runtime.NewScheme()
+const (
+	clusterCIDRFlagName = "cluster-cidr"
+	clusterCIDREnvVar   = "CLUSTER_CIDR"
+	KubeconfigFlagName  = "kubeconfig"
+)
+
+var (
+	Scheme      = runtime.NewScheme()
```
briandowns, line 27: This is defined in main should probably shouldn't/can't be exported.

pkg/controller/clusterset/clusterset.go
```diff
@@ -0,0 +1,157 @@
...
+	var clusterSet v1alpha1.ClusterSet
+	if err := c.Client.Get(ctx, types.NamespacedName{Name: req.Name, Namespace: req.Namespace}, &clusterSet); err != nil {
+		return reconcile.Result{}, fmt.Errorf("unable to get the clusterset: %w", err)
+	}
+
+	klog.Infof("got a clusterset: %v", clusterSet)
+	// create network policy
+	var (
+		setNetworkPolicy *networkingv1.NetworkPolicy
+		err              error
```
briandowns, line 61: Predefining the `err` value should be avoided in nearly all cases.

pkg/controller/clusterset/clusterset.go
```diff
@@ -0,0 +1,157 @@
...
+						UID:        clusterSet.UID,
+						Name:       clusterSet.Name,
+						APIVersion: clusterSet.APIVersion,
+						Kind:       clusterSet.Kind,
+					},
+				},
+			},
+		}
+		quota.Spec.Hard = clusterSet.Spec.MaxLimits
+		err := c.Client.Create(ctx, &quota)
```
briandowns, line 86: Scope lock this `err` val by moving the if check to be inline.

pkg/controller/clusterset/node.go
```diff
@@ -0,0 +1,134 @@
...
+	nodeController = "k3k-node-controller"
+)
+
+type NodeReconciler struct {
+	Client      ctrlruntimeclient.Client
+	Scheme      *runtime.Scheme
+	ClusterCIDR string
+}
+
+// Add adds a new controller to the manager
```
briandowns, line 31:
```suggestion
// AddNodeController adds a new controller to the manager
```

pkg/controller/clusterset/node.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+		}
+
+		return reconcile.Result{}, nil
+	}
+
+	for _, cs := range clusterSetList.Items {
+		klog.Infof("Remove CIDR %v of node %s from clusterSet %s networkpolicy", node.Spec.PodCIDR, req.Name, cs.Name)
+		if err := n.removeCIDR(ctx, cs, node.Spec.PodCIDR); err != nil {
+			return reconcile.Result{}, util.LogAndReturnErr("failed to remove cidr from networkpolicy", err)
```
briandowns, line 82: We seem to be mixing error logging and error wrapped logging together here. Is that intended?

galal-hussein: @MbolotSuse correct me if I am wrong but thats because Reconcile() in controller-runtime doesn't really log the returned error, however it begin a new cycle with a backoff for the controller, thats why I have a util.LogAndReturnErr() function to be able to log the error before returning it, that said I am open to what the controller-runtime does as a best practice for logging

MbolotSuse: I think that controller runtime will log an error if reconcile fails: https://github.com/kubernetes-sigs/controller-runtime/blob/c50f2dfd0b618220108d320e3e690340a9246a75/pkg/internal/controller/controller.go#L329. For that reason I think we can just return the error here. 

That being said, the kubebuilder book contains a sample that logs and returns the error: https://book.kubebuilder.io/cronjob-tutorial/controller-implementation#2-list-all-active-jobs-and-update-the-status. Overall, I think that logging something specific with a returned error is fine. I think that @briandowns is right that doing both through a side function is odd, I'd prefer that we logged out and returned in two different statements, despite the fact that it produces more code.

galal-hussein: That makes sense, I think controller-runtime has a guide https://github.com/kubernetes-sigs/controller-runtime/blob/main/TMP-LOGGING.md, I will try to follow this guide and adjust the logging through the whole repo, but this might be useful in a separate PR

pkg/controller/clusterset/clusterset.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("unable to make a networkpolicy for cluster set: %w", err)
+	}
+	if err := c.Client.Create(ctx, setNetworkPolicy); err != nil {
+		return reconcile.Result{}, fmt.Errorf("unable to create networkpolicy for clusterset: %w", err)
+	}
+	if clusterSet.Spec.MaxLimits != nil {
+		quota := v1.ResourceQuota{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "clusterset-quota",
```
MbolotSuse, line 69: Right now there's no limit on the number of clustersets that user can make for a given namespace. However, each clusterset will try to make a clusterset with the same name, which could cause problems.

pkg/controller/clusterset/clusterset.go
```diff
@@ -0,0 +1,152 @@
...
+						UID:        clusterSet.UID,
+						Name:       clusterSet.Name,
+						APIVersion: clusterSet.APIVersion,
+						Kind:       clusterSet.Kind,
+					},
+				},
+			},
+		}
+		quota.Spec.Hard = clusterSet.Spec.MaxLimits
+		if err := c.Client.Create(ctx, &quota); err != nil {
```
MbolotSuse, line 82: I think that this needs to handle the case where the quota already exists but we wanted to update the limits (i.e. check for conflict, and on conflict update). Do see the above note though on the name - if we don't fix the name to be exclusive to one clusterset that could cause problems.

pkg/apis/k3k.io/v1alpha1/set_types.go
```diff
@@ -0,0 +1,50 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	metav1.TypeMeta   `json:",inline"`
+
+	// Spec is the spec of the ClusterSet
+	Spec ClusterSetSpec `json:"spec"`
+	// Status is the status of the ClusterSet
+	Status ClusterSetStatus `json:"status,omitempty"`
+}
+
+type ClusterSetSpec struct {
```
MbolotSuse, line 23: I think we need some way of controlling if the cluster set creates a network policy (especially since the current policy blocks all ingress traffic, essentially making any services inaccessible outside of the cluster). For now maybe we can do a simple enable/disable bool.

pkg/controller/clusterset/clusterset.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return &networkingv1.NetworkPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      networkPolicyName,
+			Namespace: clusterSet.Namespace,
+		},
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "NetworkPolicy",
+			APIVersion: "networking.k8s.io/v1",
+		},
```
MbolotSuse, line 110: Nit: Is this necessary?

pkg/controller/clusterset/clusterset.go
```diff
@@ -0,0 +1,152 @@
...
+		Spec: networkingv1.NetworkPolicySpec{
+			PolicyTypes: []networkingv1.PolicyType{
+				networkingv1.PolicyTypeIngress,
+				networkingv1.PolicyTypeEgress,
+			},
+			Ingress: []networkingv1.NetworkPolicyIngressRule{
+				{},
```
MbolotSuse, line 117: This may be my ignorance speaking, but doesn't this block all ingress traffic, effectively preventing people from contacting the API server and using the cluster?

pkg/controller/clusterset/clusterset.go
galal-hussein: I copied that convention from all wrangler controller code, not really sure if its different in controller-runtime client, but in wrangler they always write the GVK in the resources they create programmatically

pkg/controller/clusterset/clusterset.go
galal-hussein: the ingress is block all by default, if its a list tho with an empty item then it will allow all ingress or egress https://kubernetes.io/docs/concepts/services-networking/network-policies/#allow-all-ingress-traffic

pkg/controller/clusterset/node.go
galal-hussein: I have added this issue https://github.com/rancher/k3k/issues/127 to track the logging problem in a separate pr(s)

briandowns (approved):

MbolotSuse (approved):

Issue #122, opened 2024-10-01
Virtual Kubelet: Controller Integration


MbolotSuse:
Currently the Virtual Kubelet must be deployed on it's own after the K3k cluster has been deployed. This is a poor and complicated user experience. The K3k controller should deploy the virtual kubelet and ensure that it registers to form a working virtual cluster without intervention from the user. The following requirements should be met:
- The K3k kubelet should be deployed automatically when the node type is set to shared
- The shared node type should be the default
- When the node type is shared, the server should be deployed in agentless mode
- When the node type is shared, no agents should be deployed (or permitted to be deployed)
- The k3k kubelet should be configured with the minimum permissions necessary in the host cluster
- The k3k kubelet should automatically register with the host cluster

PR #130, opened 2024-10-15
virtual-kubelet controller integration

galal-hussein:
The PR does the following:
- Convert the Agent structure into an interface which has two different implementations:
  - **virtual mode:** which is the original k3k mode in which the agent will run in a pod
  - **shared mode:** where the controller will run a virtual kubelet implementation that allows running workloads on the host
- Restructure the virtual-kubelet implementation:
  - Added a cli implementation
  - Added a config file to be parsed on the start of the program
  - Added a kubelet package where the actual registration/start of the node happens
- Add the virtual-kubelet (k3k-kubelet) image into the release workflow
Issue:
- #122


k3k-kubelet/pkg/config/config.go
```diff
@@ -0,0 +1,72 @@
+package config
```
briandowns, line 1: Let's get rid of the pkg directory.

k3k-kubelet/pkg/config/config.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"errors"
+	"os"
+
+	"gopkg.in/yaml.v2"
+)
+
+// Config has all virtual-kubelet startup options
+type Type struct {
```
briandowns, line 11: Can this be renamed to `Config`?

k3k-kubelet/pkg/kubelet/kubelet.go
```diff
@@ -0,0 +1,292 @@
+package kubelet
```
briandowns, line 1: Looks like doesn't need to be in a package and should be part of `package main`.

k3k-kubelet/pkg/config/config.go
briandowns, line 1: This doesn't appear as though it needs it's own package.

pkg/controller/cluster/agent/shared.go
```diff
@@ -0,0 +1,222 @@
...
+				Spec: s.podSpec(virtualKubeletImage, name, &selector),
+			},
+		},
+	}
+}
+
+func (s *SharedAgent) podSpec(image, name string, affinitySelector *metav1.LabelSelector) v1.PodSpec {
+	args := []string{"--config", "/opt/rancher/k3k/config.yaml"}
+	var limit v1.ResourceList
+	podSpec := v1.PodSpec{
```
briandowns, line 96: Looks like you can return here.

pkg/controller/cluster/agent/shared.go
```diff
@@ -0,0 +1,222 @@
...
+					Labels: selector.MatchLabels,
+				},
+				Spec: s.podSpec(virtualKubeletImage, name, &selector),
+			},
+		},
+	}
+}
+
+func (s *SharedAgent) podSpec(image, name string, affinitySelector *metav1.LabelSelector) v1.PodSpec {
+	args := []string{"--config", "/opt/rancher/k3k/config.yaml"}
```
briandowns, line 94: Is this path guaranteed to exist?

pkg/controller/cluster/config/server.go
```diff
@@ -9,6 +9,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	VirtualKubeletMode = "shared"
```
briandowns, line 13: Are these used outside of the package?

k3k-kubelet/pkg/kubelet/kubelet.go
```diff
@@ -0,0 +1,292 @@
...
+	pool := x509.NewCertPool()
+	pool.AddCert(certs[0])
+
+	return &tls.Config{
+		RootCAs:      pool,
+		Certificates: []tls.Certificate{clientCert},
+	}, nil
+}
+
+type LogWrapper struct {
```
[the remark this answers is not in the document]
galal-hussein: yeah its being used in the thread of running kubelet:

https://github.com/rancher/k3k/pull/130/files#diff-e53078009a4f5b2cec981fb4d89b9895034c9e083e9342f44ae5ac8c6f755effR110-R114

pkg/controller/cluster/agent/shared.go
galal-hussein: Yes so the default path is /etc/rancher/k3k/config.yaml however later in the pod spec you can see that we volume mount the config at /opt/rancher/k3k/config.yaml, but I can make it a constant since its used in two places

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,280 @@
...
+	if err := k.node.WaitReady(context.Background(), time.Minute*1); err != nil {
+		fmt.Printf("node was not ready within timeout of 1 minute: %s \n", err.Error())
+		os.Exit(-1)
+	}
+	<-k.node.Done()
+	if err := k.node.Err(); err != nil {
+		fmt.Printf("node stopped with an error: %s \n", err.Error())
+		os.Exit(-1)
+	}
+	fmt.Printf("node exited without an error")
```
briandowns, line 120:
```suggestion
	fmt.Print("node exited without an error")
```

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,280 @@
...
+			os.Exit(-1)
+		}
+		wrapped := LogWrapper{
+			*logger.Sugar(),
+		}
+		ctx = log.WithLogger(ctx, &wrapped)
+		err = k.node.Run(ctx)
+		if err != nil {
+			fmt.Printf("node errored when running: %s \n", err.Error())
+			os.Exit(-1)
```
briandowns, line 108: Let's make this a nix standard and use a positive integer between 0-255.

k3k-kubelet/main.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+	"os"
+
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli"
+)
+
+var (
+	configFile string
+	c          config
```
briandowns, line 14: Let's call this config.

k3k-kubelet/main.go
```diff
@@ -0,0 +1,101 @@
...
+	app.Action = Run
+	if err := app.Run(os.Args); err != nil {
+		logrus.Fatal(err)
+	}
+}
+
+func Run(clx *cli.Context) {
+	if err := c.Parse(configFile); err != nil {
+		fmt.Printf("failed to parse config file %s: %v", configFile, err)
+		os.Exit(-1)
```
briandowns, line 82: These exist codes should be brought into alignment with *NIX standards.

main.go
```diff
@@ -16,6 +16,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
+
```
briandowns, line 19: Get rid of newline.

pkg/controller/cluster/agent/agent.go
```diff
@@ -2,260 +2,29 @@ package agent
...
-				ObjectMeta: metav1.ObjectMeta{
-					Labels: selector.MatchLabels,
-				},
-				Spec: a.podSpec(image, agentName, cluster.Spec.AgentArgs, true, &selector),
-			},
-		},
+func New(cluster *v1alpha1.Cluster, serviceIP, sharedAgentImage string) Agent {
+	if cluster.Spec.Mode == VirtualNodeMode {
+		return NewVirtualAgent(cluster, serviceIP)
+	} else {
```
briandowns, line 23: I don' think this "else" is necessary.

pkg/controller/cluster/agent/shared.go
```diff
@@ -0,0 +1,247 @@
...
+		},
+	}, nil
+}
+
+func sharedAgentData(cluster *v1alpha1.Cluster) string {
+	return fmt.Sprintf(`clusterName: %s
+clusterNamespace: %s
+nodeName: %s
+agentHostname: %s
+token: %s`, cluster.Name, cluster.Namespace, cluster.Name+"-"+"k3k-kubelet", cluster.Name+"-"+"k3k-kubelet", cluster.Spec.Token)
```
briandowns, line 58: This can be on 2 lines with the second line being split after the comma here.

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,280 @@
...
+		return utilProvider, &nodeProvider, nil
+	}
+}
+
+func (k *kubelet) nodeOpts(srvPort, namespace, name, hostname string) nodeutil.NodeOpt {
+	return func(c *nodeutil.NodeConfig) error {
+		c.HTTPListenAddr = fmt.Sprintf(":%s", srvPort)
+		// set up the routes
+		mux := http.NewServeMux()
+		err := nodeutil.AttachProviderRoutes(mux)(c)
```
briandowns, line 141: Scope lock err value.

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,280 @@
...
+	context := clientcmdapi.NewContext()
+	context.AuthInfo = "default"
+	context.Cluster = "default"
+
+	config.Clusters["default"] = cluster
+	config.AuthInfos["default"] = authInfo
+	config.Contexts["default"] = context
+	config.CurrentContext = "default"
+
+	kubeconfig, err := clientcmd.Write(*config)
```
briandowns, line 215:
```suggestion
	return clientcmd.Write(*config)
```

k3k-kubelet/config.go
```diff
@@ -0,0 +1,76 @@
+package main
+
+import (
+	"errors"
+	"os"
+
+	"gopkg.in/yaml.v2"
+)
+
+// Config has all virtual-kubelet startup options
```
briandowns, line 10:
```suggestion
// config has all virtual-kubelet startup options.
```

k3k-kubelet/config.go
```diff
@@ -0,0 +1,76 @@
...
+	ClusterNamespace  string `yaml:"clusterNamespace,omitempty"`
+	NodeName          string `yaml:"nodeName,omitempty"`
+	Token             string `yaml:"token,omitempty"`
+	AgentHostname     string `yaml:"agentHostname,omitempty"`
+	HostConfigPath    string `yaml:"hostConfigPath,omitempty"`
+	VirtualConfigPath string `yaml:"virtualConfigPath,omitempty"`
+	KubeletPort       string `yaml:"kubeletPort,omitempty"`
+}
+
+func (t *config) unmarshalYAML(data []byte) error {
```
briandowns, line 22:
```suggestion
func (c *config) unmarshalYAML(data []byte) error {
```

k3k-kubelet/config.go
```diff
@@ -0,0 +1,76 @@
...
+	var c config
```
briandowns, line 23:
```suggestion
	var conf config
```

k3k-kubelet/config.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	return nil
+}
+
+func (t *config) Parse(path string) error {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return nil
+	}
+
+	configFileBytes, err := os.ReadFile(path)
```
briandowns, line 71:
```suggestion
	b, err := os.ReadFile(path)
```

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+	"k8s.io/client-go/kubernetes"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/client-go/util/retry"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var Scheme = runtime.NewScheme()
```
briandowns, line 35:
```suggestion
var scheme = runtime.NewScheme()
```

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	hostClient, err := ctrlruntimeclient.New(hostConfig, ctrlruntimeclient.Options{
+		Scheme: Scheme,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	virtConfig, err := virtRestConfig(context.Background(), c.VirtualConfigPath, hostClient, c.ClusterName, c.ClusterNamespace)
```
briandowns, line 64: Somewhere in the main func it'd be best to create a root context and pass it to whatever function down the exec line needs it rather than creating a one-off here.

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	return &kubelet{
+		name:       c.NodeName,
+		hostConfig: hostConfig,
+		hostClient: hostClient,
+		virtClient: virtClient,
+	}, nil
+}
+
+func (k *kubelet) RegisterNode(srvPort, namespace, name, hostname string) error {
```
briandowns, line 81:
```suggestion
func (k *kubelet) registerNode(srvPort, namespace, name, hostname string) error {
```

I don't think any of these methods on this type need to be exported given we're in package main.

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+	providerFunc := k.newProviderFunc(namespace, name, hostname)
+	nodeOpts := k.nodeOpts(srvPort, namespace, name, hostname)
+
+	var err error
+	k.node, err = nodeutil.NewNode(k.name, providerFunc, nodeutil.WithClient(k.virtClient), nodeOpts)
```
briandowns, line 86:
```suggestion
	k.node, err := nodeutil.NewNode(k.name, providerFunc, nodeutil.WithClient(k.virtClient), nodeOpts)
```

Delete the line above.

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to start kubelet: %v", err)
+	}
+	return nil
+}
+
+func (k *kubelet) Start(ctx context.Context) {
+	go func() {
+		ctx := context.Background()
```
briandowns, line 95: This should receive the same context from func main. The sig on this method will need to be updated to take that ctx value.

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+		logger, err := zap.NewProduction()
+		if err != nil {
+			fmt.Printf("unable to create logger: %s", err.Error())
```
briandowns, line 98:
```suggestion
			fmt.Println("unable to create logger: ", err.Error())
```

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+			os.Exit(1)
+		}
+		wrapped := LogWrapper{
+			*logger.Sugar(),
+		}
+		ctx = log.WithLogger(ctx, &wrapped)
+		err = k.node.Run(ctx)
```
briandowns, line 105:
```suggestion
		if err := k.node.Run(ctx)
```

These err values should be scope locked.

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+		if err != nil {
+			fmt.Printf("node errored when running: %s \n", err.Error())
+			os.Exit(1)
+		}
+	}()
+	if err := k.node.WaitReady(context.Background(), time.Minute*1); err != nil {
+		fmt.Printf("node was not ready within timeout of 1 minute: %s \n", err.Error())
```
briandowns, line 112:
```suggestion
		fmt.Println("node was not ready within timeout of 1 minute:", err.Error())
```

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+	return func(c *nodeutil.NodeConfig) error {
+		c.HTTPListenAddr = fmt.Sprintf(":%s", srvPort)
+		// set up the routes
+		mux := http.NewServeMux()
+		if err := nodeutil.AttachProviderRoutes(mux)(c); err != nil {
+			return fmt.Errorf("unable to attach routes: %w", err)
+		}
+		c.Handler = mux
+
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
```
briandowns, line 146: This should receive the same context from func main passed down and through. Goroutine management can't be had if we don't which will result in leaks.

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+func virtRestConfig(ctx context.Context, VirtualConfigPath string, hostClient ctrlruntimeclient.Client, clusterName, clusterNamespace string) (*rest.Config, error) {
+	if VirtualConfigPath != "" {
+		return clientcmd.BuildConfigFromFlags("", VirtualConfigPath)
+	}
+	// virtual kubeconfig file is empty, trying to fetch the k3k cluster kubeconfig
+	var cluster v1alpha1.Cluster
+	if err := hostClient.Get(ctx, types.NamespacedName{Namespace: clusterNamespace, Name: clusterName}, &cluster); err != nil {
+		return nil, err
+	}
+	endpoint := fmt.Sprintf("%s.%s", server.ServiceName(cluster.Name), cluster.Namespace)
```
briandowns, line 166:
```suggestion
	endpoint := server.ServiceName(cluster.Name)+"."+cluster.Namespace
```
We don't need to pressure the runtime with unnecessary type determination and unnecessary memory allocations here.

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+			fmt.Printf("unable to create logger: %s", err.Error())
+			os.Exit(1)
+		}
+		wrapped := LogWrapper{
+			*logger.Sugar(),
+		}
+		ctx = log.WithLogger(ctx, &wrapped)
+		err = k.node.Run(ctx)
+		if err != nil {
+			fmt.Printf("node errored when running: %s \n", err.Error())
```
briandowns, line 107:
```suggestion
			fmt.Println("node errored when running:", err.Error())
```

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+		tlsConfig, err := loadTLSConfig(ctx, k.hostClient, name, namespace, k.name, hostname)
+		if err != nil {
+			return fmt.Errorf("unable to get tls config: %w", err)
+		}
+		c.TLSConfig = tlsConfig
+		return nil
+	}
+}
+
+func virtRestConfig(ctx context.Context, VirtualConfigPath string, hostClient ctrlruntimeclient.Client, clusterName, clusterNamespace string) (*rest.Config, error) {
```
briandowns, line 157:
```suggestion
func virtRestConfig(ctx context.Context, virtualConfigPath string, hostClient ctrlruntimeclient.Client, clusterName, clusterNamespace string) (*rest.Config, error) {
```

We never need to capitalize a variable name in an argument definition.

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+		return nil, fmt.Errorf("unable to get cert and key: %w", err)
+	}
+	clientCert, err := tls.X509KeyPair(cert, key)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get key pair: %w", err)
+	}
+	// create rootCA CertPool
+	certs, err := certutil.ParseCertsPEM([]byte(b.ServerCA.Content))
+	if err != nil {
+		return nil, fmt.Errorf("unable to create certs: %w", err)
```
briandowns, line 249:
```suggestion
		return nil, fmt.Errorf("unable to create ca certs: %w", err)
```

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	pool := x509.NewCertPool()
+	pool.AddCert(certs[0])
```
briandowns, line 252: Should we do a bounds check here to be safe?

k3k-kubelet/kubelet.go
```diff
@@ -0,0 +1,274 @@
...
+
+	return &tls.Config{
+		RootCAs:      pool,
+		Certificates: []tls.Certificate{clientCert},
+	}, nil
+}
+
+type LogWrapper struct {
```
briandowns, line 260: Is any of this used or useful?

k3k-kubelet/provider/provider.go
```diff
@@ -313,23 +311,5 @@ func (p *Provider) translateFrom(hostPod *corev1.Pod) *corev1.Pod {
 }
 
 func (p *Provider) hostName(virtualNamespace string, virtualName string) string {
```
briandowns, line 313:
```suggestion
func (p *Provider) hostName(virtualNamespace, virtualName string) string {
```

pkg/controller/cluster/agent/agent.go
```diff
@@ -2,260 +2,28 @@ package agent
...
-	"github.com/rancher/k3k/pkg/controller/util"
-	apps "k8s.io/api/apps/v1"
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/utils/ptr"
-)
-
-const agentName = "k3k-agent"
+	"github.com/rancher/k3k/pkg/controller"
```
briandowns, line 5: If the line under this import is empty, remove it.

pkg/controller/controller.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// return all the nodes external addresses, if not found then return internal addresses
+func Addresses(ctx context.Context, client ctrlruntimeclient.Client) ([]string, error) {
+	var nodeList v1.NodeList
+	if err := client.List(ctx, &nodeList); err != nil {
+		return nil, err
+	}
+
+	var addresses []string
```
briandowns, line 67:
```suggestion
        addresses := make([]string, len(nodeList.Items))
```

pkg/controller/controller.go
```diff
@@ -0,0 +1,96 @@
...
+	return internalIP
+}
+
+// return all the nodes external addresses, if not found then return internal addresses
+func Addresses(ctx context.Context, client ctrlruntimeclient.Client) ([]string, error) {
+	var nodeList v1.NodeList
+	if err := client.List(ctx, &nodeList); err != nil {
+		return nil, err
+	}
+
```
briandowns, line 66: Should we have a len check on the list to if it's > 1, perform the action below?

k3k-kubelet/kubelet.go
galal-hussein: yeah its being added to the context of the thread of running the node

k3k-kubelet/main.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	app.Action = Run
+	if err := app.Run(os.Args); err != nil {
+		logrus.Fatal(err)
+	}
+}
+
+func Run(clx *cli.Context) {
+	if err := cfg.Parse(configFile); err != nil {
+		fmt.Printf("failed to parse config file %s: %v", configFile, err)
```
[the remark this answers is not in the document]
galal-hussein: fmt.Printf doesnt support %w

pkg/controller/controller.go
galal-hussein: I dont think so, because we wont need the node addresses if the nodes arent registered yet

k3k-kubelet/main.go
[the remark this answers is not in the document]
briandowns: Would fmt.Errorf make more sense here then?

k3k-kubelet/kubelet.go
briandowns: What is the resolution for this?

k3k-kubelet/main.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func run(clx *cli.Context) {
+	if err := cfg.parse(configFile); err != nil {
+		fmt.Printf("failed to parse config file %s: %v", configFile, err)
+		os.Exit(1)
+	}
+
+	if err := cfg.validate(); err != nil {
+		fmt.Printf("failed to validate config: %v", err)
```
briandowns, line 86:
```suggestion
		fmt.Errorf("failed to validate config: %w", err)
```

pkg/controller/cluster/agent/virtual.go
```diff
@@ -0,0 +1,223 @@
...
+			Name:      configSecretName(v.cluster.Name),
+			Namespace: v.cluster.Namespace,
+		},
+		Data: map[string][]byte{
+			"config.yaml": []byte(config),
+		},
+	}, nil
+}
+
+func (v *VirtualAgent) Resources() ([]ctrlruntimeclient.Object, error) {
```
briandowns, line 50: We can probably update this, the interface, and the other implementations to not return an error unless the other implementations do.

pkg/controller/cluster/agent/virtual.go
```diff
@@ -0,0 +1,223 @@
...
+	var objs []ctrlruntimeclient.Object
+	objs = append(objs, v.deployment())
+	return objs, nil
```
briandowns, line 53:
```suggestion
	return []ctrlruntimeclient.Object{v.deployment()}, nil
```

pkg/controller/cluster/agent/shared.go
```diff
@@ -0,0 +1,249 @@
...
+nodeName: %s
+agentHostname: %s
+token: %s`,
+		cluster.Name, cluster.Namespace, nodeName, nodeName, cluster.Spec.Token)
+}
+
+func (s *SharedAgent) Resources() ([]ctrlruntimeclient.Object, error) {
+	var objs []ctrlruntimeclient.Object
+	objs = append(objs, s.serviceAccount(), s.role(), s.roleBinding(), s.service(), s.deployment())
+	return objs, nil
```
briandowns, line 66:
```suggestion
	return []ctrlruntimeclient.Object{s.serviceAccount(), s.role(), s.roleBinding(), s.service(), s.deployment()}, nil
```

Not sure if this will work but makes sense and will remove a line from the function.

briandowns (approved):

k3k-kubelet/kubelet.go
galal-hussein: sorry missed this one, adding a check now

galal-hussein: added a check for the length of the certs after this line

MbolotSuse (approved):
No major concerns, thanks for all the work on this PR.

k3k-kubelet/config.go
```diff
@@ -0,0 +1,76 @@
...
+		c.KubeletPort = conf.KubeletPort
+	}
+	if c.AgentHostname == "" {
+		c.AgentHostname = conf.AgentHostname
+	}
+	if c.NodeName == "" {
+		c.NodeName = conf.NodeName
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
galal-hussein: the problem is that we want the user to be able to mix and match for example, the user can set the --agent-hostname as a flag and at the same time add `nodeName` in the config file, if we just returned the unmarshalled config it will ignore all the flags sat on the command line

Issue #127, opened 2024-10-03
Fix logging throughout the repo


galal-hussein:
Currently there are multiple controllers in k3k regarding clsutersets, ha controller, network policy node controller and clsuter controller, logging need to be more structured and organized through all controllers.

PR #133, opened 2024-10-21
Add structured logging via zap

galal-hussein:
The PR adds the following:
- Removes different log messages
- Removes the wrong wrapper function
- Adds structured logging (zap) with a wrapper to be used in `virtual-kubelet` as well
- Configure controller-runtime with the reconciler logger
- Configure the logger for each controller to have a different fields according to the controller logic

Issue:
- https://github.com/rancher/k3k/issues/127

### DIFF
--- a/k3k-kubelet/kubelet.go
+++ b/k3k-kubelet/kubelet.go
@@ -109,7 +109,7 @@ func (k *kubelet) start(ctx context.Context) {
 	if err := k.node.Err(); err != nil {
 		k.logger.Fatalw("node stopped with an error", zap.Error(err))
 	}
-	k.logger.Info("node exited without an error")
+	k.logger.Info("node exited successfully")
 }
 
 func (k *kubelet) newProviderFunc(namespace, name, hostname string) nodeutil.NewProviderFunc {
--- a/k3k-kubelet/kubelet.go
+++ b/k3k-kubelet/kubelet.go
@@ -6,7 +6,6 @@ import (
 	"crypto/x509"
 	"fmt"
 	"net/http"
-	"os"
 	"time"
 
 	certutil "github.com/rancher/dynamiclistener/cert"
@@ -16,6 +15,7 @@ import (
 	"github.com/rancher/k3k/pkg/controller/cluster/server"
 	"github.com/rancher/k3k/pkg/controller/cluster/server/bootstrap"
 	"github.com/rancher/k3k/pkg/controller/kubeconfig"
+	k3klog "github.com/rancher/k3k/pkg/log"
 	"github.com/virtual-kubelet/virtual-kubelet/log"
 	"github.com/virtual-kubelet/virtual-kubelet/node"
 	"github.com/virtual-kubelet/virtual-kubelet/node/nodeutil"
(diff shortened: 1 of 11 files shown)


pkg/log/zap.go
```diff
@@ -0,0 +1,51 @@
...
+
+type Logger struct {
+	*zap.SugaredLogger
+}
+
+func New(debug bool) *Logger {
+	return &Logger{newZappLogger(debug).Sugar()}
+}
+
+func (k *Logger) WithError(err error) log.Logger {
```
briandowns, line 20:
```suggestion
func (l *Logger) WithError(err error) log.Logger {
```

pkg/controller/clusterset/node.go
```diff
@@ -43,33 +46,36 @@ func AddNodeController(ctx context.Context, mgr manager.Manager) error {
...
-	if err := n.ensureNetworkPolicies(ctx, clusterSetList); err != nil {
+	if err := n.ensureNetworkPolicies(ctx, clusterSetList, log); err != nil {
 		return reconcile.Result{}, err
 	}
 
 	return reconcile.Result{}, nil
 }
 
-func (n *NodeReconciler) ensureNetworkPolicies(ctx context.Context, clusterSetList v1alpha1.ClusterSetList) error {
+func (n *NodeReconciler) ensureNetworkPolicies(ctx context.Context, clusterSetList v1alpha1.ClusterSetList, log *zap.SugaredLogger) error {
```
briandowns, line 66: Does it need to be passed in as an arg since it's on the struct?

pkg/log/zap.go
briandowns, line 20: How would this be used?

k3k-kubelet/kubelet.go
```diff
@@ -92,30 +97,19 @@ func (k *kubelet) registerNode(ctx context.Context, srvPort, namespace, name, ho
...
+		k.logger.Fatalw("node was not ready within timeout of 1 minute", zap.Error(err))
 	}
 	<-k.node.Done()
 	if err := k.node.Err(); err != nil {
-		fmt.Println("node stopped with an error:", err.Error())
-		os.Exit(1)
+		k.logger.Fatalw("node stopped with an error", zap.Error(err))
 	}
-	fmt.Println("node exited without an error")
+	k.logger.Info("node exited without an error")
```
briandowns, line 112:
```suggestion
	k.logger.Info("node exited successfully")
```

main.go
```diff
@@ -63,9 +74,12 @@ func main() {
...
 	app.Action = run
 	app.Version = version + " (" + gitCommit + ")"
-
+	app.Before = func(clx *cli.Context) error {
+		logger = log.New(debug)
+		return nil
+	}
 	if err := app.Run(os.Args); err != nil {
-		klog.Fatal(err)
+		logger.Fatalw("Failed to run k3k controller", zap.Error(err))
```
briandowns, line 82:
```suggestion
		logger.Fatalw("failed to run k3k controller", zap.Error(err))
```
Lower casing here since it would conform to the rest of the messages.

pkg/controller/clusterset/node.go
galal-hussein: short answer is yes, the log is indeed in the structure however, for each reconcile loop the log object is initiated with a .With fields to describe the object name in that reconcile loop and this one gets passed around through the functions, its more lines of code but its the convention with logging in controller-runtime apparently

pkg/log/zap.go
galal-hussein: This is used in the virtual kubelet thread, https://github.com/rancher/k3k/pull/133/files#diff-e53078009a4f5b2cec981fb4d89b9895034c9e083e9342f44ae5ac8c6f755effR100, the virtual kubelet logger needs this interface to be implemented

briandowns (approved):

Issue #115, opened 2024-10-01
Virtual Kubelet: Pod Volumes + Mounts


MbolotSuse:
Currently, the Virtual Kubelet removes any volumes/volume mounts from a created pods configuration before syncing the pod to the host cluster. However, this is not acceptable outside of the POC - users should be able to run pods which use volumes.

Instead, for volumes that are derived from API resources (such as configmaps, secrets, service account tokens, API Server CA, projected volumes, pvcs) these resources should be synced from the virtual cluster to the host cluster. The version of these resources in the host cluster should have a name unique to the virtual cluster (so that the configmap "example" in namespace "default" can exist in two different virtual clusters in the same tenant). If users update the source resource, the new version should be automatically synchronized to the host cluster. If these resources are removed from the virtual cluster, they should be removed from the virtual cluster.

For volumes that are derived from resources outside of the cluster (e.x. nfs), these resources should be left as-is (allowing users to use properly named values from the host cluster).

PR #137, opened 2024-10-28
Adding basic volume syncing

MbolotSuse:
Adds syncing for basic volume types (secret/configmap/projected secret and configmap). Also changes the virtual kubelet to use a cache from controller-runtime rather than a client for some operations.

Related to #115.

### DIFF
--- a/k3k-kubelet/controller/configmap.go
+++ b/k3k-kubelet/controller/configmap.go
@@ -1,0 +1,160 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type ConfigMapSyncer struct {
+	sync.RWMutex
+	// VirtualClient is the client for the virtual cluster
+	VirtualClient client.Client
+	// CoreClient is the client for the host cluster
+	HostClient client.Client
+	// TranslateFunc is the function that translates a given resource from it's virtual representation to the host
+	// representation
+	TranslateFunc func(*corev1.ConfigMap) (*corev1.ConfigMap, error)
+	// objs are the objects that the syncer should watch/syncronize. Should only be manipulated
+	// through add/remove
+	objs sets.Set[types.NamespacedName]
+}
+
+// Reconcile implements reconcile.Reconciler and synchronizes the objects in objs to the host cluster
+func (s *ConfigMapSyncer) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	if !s.isWatching(req.NamespacedName) {
+		// return immediately without re-enqueueing. We aren't watching this resource
+		return reconcile.Result{}, nil
+	}
+	var virtual corev1.ConfigMap
+	err := s.VirtualClient.Get(ctx, req.NamespacedName, &virtual)
+	if err != nil {
+		return reconcile.Result{
+			Requeue: true,
+		}, fmt.Errorf("unable to get configmap %s/%s from virtual cluster: %w", req.Namespace, req.Name, err)
+	}
+	translated, err := s.TranslateFunc(&virtual)
+	if err != nil {
+		return reconcile.Result{
+			Requeue: true,
+		}, fmt.Errorf("unable to translate configmap %s/%s from virtual cluster: %w", req.Namespace, req.Name, err)
+	}
+	translatedKey := types.NamespacedName{
+		Namespace: translated.Namespace,
+		Name:      translated.Name,
+	}
+	var host corev1.ConfigMap
+	err = s.HostClient.Get(ctx, translatedKey, &host)
+	if apierrors.IsNotFound(err) {
+		err = s.HostClient.Create(ctx, translated)
+		// for simplicity's sake, we don't check for conflict errors. The existing object will get
+		// picked up on in the next re-enqueue
+		return reconcile.Result{
+				Requeue: true,
+			}, fmt.Errorf("unable to create host configmap %s/%s for virtual configmap %s/%s: %w",
+				translated.Namespace, translated.Name, req.Namespace, req.Name, err)
+	}
+	// we are going to use the host in order to avoid conflicts on update
+	host.Data = translated.Data
+	if host.Labels == nil {
+		host.Labels = map[string]string{}
+	}
+	// we don't want to override labels made on the host cluster by other applications
+	// but we do need to make sure the labels that the kubelet uses to track host cluster values
+	// are being tracked appropriately
+	for key, value := range translated.Labels {
+		host.Labels[key] = value
+	}
+	err = s.HostClient.Update(ctx, &host)
+	if err != nil {
+		return reconcile.Result{
+				Requeue: true,
+			}, fmt.Errorf("unable to update host configmap %s/%s for virtual configmap %s/%s: %w",
+				translated.Namespace, translated.Name, req.Namespace, req.Name, err)
+
+	}
+	return reconcile.Result{}, nil
+}
+
+// isWatching is a utility method to determine if a key is in objs without the caller needing
+// to handle mutex lock/unlock.
+func (s *ConfigMapSyncer) isWatching(key types.NamespacedName) bool {
+	s.RLock()
+	defer s.RUnlock()
+	return s.objs.Has(key)
+}
+
+// Add adds a given resource to the list of resources that will be synced. Safe to call multiple times for the
+// same resource.
+func (s *ConfigMapSyncer) AddResource(ctx context.Context, namespace string, name string) error {
+	objKey := types.NamespacedName{
+		Namespace: namespace,
+		Name:      name,
+	}
+	// if we already sync this object, no need to writelock/add it
+	if s.isWatching(objKey) {
+		return nil
+	}
+	// lock in write mode since we are now adding the key
+	s.Lock()
+	if s.objs == nil {
+		s.objs = sets.Set[types.NamespacedName]{}
+	}
+	s.objs = s.objs.Insert(objKey)
+	s.Unlock()
+	_, err := s.Reconcile(ctx, reconcile.Request{
+		NamespacedName: objKey,
+	})
+	if err != nil {
+		return fmt.Errorf("unable to reconcile new object %s/%s: %w", objKey.Namespace, objKey.Name, err)
+	}
+	return nil
+}
+
+// Remove removes a given resource from the list of resources that will be synced. Safe to call for an already
+// removed resource.
+func (s *ConfigMapSyncer) RemoveResource(ctx context.Context, namespace string, name string) error {
+	objKey := types.NamespacedName{
+		Namespace: namespace,
+		Name:      name,
+	}
+	// if we don't sync this object, no need to writelock/add it
+	if !s.isWatching(objKey) {
+		return nil
+	}
+	// lock in write mode since we are now adding the key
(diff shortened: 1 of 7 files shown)


k3k-kubelet/kubelet.go
```diff
@@ -71,15 +78,51 @@ func newKubelet(ctx context.Context, c *config, logger *k3klog.Logger) (*kubelet
...
 
 	virtClient, err := kubernetes.NewForConfig(virtConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	hostMgr, err := ctrl.NewManager(hostConfig, manager.Options{
+		Scheme: baseScheme,
+		Metrics: ctrlserver.Options{
+			BindAddress: ":8083",
```
galal-hussein, line 88: do we need the metrics here?

k3k-kubelet/kubelet.go
```diff
@@ -71,15 +78,51 @@ func newKubelet(ctx context.Context, c *config, logger *k3klog.Logger) (*kubelet
...
+	if err != nil {
+		fmt.Printf("unable to create controller-runtime mgr for host cluster: %s", err.Error())
+		os.Exit(-1)
+	}
+
+	virtualScheme := runtime.NewScheme()
+	// virtual client will only use core types (for now), no need to add anything other than the basics
+	err = clientgoscheme.AddToScheme(virtualScheme)
+	if err != nil {
+		fmt.Printf("unable to add client go types to virtual cluster scheme: %s", err.Error())
```
galal-hussein, line 105:
```suggestion
		return fmt.Erorrf("unable to add client go types to virtual cluster scheme: %s", err.Error()")
```
You can just return the error here and the main will exit no need to call os.Exit(-1)

k3k-kubelet/kubelet.go
```diff
@@ -96,12 +139,33 @@ func (k *kubelet) registerNode(ctx context.Context, srvPort, namespace, name, ho
 }
 
 func (k *kubelet) start(ctx context.Context) {
+	// any one of the following 3 tasks (host manager, virtual manager, node) crashing will stop the
+	// program, and all 3 of them block on start, so we start them here in go-routines
+	go func() {
+		err := k.hostMgr.Start(context.Background())
+		if err != nil {
+			fmt.Printf("host manager stopped: %s \n", err.Error())
```
galal-hussein, line 147:
```suggestion
			k.logger.Fatalw("host manager stopped", zap.Error(err))
```

k3k-kubelet/kubelet.go
```diff
@@ -96,12 +139,33 @@ func (k *kubelet) registerNode(ctx context.Context, srvPort, namespace, name, ho
...
+			os.Exit(-1)
+		}
+	}()
+
+	go func() {
+		err := k.virtualMgr.Start(context.Background())
+		if err != nil {
+			fmt.Printf("host manager stopped: %s \n", err.Error())
```
galal-hussein, line 155:
```suggestion
		k.logger.Fatalw("host manager stopped", zap.Error(err))
```

k3k-kubelet/provider/provider.go
```diff
@@ -193,75 +221,223 @@ func (p *Provider) PortForward(ctx context.Context, namespace, pod string, port
...
-	return err
+	tPod.UID = ""
+	tPod.ResourceVersion = ""
+
+	// the node was scheduled on the virtual kubelet, but leaving it this way will make it pending indefinitely
+	tPod.Spec.NodeName = ""
+
+	// volumes will often refer to resources in the virtual cluster, but instead need to refer to the sync'd
+	// host cluster version
+	err := p.transformVolumes(ctx, pod.Namespace, tPod.Spec.Volumes)
```
galal-hussein, line 236: nit: scope locking the err value

k3k-kubelet/provider/provider.go
```diff
@@ -193,75 +221,223 @@ func (p *Provider) PortForward(ctx context.Context, namespace, pod string, port
...
+	if err != nil {
+		return fmt.Errorf("unable to sync volumes for pod %s/%s: %w", pod.Namespace, pod.Name, err)
+	}
+	p.logger.Infof("Creating pod %s/%s for pod %s/%s", tPod.Namespace, tPod.Name, pod.Namespace, pod.Name)
```
galal-hussein, line 240: just to be consistent with the rest of the code I think we can just use 

```suggestion
	p.logger.Infow("Creating pod", "Translated Namespace", tPod.Namespace, "Translated Name", tPod.Name, "Namespace", pod.Namespace, "Name", pod.Name)
```

k3k-kubelet/provider/provider.go
```diff
@@ -193,75 +221,223 @@ func (p *Provider) PortForward(ctx context.Context, namespace, pod string, port
...
+}
+
+// transformVolumes changes the volumes to the representation in the host cluster. Will return an error
+// if one/more volumes couldn't be transformed
+func (p *Provider) transformVolumes(ctx context.Context, podNamespace string, volumes []corev1.Volume) error {
+	for _, volume := range volumes {
+		// note: this needs to handle downward api volumes as well, but more thought is needed on how to do that
+		if volume.ConfigMap != nil {
+			if err := p.syncConfigmap(ctx, podNamespace, volume.ConfigMap.Name); err != nil {
+				return fmt.Errorf("unable to sync configmap volume %s: %w", volume.Name, err)
```
galal-hussein, line 251: Not really sure what is the correct way here but since we are using zap all around maybe we can use zap.Fields instead of fmt.Errorf, better actually if the logger can be passed to the transformVolumes function and add the fields there so that it can actually appear what kind of resources are we working on, as an example:

https://github.com/rancher/k3k/blob/main/pkg/controller/cluster/pod.go#L65-L66

k3k-kubelet/provider/provider.go
```diff
@@ -193,75 +221,223 @@ func (p *Provider) PortForward(ctx context.Context, namespace, pod string, port
...
 func (p *Provider) DeletePod(ctx context.Context, pod *corev1.Pod) error {
-	err := p.CoreClient.Pods(p.ClusterNamespace).Delete(ctx, p.translateTo(pod).Name, metav1.DeleteOptions{})
+	p.logger.Infof("Got request to delete pod %s", pod.Name)
+	hostName := p.Translater.TranslateName(pod.Namespace, pod.Name)
+	err := p.CoreClient.Pods(p.ClusterNamespace).Delete(ctx, hostName, metav1.DeleteOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to delete pod %s/%s: %w", pod.Namespace, pod.Name, err)
+	}
+	err = p.pruneUnusedVolumes(ctx, pod)
+	if err != nil {
```
galal-hussein, line 347: Nit: Err scope lock

k3k-kubelet/provider/provider.go
```diff
@@ -193,75 +221,223 @@ func (p *Provider) PortForward(ctx context.Context, namespace, pod string, port
...
+	for _, configMapName := range pruneConfigMap.UnsortedList() {
+		var configMap corev1.ConfigMap
+		err := p.VirtualClient.Get(ctx, types.NamespacedName{
+			Name:      configMapName,
+			Namespace: pod.Namespace,
+		}, &configMap)
+		if err != nil {
+			return fmt.Errorf("unable to get configMap %s/%s for pod volume: %w", pod.Namespace, configMapName, err)
+		}
+		err = p.Handler.RemoveResource(ctx, &configMap)
```
galal-hussein, line 402: since this is not really a controller the k8s api can return an error if the resource is already locked by any other operation and virtual kubelet may create an error then, I think we can either add a retry-backoff mechanism or use a thread to delete the pod?

k3k-kubelet/translate/host.go
```diff
@@ -0,0 +1,105 @@
...
+	"fmt"
+
+	"github.com/rancher/k3k/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// ClusterNameLabel is the key for the label that contains the name of the virtual cluster
+	// this resource was made in
+	ClusterNameLabel = "virt.cattle.io/clusterName"
```
galal-hussein, line 14: are we using `virt.cattle.io` all around the cluster? since we already using k3k.io to the finalizers we add

k3k-kubelet/controller/configmap.go
```diff
@@ -0,0 +1,160 @@
...
+	// VirtualClient is the client for the virtual cluster
+	VirtualClient client.Client
+	// CoreClient is the client for the host cluster
+	HostClient client.Client
+	// TranslateFunc is the function that translates a given resource from it's virtual representation to the host
+	// representation
+	TranslateFunc func(*corev1.ConfigMap) (*corev1.ConfigMap, error)
+	// objs are the objects that the syncer should watch/syncronize. Should only be manipulated
+	// through add/remove
+	objs sets.Set[types.NamespacedName]
```
galal-hussein, line 27: I think we can add the logger to the controller here as well

k3k-kubelet/kubelet.go
MbolotSuse: I figure that it can't hurt. I can disable it if you think that's necessary, but I think that it makes sense to leave that on unless we have a reason.

k3k-kubelet/kubelet.go
MbolotSuse: My bad, I think that I wrote this back when this was the main method. I'll fix this and the other similar occurrences that I see.

galal-hussein: No worries, Thank you!

k3k-kubelet/kubelet.go
galal-hussein: makes sense

k3k-kubelet/provider/provider.go
MbolotSuse: I don't have a problem with things near log statement using zap, but I think that errors in the program shouldn't be coupled tightly to the logger in case we want to change it.

k3k-kubelet/translate/host.go
MbolotSuse: I'm fine with changing the label, I can use "k3k.io/" as the prefix if you would prefer.

k3k-kubelet/provider/provider.go
briandowns: Retry would be cleaner and likely more inline with k8s code expectations.

k3k-kubelet/provider/provider.go
```diff
@@ -270,46 +446,44 @@ func (p *Provider) GetPodStatus(ctx context.Context, namespace, name string) (*c
...
-	delete(virtualPod.Annotations, podNamespaceAnnotation)
-	return virtualPod
-}
-
-func (p *Provider) hostName(virtualNamespace, virtualName string) string {
-	return controller.SafeConcatName(p.ClusterName, p.ClusterNamespace, virtualNamespace, virtualName)
+// getSecretsAndConfigmaps retrieves a list of all secrets/configmaps that are in use by a given pod. Useful
+// for removing/seeing which virtual cluster resources need to be in the host cluster.
+func getSecretsAndConfigmaps(pod *corev1.Pod) ([]string, []string) {
+	secrets := []string{}
```
briandowns, line 471: var secrets []string for here and below. This is the standard the repo is putting forth.

briandowns (changes requested):

k3k-kubelet/provider/provider.go
MbolotSuse: I'm ok with adding that elsewhere, but I don't think we can add that here. The provider doesn't "know" how to create/update/or delete resources (other than pods) in the host cluster - the controllers and the handlers are what know how to do that.

k3k-kubelet/provider/provider.go
MbolotSuse: Like I said above, I can definitely change this when logging, but when returning an error internally we should avoid reliance on unnecessary third party libraries. I can make a structured log statement here in addition to the error return if you would prefer.

galal-hussein (approved):
LGTM, thanks for the PR!

briandowns (approved):

PR #139, opened 2024-11-06
Serviceaccount token synchronization

galal-hussein:
- https://github.com/rancher/k3k/issues/115

### DIFF
--- a/k3k-kubelet/main.go
+++ b/k3k-kubelet/main.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"os"
 
+	"github.com/go-logr/zapr"
 	"github.com/rancher/k3k/pkg/log"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 	"go.uber.org/zap"
+	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 var (
@@ -66,6 +68,12 @@ func main() {
 			EnvVar:      "AGENT_HOSTNAME",
 		},
 		cli.StringFlag{
+			Name:        "agent-ip",
+			Usage:       "Agent IP used for registring the virtual kubelet to the cluster",
+			Destination: &cfg.AgentIP,
+			EnvVar:      "AGENT_IP",
+		},
+		cli.StringFlag{
 			Name:        "config",
 			Usage:       "Path to k3k-kubelet config file",
 			Destination: &configFile,
@@ -81,6 +89,7 @@ func main() {
 	}
 	app.Before = func(clx *cli.Context) error {
 		logger = log.New(debug)
+		ctrlruntimelog.SetLogger(zapr.NewLogger(logger.Desugar().WithOptions(zap.AddCallerSkip(1))))
 		return nil
 	}
 	app.Action = run
@@ -103,7 +112,7 @@ func run(clx *cli.Context) {
 		logger.Fatalw("failed to create new virtual kubelet instance", zap.Error(err))
 	}
 
-	if err := k.registerNode(ctx, cfg.KubeletPort, cfg.ClusterNamespace, cfg.ClusterName, cfg.AgentHostname); err != nil {
+	if err := k.registerNode(ctx, cfg.AgentIP, cfg.KubeletPort, cfg.ClusterNamespace, cfg.ClusterName, cfg.AgentHostname); err != nil {
 		logger.Fatalw("failed to register new node", zap.Error(err))
 	}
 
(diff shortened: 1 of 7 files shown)


k3k-kubelet/provider/provider.go
```diff
@@ -236,28 +243,44 @@ func (p *Provider) CreatePod(ctx context.Context, pod *corev1.Pod) error {
 // if one/more volumes couldn't be transformed
 func (p *Provider) transformVolumes(ctx context.Context, podNamespace string, volumes []corev1.Volume) error {
 	for _, volume := range volumes {
```
briandowns, line 245: This should be moved down closer to where it's used. It's unnecessary to have on the stack if the conditional block is t entered into.

k3k-kubelet/provider/provider.go
```diff
@@ -402,7 +432,7 @@ func (p *Provider) pruneUnusedVolumes(ctx context.Context, pod *corev1.Pod) erro
 // concurrently outside of the calling goroutine. Therefore it is recommended
 // to return a version after DeepCopy.
 func (p *Provider) GetPod(ctx context.Context, namespace, name string) (*corev1.Pod, error) {
-	p.logger.Errorf("got a request for get pod %s, %s", namespace, name)
```
briandowns, line 405: Smart change.

k3k-kubelet/provider/token.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	k3kcontroller "github.com/rancher/k3k/pkg/controller"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
briandowns, line 12: Remove new line.

briandowns (changes requested):

k3k-kubelet/main.go
```diff
@@ -65,6 +67,12 @@ func main() {
 			Destination: &cfg.AgentHostname,
 			EnvVar:      "AGENT_HOSTNAME",
 		},
+		cli.StringFlag{
+			Name:        "agent-ip",
+			Usage:       "Agent IP used for registring the virtual kubelet to the cluster",
```
enrichman, line 72:
```suggestion
			Usage:       "Agent IP used for registering the virtual kubelet to the cluster",
```

briandowns (approved):

k3k-kubelet/provider/provider.go
```diff
@@ -249,28 +255,44 @@ func (p *Provider) CreatePod(ctx context.Context, pod *corev1.Pod) error {
...
+			if err := p.syncConfigmap(ctx, podNamespace, volume.ConfigMap.Name, optional); err != nil {
 				return fmt.Errorf("unable to sync configmap volume %s: %w", volume.Name, err)
 			}
 			volume.ConfigMap.Name = p.Translater.TranslateName(podNamespace, volume.ConfigMap.Name)
 		} else if volume.Secret != nil {
-			if err := p.syncSecret(ctx, podNamespace, volume.Secret.SecretName); err != nil {
+			if volume.Secret.Optional != nil {
+				optional = *volume.Secret.Optional
+			}
+			if err := p.syncSecret(ctx, podNamespace, volume.Secret.SecretName, optional); err != nil {
```
enrichman, line 275: Imho the code will look more readable changing the structure to have less branches, and passing the struct to the `syncXXX` funcs. Something like:

```go
if volume.ConfigMap != nil {
	if err := p.syncConfigmap(ctx, podNamespace, volume.ConfigMap; err != nil {
		return fmt.Errorf("unable to sync configmap volume %s: %w", volume.Name, err)
	}
	volume.ConfigMap.Name = p.Translater.TranslateName(podNamespace, volume.ConfigMap.Name)
	continue
}

if volume.Secret != nil {
	// same
}
```

What do you think?

galal-hussein: that makes sense, fixing and testing now

galal-hussein: hmmm, I think why Michael didnt want to do that in the first place, simply because there are 2 types of volumes , there is simple configmap and projection configmap, which are two different types, thats why he only sent the names and not the whole object, that said Michael wanted to make the syncer more generic anyway, so I think we can merge for now and for future iteration we can conver the whole syncer to a more generic functions to accept any type of object

enrichman: Yes, I've just realized that. :/

Kinda annoying but I agree, thanks anyway! :+1:

enrichman (approved):

Issue #121, opened 2024-10-01
[EPIC] ClusterSet Integration


MbolotSuse:
A review should be done to make sure that the virtual kubelet implementation integrates well with the ClusterSet. Specifically, the following features should be integrated/implemented/tested:
- The cluster set should be able to provide a network policy preventing clusters from accessing a cluster in another cluster set
- The cluster set should be able to specify a PSS that applies to every pod in every virtual cluster in the set
- The cluster set should be able to specify a priority class that applies to every pod in every virtual cluster in the set
- The cluster set's resource limits and defaults should apply to all workloads in every virtual cluster in the set
- Only workloads which are currently running should apply to the limits set in the cluster set
- The cluster set should specify which node type (shared vs virtual) is allowed

enrichman:
Not yet, some of the bullet points still needs to be implemented:

- [x] The cluster set should be able to provide a network policy preventing clusters from accessing a cluster in another cluster set
- [x] The cluster set should be able to specify a PSS that applies to every pod in every virtual cluster in the set
- [ ] The cluster set should be able to specify a priority class that applies to every pod in every virtual cluster in the set
- [ ] The cluster set's resource limits and defaults should apply to all workloads in every virtual cluster in the set
- [ ] Only workloads which are currently running should apply to the limits set in the cluster set
- [ ] The cluster set should specify which node type (shared vs virtual) is allowed

ShylajaDevadiga:
Closing as all the sub-issues mentioned above are validated.

PR #144, opened 2024-11-11
Added `allowedNodeTypes` to ClusterSet, and fixed NetworkPolicy reconciliation

enrichman:
Ref:
- #121

This PR adds the `allowedNodeTypes` to the ClusterSet CRD, as per RFC. It uses a enum to enforce the value to be `shared` or `virtual`, also using kubebuilder annotations for the validation, and default value:

```go
// +kubebuilder:default={shared}
// +kubebuilder:validation:XValidation:message="mode is immutable",rule="self == oldSelf"
// +kubebuilder:validation:MinItems=1
//
// AllowedNodeTypes are the allowed cluster provisioning modes. Defaults to [shared].
AllowedNodeTypes []ClusterMode `json:"allowedNodeTypes,omitempty"`
```

```go
// +kubebuilder:validation:Enum=shared;virtual
// +kubebuilder:default="shared"
//
// ClusterMode is the possible provisioning mode of a Cluster.
type ClusterMode string

const (
	SharedClusterMode  = ClusterMode("shared")
	VirtualClusterMode = ClusterMode("virtual")
)
```

Updated the Cluster spec, added some tests, and an example.

This PR also adds a couple of fixes to the associated NetworkPolicy. It fixes the reconciliation loop, deleting the NetPol if the `disableNetworkPolicy` is changed, and it adds the ClusterSet owner reference, to allow the garbage collection when the ClusterSet is deleted.

### DIFF
--- a/pkg/apis/k3k.io/v1alpha1/types.go
+++ b/pkg/apis/k3k.io/v1alpha1/types.go
@@ -88,8 +88,8 @@ type ClusterSpec struct {
 type ClusterMode string
 
 const (
-	Shared  = ClusterMode("shared")  // e.g., 12e6
-	Virtual = ClusterMode("virtual") // e.g., 12Mi (12 * 2^20)
+	Shared  = ClusterMode("shared")
+	Virtual = ClusterMode("virtual")
 )
 
 type ClusterLimit struct {
--- a/pkg/apis/k3k.io/v1alpha1/types.go
+++ b/pkg/apis/k3k.io/v1alpha1/types.go
@@ -84,7 +84,10 @@ type ClusterSpec struct {
 	Expose *ExposeConfig `json:"expose,omitempty"`
 }
 
-// ClusterMode lists the two possible provisioning mode of a Cluster.
+// +kubebuilder:validation:Enum=shared;virtual
+// +kubebuilder:default="shared"
+//
+// ClusterMode is the possible provisioning mode of a Cluster.
 type ClusterMode string
 
 const (
--- a/pkg/apis/k3k.io/v1alpha1/types.go
+++ b/pkg/apis/k3k.io/v1alpha1/types.go
@@ -88,8 +88,8 @@ type ClusterSpec struct {
 type ClusterMode string
 
 const (
-	Shared  = ClusterMode("shared")
-	Virtual = ClusterMode("virtual")
+	SharedClusterMode  = ClusterMode("shared")
+	VirtualClusterMode = ClusterMode("virtual")
 )
 
 type ClusterLimit struct {
--- a/pkg/apis/k3k.io/v1alpha1/types.go
+++ b/pkg/apis/k3k.io/v1alpha1/types.go
@@ -21,53 +21,76 @@ type Cluster struct {
 type ClusterSpec struct {
 	// Version is a string representing the Kubernetes version to be used by the virtual nodes.
 	Version string `json:"version"`
-	// +kubebuilder:validation:XValidation:message="cluster must have at least one server",rule="self >= 1"
+
 	// Servers is the number of K3s pods to run in server (controlplane) mode.
+	// +kubebuilder:validation:XValidation:message="cluster must have at least one server",rule="self >= 1"
 	Servers *int32 `json:"servers"`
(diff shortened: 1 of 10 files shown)


pkg/controller/clusterset/clusterset.go
```diff
@@ -50,14 +51,34 @@ func Add(ctx context.Context, mgr manager.Manager, clusterCIDR string, logger *l
...
+	var clusterSet v1alpha1.ClusterSet
+	if err := c.Client.Get(ctx, req.NamespacedName, &clusterSet); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Check if a ClusterSet in the same namespace already exists.
+	// We retry in case of timeout because it could happen the Informer needs to sync
+	clusterSetList := &v1alpha1.ClusterSetList{}
+	listOpts := &ctrlruntimeclient.ListOptions{Namespace: req.Namespace}
+	err := retry.OnError(retry.DefaultBackoff, apierrors.IsTimeout, func() error {
```
briandowns, line 65: The repo standard for this return is to scope lock the error.

briandowns (approved):

galal-hussein:
The PR looks fine, we should however make sure that any cluster created under the clusterset is using the same mode as the clusterset and preventing multiple modes in the same set

pkg/apis/k3k.io/v1alpha1/types.go
```diff
@@ -21,54 +21,77 @@ type Cluster struct {
...
 	// clusterIP which is relatively secure, but difficult to access outside of the cluster.
+	// +optional
 	Expose *ExposeConfig `json:"expose,omitempty"`
 }
 
+// ClusterMode lists the two possible provisioning mode of a Cluster.
+type ClusterMode string
+
+const (
+	Shared  = ClusterMode("shared")
```
galal-hussein, line 91: Can we add more explicit name here, like SharedClusterMode so that it will be clear when used

pkg/controller/clusterset/clusterset.go
enrichman: Not sure what do you mean. Can you re-elaborate?

I had to add this retry because the tests were running probably too fast, and I was receiving an error about the cache not being synced yet.

pkg/apis/k3k.io/v1alpha1/types.go
enrichman: Yeah, sure!

galal-hussein:
> > he PR looks fine, we should however make sure that any cluster created under the clusterset is using the same mode as the clusterset and preventing multiple modes in the same set
> 
> @galal-hussein I was not sure if adding this logic in this PR but it make sense.
> 
> Question: what should happen if Clusters with different modes are already present?
> 
> IMHO the ClusterSet should report some a failing status. In the same fashion if a Cluster is created in a not-supported mode it should report a faliing Status.
> 
> We should report the same failing status also in the case of an already existing ClusterSet probably, without deleting it.
> 
> The best would be having a Webhook to prevent the creation of these resources, but not yet.
> 
> WDYT?

I think webhook might complicate  things, I would try to avoid webhook as much as we can, I think we should keep it simple, we shouldn't allow creating multiple clusters in the same namespace if there is no cluster set there to begin with, with that we will guarantee that we wont have a multiple modes in the same tenant

enrichman:
> I think webhook might complicate things, I would try to avoid webhook as much as we can, I think we should keep it simple, we shouldn't allow creating multiple clusters in the same namespace if there is no cluster set there to begin with, with that we will guarantee that we wont have a multiple modes in the same tenant

I'm still not convinced about this, a more declarative approach looks easier to understand. Otherwise we will need to delete resources, and the user will not understand why, until he looks at the logs. Without a webhook we cannot really prevent a user to create Clusters. They will, no error will appear, but then we will delete them if not valid.

enrichman:
@galal-hussein @briandowns 

Because it's still not clear how to properly handle some edge cases I've implemented the feature just reflecting what was in the RFC. It look like the field was meant to be an array of allowed types. At the moment it's probably easier to relax this constraint, until we don't have an idea on how to handle errors.

galal-hussein (approved):
Looks great, Thanks!!, LGTM

PR #151, opened 2024-11-28
updates to readability, standard practices, optimizations, etc

briandowns:


galal-hussein (approved):
LGTM, Thanks!

k3k-kubelet/controller/handler.go
```diff
@@ -109,7 +110,8 @@ func (c *ControllerHandler) RemoveResource(ctx context.Context, obj client.Objec
 	ctrl, ok := c.controllers[obj.GetObjectKind().GroupVersionKind()]
 	c.RUnlock()
 	if !ok {
-		return fmt.Errorf("no controller found for gvk" + obj.GetObjectKind().GroupVersionKind().String())
+		return errors.New("no controller found for gvk" + obj.GetObjectKind().GroupVersionKind().String())
```
enrichman, line 113: Missing a space:

```suggestion
		return errors.New("no controller found for gvk " + obj.GetObjectKind().GroupVersionKind().String())
```

k3k-kubelet/kubelet.go
```diff
@@ -273,35 +282,40 @@ func loadTLSConfig(ctx context.Context, hostClient ctrlruntimeclient.Client, clu
...
 	if err := retry.OnError(controller.Backoff, func(err error) bool {
 		return err != nil
 	}, func() error {
 		var err error
 		b, err = bootstrap.DecodedBootstrap(token, endpoint)
 		return err
 	}); err != nil {
-		return nil, fmt.Errorf("unable to decode bootstrap: %w", err)
+		return nil, err
 	}
```
enrichman, line 295: Since we are using this func a lot maybe it's more clear if we extract this in a common/util/error package:

```go
func IsNotNil(err error) bool {
    return err != nil
}
```

Then if we want to chain the err in the same line I like more to have short definition, so something like:

```go
decode := func() error {
	var err error
	b, err = bootstrap.DecodedBootstrap(token, endpoint)
	return err
}
if err := retry.OnError(controller.Backoff, errors.IsNotNil, decode); err != nil {
	return nil, err
}
```

 Just a personal nit. And not sure about the scope of `b` in the closure!

k3k-kubelet/provider/node.go
```diff
@@ -20,3 +23,72 @@ func (n *Node) Ping(context.Context) error {
...
+		{
+			Type:    v1.NodeInternalIP,
+			Address: ip,
+		},
+	}
+	node.Status.Capacity = v1.ResourceList{
+		v1.ResourceCPU:    resource.MustParse("8"),
+		v1.ResourceMemory: resource.MustParse("326350752922"),
+		v1.ResourcePods:   resource.MustParse("110"),
+	}
```
enrichman, line 90: Are this random? Maybe we should add a comment if we need to change them somehow.

k3k-kubelet/translate/host.go
```diff
@@ -96,10 +99,13 @@ func (t *ToHostTranslater) TranslateName(namespace string, name string) string {
 	// - a valid k8s name
 	// - idempotently calculatable
 	// - unique for this combination of name/namespace/cluster
-	namePrefix := fmt.Sprintf("%s-%s-%s", name, namespace, t.ClusterName)
+	namePrefix := name + "-" + namespace + "-" + t.ClusterName
+
 	// use + as a separator since it can't be in an object name
-	nameKey := fmt.Sprintf("%s+%s+%s", name, namespace, t.ClusterName)
+	nameKey := name + "+" + namespace + "+" + t.ClusterName
+
```
enrichman, line 106: In this case (with 2+ args) I find the `Sprintf` more readable, but that's a personal preference.

pkg/controller/cluster/cluster.go
```diff
@@ -39,6 +39,9 @@ const (
 	defaultClusterServiceCIDR    = "10.45.0.0/16"
 	defaultStoragePersistentSize = "1G"
 	memberRemovalTimeout         = time.Minute * 1
+
+	OperatingModeVirtual = "virtual"
+	OperatingModeShared  = "shared"
```
enrichman, line 44: Same as the first comment, we can now use the ClusterMode.

enrichman (changes requested):
Few comments, but generally LGTM. I'll request changes mostly for the ClusterMode const. :+1:

cli/cmds/cluster/create.go
```diff
@@ -226,8 +227,8 @@ func validateCreateFlags() error {
 	if cmds.Kubeconfig == "" && os.Getenv("KUBECONFIG") == "" {
 		return errors.New("empty kubeconfig")
 	}
-	if mode != "shared" && mode != "virtual" {
-		return errors.New(`mode should be one of "shared" or "virtual"`)
+	if mode != k3kcluster.OperatingModeShared && mode != k3kcluster.OperatingModeVirtual {
+		return errors.New("mode should be one of " + k3kcluster.OperatingModeShared + " or " + k3kcluster.OperatingModeVirtual)
```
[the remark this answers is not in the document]
briandowns: I'd agree, overkill.

k3k-kubelet/translate/host.go
briandowns: The trade off is an insane amount of heap and garbage collector pressure for no real gain when just concatenating strings.

enrichman: That's right. I'm still in favor of a better readability over small optimizations but in this case we are not "obfuscating" too much. :)

FYI (and future-me): I didn't know about the `-gcflags "-m -l"` flags. I did some experiment and if you create a small main.go and you run it with `go run -gcflags "-m -l" main.go` it will tell you if the variable escapes to the heap or not! Pretty cool.

k3k-kubelet/kubelet.go
briandowns: I'd probably be beneficial to move this out to another function but more so in another PR.

k3k-kubelet/provider/node.go
briandowns: This is copied from where it originally was in the code base. I didn't make and changes or decisions on the values.

cli/cmds/cluster/create.go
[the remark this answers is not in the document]
galal-hussein: yeah I think we  can skip this for now, LGTM for me too

enrichman (approved):

enrichman:
I think this was staling for too long. I'm going to close this to cleanup a bit the PRs.

PR #154, opened 2024-11-29
Initial networking support for shared mode

galal-hussein:
The PR does the following:

- Syncs all the services created in the virtual cluster to the host cluster
- Creates a coredns service for each virtual cluster
- In k3k-kubelet, it injects networking information to each pod created including:
  - connection information to the virtual cluster API.
  - DNS information to the coredns service created in the second point.

### DIFF
--- a/k3k-kubelet/controller/service.go
+++ b/k3k-kubelet/controller/service.go
@@ -25,7 +25,6 @@ const (
 	serviceFinalizerName    = "service.k3k.io/finalizer"
 )
 
-// TODO: change into a generic syncer
 type ServiceReconciler struct {
 	virtualClient    ctrlruntimeclient.Client
 	hostClient       ctrlruntimeclient.Client
@@ -64,7 +63,6 @@ func AddServiceSyncer(ctx context.Context, virtMgr, hostMgr manager.Manager, clu
 
 func (s *ServiceReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	log := s.logger.With("Cluster", s.clusterName, "Service", req.NamespacedName)
-	// skip kubernetes service
 	if req.Name == "kubernetes" || req.Name == "kube-dns" {
 		return reconcile.Result{}, nil
 	}
@@ -84,6 +82,8 @@ func (s *ServiceReconciler) Reconcile(ctx context.Context, req reconcile.Request
 	if err := controllerutil.SetControllerReference(&cluster, syncedService, s.HostScheme); err != nil {
 		return reconcile.Result{}, err
 	}
+
+	// handle deletion
 	if !virtService.DeletionTimestamp.IsZero() {
 		// deleting the synced service if exists
 		if err := s.hostClient.Delete(ctx, syncedService); err != nil {
@@ -98,6 +98,7 @@ func (s *ServiceReconciler) Reconcile(ctx context.Context, req reconcile.Request
 		}
 		return reconcile.Result{}, nil
 	}
+
 	// Add finalizer if it does not exist
 	if !controllerutil.ContainsFinalizer(&virtService, serviceFinalizerName) {
 		controllerutil.AddFinalizer(&virtService, serviceFinalizerName)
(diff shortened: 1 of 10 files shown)


k3k-kubelet/provider/provider.go
```diff
@@ -496,6 +503,47 @@ func (p *Provider) GetPods(ctx context.Context) ([]*corev1.Pod, error) {
...
+	}
+	// injecting cluster DNS IP to the pods except for coredns pod
+	if !strings.HasPrefix(podName, "coredns") {
+		pod.Spec.DNSPolicy = corev1.DNSNone
+		pod.Spec.DNSConfig = &corev1.PodDNSConfig{
+			Nameservers: []string{
+				p.dnsIP,
+			},
+			Searches: []string{
+				podNamespace + ".svc.cluster.local", "svc.cluster.local", "cluster.local", "hgalal.az",
```
enrichman, line 540:
```suggestion
				podNamespace + ".svc.cluster.local", "svc.cluster.local", "cluster.local",
```

k3k-kubelet/controller/service.go
```diff
@@ -0,0 +1,89 @@
...
+	if req.Name == "kubernetes" || req.Name == "kube-dns" {
+		return reconcile.Result{}, nil
+	}
+	var (
+		virtService v1.Service
+		hostService v1.Service
+	)
+	if err := s.virtualClient.Get(ctx, req.NamespacedName, &virtService); err != nil {
+		return reconcile.Result{}, err
+	}
```
enrichman, line 72: We are missing the cleanup of the syncedService here. If we do a Get and we receive a not-found it means the virtService was deleted, and we need to delete the syncedService as well.

k3k-kubelet/controller/service.go
```diff
@@ -0,0 +1,89 @@
...
+	syncedService := s.service(&virtService)
+	if err := s.hostClient.Get(ctx, types.NamespacedName{Name: syncedService.Name, Namespace: s.clusterNamespace}, &hostService); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("creating the service for the first time on the host cluster")
+			return reconcile.Result{}, s.hostClient.Create(ctx, syncedService)
+		}
+		return reconcile.Result{}, err
+	}
+	log.Info("updating service on the host cluster")
+	return reconcile.Result{}, s.hostClient.Update(ctx, syncedService)
```
enrichman, line 82: We should add the ownerRef to the created syncedService. We are not using for garbage collection since we are anyway listening to all the Service reconciliation, but it's better to add the reference, to flag we have created the service.

enrichman (changes requested):
I left some comments. CI is also failing. I have a small PR that should at least fix the tests.

k3k-kubelet/controller/service.go
galal-hussein: Got it, will fix

k3k-kubelet/controller/service.go
galal-hussein: yup, will add it

enrichman:
Dismissing the review because I'll be OOO and don't want to block this. :)

k3k-kubelet/controller/service.go
```diff
@@ -0,0 +1,126 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	serviceSyncerController = "service-syncer-controller"
+	maxConcurrentReconciles = 1
+	serviceFinalizerName    = "service.k3k.io/finalizer"
+)
+
+// TODO: change into a generic syncer
```
briandowns, line 28: Let's remove this in favor of an issue in the repo.

k3k-kubelet/kubelet.go
```diff
@@ -109,24 +115,50 @@ func newKubelet(ctx context.Context, c *config, logger *k3klog.Logger) (*kubelet
...
 
+	logger.Info("adding service syncer controller")
+	if k3kkubeletcontroller.AddServiceSyncer(ctx, virtualMgr, hostMgr, c.ClusterName, c.ClusterNamespace, k3klog.New(false)); err != nil {
+		return nil, fmt.Errorf("failed to add service syncer controller: %v", err)
+	}
+
+	clusterIP, err := clusterIP(ctx, c.AgentHostname, c.ClusterNamespace, hostClient)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to extract the clusterIP for the server service %s", err.Error())
```
briandowns, line 125: Let's use errors.New and the + operator here and anywhere we're concatenating strings.

briandowns (approved):

Issue #119, opened 2024-10-01
Virtual Kubelet: Node Metrics


MbolotSuse:
Currently, the Virtual Kubelet methods for pod metrics are stub methods. They should be fully implemented, so that users of the virtual cluster can get metrics on the pods running in the virtual cluster. The metrics API of the host cluster should be the source of truth for the current resource consumption, and the virtual kubelet should merely reflect what the host cluster reports for these values.

PR #163, opened 2024-12-19
Implementation of `GetStatsSummary` and `GetMetricsResource` for Virtual Kubelet

enrichman:
Ref:
- #119 
---

This PR implements the `GetStatsSummary` and `GetMetricsResource` methods for Virtual Kubelet.

To check the metrics these two commands can be run against the virtual cluster:

```
kubectl get --raw "/api/v1/nodes/k3k-example-cluster-kubelet/proxy/metrics/resource"
kubectl get --raw "/api/v1/nodes/k3k-example-cluster-kubelet/proxy/stats/summary"
```

The data from the Node are the one of the node where the virtual kubelet is running, while the pods are the one owned by the virtual clusters.

To use the `node` and `node/proxy` we'll need a new ClusterRole, and a ClusterRoleBinding for the kubelet. The ClusterRoleBinding will be updated from the K3k controller when a new cluster is created or deleted, adding or removing the kubelet ServiceAccount.

Note: the collector used is copied from the [Azure ACI virtual kubelet](https://github.com/virtual-kubelet/azure-aci/tree/master/pkg/metrics/collectors) implementation.

### DIFF
--- a/k3k-kubelet/provider/provider.go
+++ b/k3k-kubelet/provider/provider.go
@@ -34,7 +34,6 @@ import (
 	"k8s.io/client-go/tools/remotecommand"
 	"k8s.io/client-go/transport/spdy"
 	compbasemetrics "k8s.io/component-base/metrics"
-	metricset "k8s.io/metrics/pkg/client/clientset/versioned"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
@@ -48,7 +47,6 @@ type Provider struct {
 	VirtualClient    client.Client
 	ClientConfig     rest.Config
 	CoreClient       cv1.CoreV1Interface
-	MetricsClient    metricset.Interface // TODO: do we need this?
 	ClusterNamespace string
 	ClusterName      string
 	serverIP         string
--- a/k3k-kubelet/provider/provider.go
+++ b/k3k-kubelet/provider/provider.go
@@ -2,20 +2,24 @@ package provider
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 	"strconv"
 	"strings"
 
+	"github.com/pkg/errors"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/rancher/k3k/k3k-kubelet/controller"
+	"github.com/rancher/k3k/k3k-kubelet/provider/collectors"
 	"github.com/rancher/k3k/k3k-kubelet/translate"
 	"github.com/rancher/k3k/pkg/apis/k3k.io/v1alpha1"
 	k3klog "github.com/rancher/k3k/pkg/log"
 	"github.com/virtual-kubelet/virtual-kubelet/node/api"
 	"github.com/virtual-kubelet/virtual-kubelet/node/api/statsv1alpha1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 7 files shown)


k3k-kubelet/provider/provider.go
```diff
@@ -42,7 +48,7 @@ type Provider struct {
 	VirtualClient    client.Client
 	ClientConfig     rest.Config
 	CoreClient       cv1.CoreV1Interface
-	MetricsClient    metricset.Interface
+	MetricsClient    metricset.Interface // TODO: do we need this?
```
galal-hussein, line 51: I think it was just copied from the virtual-kubelet definition, I think we should remove it until we use it

k3k-kubelet/provider/provider.go
```diff
@@ -177,13 +185,96 @@ func (p *Provider) AttachToContainer(ctx context.Context, namespace, podName, co
...
 
 // GetStatsSummary gets the stats for the node, including running pods
-func (p *Provider) GetStatsSummary(context.Context) (*statsv1alpha1.Summary, error) {
-	return nil, fmt.Errorf("not implemented")
+func (p *Provider) GetStatsSummary(ctx context.Context) (*statsv1alpha1.Summary, error) {
+	p.logger.Debug("GetStatsSummary")
+
+	nodeList := &v1.NodeList{}
+	if err := p.CoreClient.RESTClient().Get().Resource("nodes").Do(ctx).Into(nodeList); err != nil {
+		return nil, fmt.Errorf("unable to get nodes of cluster %s in namespace %s: %w", p.ClusterName, p.ClusterNamespace, err)
```
galal-hussein, line 193: can we use errors.New() instead of fmt.Errorf, we started to standardize this across all the repo

pkg/controller/cluster/cluster.go
```diff
@@ -279,6 +287,56 @@ func (c *ClusterReconciler) server(ctx context.Context, cluster *v1alpha1.Cluste
 	return nil
 }
 
+func (c *ClusterReconciler) bindNodeProxyClusterRole(ctx context.Context, cluster *v1alpha1.Cluster) error {
```
galal-hussein, line 290: I dont think this should be added here, this is a part of the resources related to the shared agent, it should be added in the resources function of the shared agent implementation https://github.com/rancher/k3k/blob/main/pkg/controller/cluster/agent/shared.go#L67

k3k-kubelet/provider/provider.go
enrichman: Yes, it's just an interface that can be implemented if needed. Probably useful for testing, but not necessary at the moment.

k3k-kubelet/provider/provider.go
enrichman: Hm, they are not the same. If you are just printing a string they are equivalent (even though it depends about the number of arguments formatted, and the readability). But if you are formatting an error it's useful to use the `fmt.Errorf` with the `%w` verb. [Since Go 1.13](https://go.dev/blog/go1.13-errors#errors-in-go-113) it wraps the error, and the caller could eventually check for specific errors with the `errors.Is` or `errors.As`.

pkg/controller/cluster/cluster.go
enrichman: Yes, it's what I initially did but unfortunately it doesn't work, I forgot to add a comment about this. Basically the ClusterRole is a Cluster resource, and the Controller is Namespaced. That means it cannot create namespaced resource, and it fails with the following error: `cluster-scoped resource must not have a namespace-scoped owner`. :slightly_frowning_face: 

The only solution I found was to create the ClusterRole and the Binding during the Helm installation, and add/remove the subjects during the cluster creation/deletion.

galal-hussein: hmmm, interesting, thats fine then, should we have a check for the virtual mode vs shared mode, since we dont need to invoke the function in the agent mode

galal-hussein (approved):

Issue #117, opened 2024-10-01
Virtual Kubelet: Node Usage


MbolotSuse:
Currently, the Virtual Kubelet uses hardcoded values for the consumed/available capacity for the virtual node. This isn't very accurate, and isn't flexible to the different configurations that customers may use.

Instead, the Virtual Kubelet should list all of the available nodes (using the node selector), and combine the free resources into the available/consumed fields on the virtual node. This should update when several conditions occur, specifically:
- When a new node is added
- When an existing node is removed
- (Periodically) When resources are consumed/freed

PR #169, opened 2025-01-02
Add real node resources to virtual node

enrichman:
Ref:
- #117
---

This PR adds the real capacity and allocatable resources of the virtual node.

It will get all the nodes, filtered by the nodeSelector field if present, sum the capacity and the allocatable resources, and updating the virtual node. Since the nodes in the host cluster could change (adding or removing), the update will be done every `10s`.


galal-hussein (approved):
This looks great, Thank you!

PR #173, opened 2025-01-07
Test release workflows

enrichman:
This PR adds two workflows that can be used to create and deletes testing releases.

It will use [Goreleaser](https://goreleaser.com/) as a tool to create the release, and a dedicate configuration file for this "draft releases". The changes to the Dockerfiles are needed because of the different Docker build context used from Goreleaser.

### Release - Test Build

This workflow can be triggered manually to create the docker images, pushed to the Github Container registry (`ghcr.io`). It will also create a draft release where it will upload the binaries, including the `k3kcli`.

Note: the tag used by Goreleaser is a "fake" one, it will not be pushed to the remote to avoid polluting the git history.

### Release - Delete Draft

This workflow can be triggered manually to delete the previously created images and release. It needs the tag to delete as an input. It will check the existence of the draft release, and it will then delete the images and the release.


galal-hussein (approved):

enrichman:
Ok, I've tested the images with an Helm installation and it worked. Merging this, thanks.

PR #179, opened 2025-01-10
Add PVC syncing support

galal-hussein:
- Adds PVC syncing support by default in shared mode.
- Creates a webhook to assign `spec.nodeName` to the virtual-kubelet node name.
- Automatically generates the certs bundle and ca bundle for the webhook configuration.
- The webhook will run in k3k-kubelet process and will listen to port 9443.
- Automatically creates the mutation webhook configuration to the virtual cluster and assign the CA bundle to it.

### DIFF
--- a/k3k-kubelet/controller/persistentvolumeclaims.go
+++ b/k3k-kubelet/controller/persistentvolumeclaims.go
@@ -1,0 +1,125 @@
+package controller
+
+import (
+	"context"
+
+	"github.com/rancher/k3k/k3k-kubelet/translate"
+	"github.com/rancher/k3k/pkg/apis/k3k.io/v1alpha1"
+	"github.com/rancher/k3k/pkg/log"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	pvcController    = "pvc-syncer-controller"
+	pvcFinalizerName = "pv.k3k.io/finalizer"
+)
+
+type PVCReconciler struct {
+	virtualClient    ctrlruntimeclient.Client
+	hostClient       ctrlruntimeclient.Client
+	clusterName      string
+	clusterNamespace string
+	Scheme           *runtime.Scheme
+	HostScheme       *runtime.Scheme
+	logger           *log.Logger
+	Translater       translate.ToHostTranslater
+	//objs             sets.Set[types.NamespacedName]
+}
+
+// AddPVCSyncer adds persistentvolumeclaims syncer controller to k3k-kubelet
+func AddPVCSyncer(ctx context.Context, virtMgr, hostMgr manager.Manager, clusterName, clusterNamespace string, logger *log.Logger) error {
+	translater := translate.ToHostTranslater{
+		ClusterName:      clusterName,
+		ClusterNamespace: clusterNamespace,
+	}
+	// initialize a new Reconciler
+	reconciler := PVCReconciler{
+		virtualClient:    virtMgr.GetClient(),
+		hostClient:       hostMgr.GetClient(),
+		Scheme:           virtMgr.GetScheme(),
+		HostScheme:       hostMgr.GetScheme(),
+		logger:           logger.Named(pvcController),
+		Translater:       translater,
+		clusterName:      clusterName,
+		clusterNamespace: clusterNamespace,
+	}
+	return ctrl.NewControllerManagedBy(virtMgr).
+		For(&v1.PersistentVolumeClaim{}).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: maxConcurrentReconciles,
+		}).
+		Complete(&reconciler)
+}
+
+func (v *PVCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := v.logger.With("Cluster", v.clusterName, "PersistentVolumeClaim", req.NamespacedName)
+	var (
+		virtPVC v1.PersistentVolumeClaim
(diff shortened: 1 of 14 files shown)


k3k-kubelet/controller/persistentvolumeclaims.go
```diff
@@ -0,0 +1,125 @@
...
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	pvcController    = "pvc-syncer-controller"
+	pvcFinalizerName = "pv.k3k.io/finalizer"
```
enrichman, line 23: Should this be `pvc.k3k.io/finalizer`?

```suggestion
	pvcFinalizerName = "pvc.k3k.io/finalizer"
```

k3k-kubelet/controller/persistentvolumeclaims.go
```diff
@@ -0,0 +1,125 @@
...
+type PVCReconciler struct {
+	virtualClient    ctrlruntimeclient.Client
+	hostClient       ctrlruntimeclient.Client
+	clusterName      string
+	clusterNamespace string
+	Scheme           *runtime.Scheme
+	HostScheme       *runtime.Scheme
+	logger           *log.Logger
+	Translater       translate.ToHostTranslater
+	//objs             sets.Set[types.NamespacedName]
```
enrichman, line 35: Can we drop the comment?

```suggestion
```

k3k-kubelet/controller/persistentvolumeclaims.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	return ctrl.NewControllerManagedBy(virtMgr).
+		For(&v1.PersistentVolumeClaim{}).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: maxConcurrentReconciles,
+		}).
+		Complete(&reconciler)
+}
+
+func (v *PVCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
```
enrichman, line 63: This is probably a nit but I had some issues to understand what `v` was. Maybe an `r` for Reconciler` is clearer?

k3k-kubelet/controller/persistentvolumeclaims.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	log.Info("updating pvc on the host cluster")
+	return reconcile.Result{}, v.hostClient.Update(ctx, syncedPVC)
+
+}
+
+func (v *PVCReconciler) pvc(obj *v1.PersistentVolumeClaim) *v1.PersistentVolumeClaim {
+	hostPVC := obj.DeepCopy()
+	v.Translater.TranslateTo(hostPVC)
+	// don't sync finalizers to the host
```
enrichman, line 123: Does it mean we are still missing to remove the finalizers here?

k3k-kubelet/controller/persistentvolumeclaims.go
```diff
@@ -0,0 +1,125 @@
...
+	syncedPVC := v.pvc(&virtPVC)
+	if err := controllerutil.SetControllerReference(&cluster, syncedPVC, v.HostScheme); err != nil {
+		return reconcile.Result{}, err
+	}
+	// handle deletion
+	if !virtPVC.DeletionTimestamp.IsZero() {
+		// deleting the synced service if exists
+		if err := v.hostClient.Delete(ctx, syncedPVC); err != nil {
+			return reconcile.Result{}, ctrlruntimeclient.IgnoreNotFound(err)
+		}
```
enrichman, line 87: Hm, I think this could lead to some orphaned resource. I.e. what if the syncedPVC does not exist? This will error, and it will not be requed because of the `IgnoreNotFound`.

Maybe it's better to return only if the error is a different one. In case of ErrNotFound we should continue the deletion of the virtual PVC.

k3k-kubelet/controller/persistentvolumeclaims.go
```diff
@@ -0,0 +1,125 @@
...
+	// getting the cluster for setting the controller reference
+
+	// Add finalizer if it does not exist
+	if !controllerutil.ContainsFinalizer(&virtPVC, pvcFinalizerName) {
+		controllerutil.AddFinalizer(&virtPVC, pvcFinalizerName)
+		if err := v.virtualClient.Update(ctx, &virtPVC); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+	// create or update the pv on host
```
enrichman, line 107:
```suggestion
	// create or update the pvc on host
```

enrichman (commented):
Left some comments, it looks ok, I'm just not completely aware of the TLS/cert stuff, but just because I'm not an expert.

IIRC `cert-manager` is a required dependency, or maybe not yet, and it will be required for the CAPI provider? Could it simplify that part?

k3k-kubelet/controller/persistentvolumeclaims.go
galal-hussein: I have no idea why its v :D

k3k-kubelet/controller/persistentvolumeclaims.go
galal-hussein: no this is an old comment probably just copied by mistake, the finalizers are removed in the translateTo function now

k3k-kubelet/controller/webhook/pod.go
```diff
@@ -0,0 +1,120 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+const (
+	webhookName    = "nodename.podmutator.k3k.io"
+	webhookTimeout = int32(10)
+	webhookPort    = "9443"
+	webhookPath    = "/mutate--v1-pod"
```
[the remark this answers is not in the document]
galal-hussein: yeah unfortunately, its supposed to be the apigroup, but since this is a pod mutator the default is nil hence the --, we can have a custom path but its a bit of a hassle to change it

galal-hussein:
> Left some comments, it looks ok, I'm just not completely aware of the TLS/cert stuff, but just because I'm not an expert.
> 
> IIRC `cert-manager` is a required dependency, or maybe not yet, and it will be required for the CAPI provider? Could it simplify that part?

The cert manager is required if the hook is for the host cluster, but since we manage everything for the virtual cluster we can automatically just create an ad-hook cert bundle using dynamic listener and inject the configuration to the virtual cluster, this is much easier in my opinion

PR #180, opened 2025-01-10
Added priorityClass to Clusters and ClusterSets

enrichman:
This PR adds the `defaultPriorityClass` to ClusterSets and the `priorityClass` to the Cluster.

It adds a Watch to the ClusterSet controller that will trigger when a Cluster is changed. When this happens then all the ClusterSet in the namespace will be reconciled.

Tests.

Ref #121


galal-hussein (approved):

galal-hussein:
ref #295

Issue #184, opened 2025-01-15
Pods are not registering with k3k-kubelet if they have a fieldpath status.podIP


galal-hussein:
If pod information specifically `status.podIP` is exposed either by defining it as a downward api volume or an environment variable for a pod, then the pod will be stuck in a pending state forever, this maybe due to the a chicken and egg problem where the provider is waiting for the pod to be moved out of pending state to schedule it and the pod itself is waiting for kubelet to get an IP, one way to solve this is through the pod mutation webhook where we can remove the status pod IP and re-add it when the provider is translating the pod

PR #185, opened 2025-01-15
Fix downward api for status fields in k3k-kubelet

galal-hussein:
- #184

### DIFF
--- a/k3k-kubelet/controller/webhook/pod.go
+++ b/k3k-kubelet/controller/webhook/pod.go
@@ -77,17 +77,18 @@ func (w *webhookHandler) Default(ctx context.Context, obj runtime.Object) error 
 	}
 	for i, container := range pod.Spec.Containers {
 		for j, env := range container.Env {
-			if env.ValueFrom != nil {
-				if env.ValueFrom.FieldRef != nil {
-					if strings.Contains(env.ValueFrom.FieldRef.FieldPath, "status.") {
-						pod.Annotations[FieldpathField+"_"+strconv.Itoa(i)+"_"+env.Name] = env.ValueFrom.FieldRef.FieldPath
-						pod.Spec.Containers[i].Env = removeEnv(pod.Spec.Containers[i].Env, j)
-					}
-				}
+			if env.ValueFrom == nil || env.ValueFrom.FieldRef == nil {
+				continue
+			}
+
+			fieldPath := env.ValueFrom.FieldRef.FieldPath
+			if strings.Contains(fieldPath, "status.") {
+				annotationKey := fmt.Sprintf("%s_%d_%s", FieldpathField, i, env.Name)
+				pod.Annotations[annotationKey] = fieldPath
+				pod.Spec.Containers[i].Env = removeEnv(pod.Spec.Containers[i].Env, j)
 			}
 		}
 	}
-
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


k3k-kubelet/controller/webhook/pod.go
```diff
@@ -63,10 +67,27 @@ func (w *webhookHandler) Default(ctx context.Context, obj runtime.Object) error
...
+			if env.ValueFrom != nil {
+				if env.ValueFrom.FieldRef != nil {
+					if strings.Contains(env.ValueFrom.FieldRef.FieldPath, "status.") {
+						pod.Annotations[FieldpathField+"_"+strconv.Itoa(i)+"_"+env.Name] = env.ValueFrom.FieldRef.FieldPath
+						pod.Spec.Containers[i].Env = removeEnv(pod.Spec.Containers[i].Env, j)
+					}
+				}
+			}
+		}
+	}
```
enrichman, line 89: When possible I try to avoid too many nested ifs, what about a small rewrite like:

```suggestion
	// look for status.* fields in the env
	if pod.Annotations == nil {
		pod.Annotations = make(map[string]string)
	}
	for i, container := range pod.Spec.Containers {
		for j, env := range container.Env {
			if env.ValueFrom == nil || env.ValueFrom.FieldRef == nil {
				continue
			}
			
			fieldPath := env.ValueFrom.FieldRef.FieldPath
			if strings.Contains(fieldPath, "status.") {
				annotationKey := fmt.Sprintf("%s_%d_%s", FieldpathField, i, env.Name)
				pod.Annotations[annotationKey] = fieldPath
				pod.Spec.Containers[i].Env = removeEnv(pod.Spec.Containers[i].Env, j)
			}
		}
	}
```

Also because of the strconv and all the params I find cleaner to use a `fmt.Sprintf`.

I would also move this logic in a separate func, with a specific name and a small comment on why it's needed. Something like `cleanupFieldRef`.

k3k-kubelet/provider/provider.go
```diff
@@ -665,3 +670,37 @@ func getSecretsAndConfigmaps(pod *corev1.Pod) ([]string, []string) {
...
+		if strings.Contains(name, webhook.FieldpathField) {
+			s := strings.SplitN(name, "_", 3)
+			if len(s) != 3 {
+				continue
+			}
+			containerIndex, err = strconv.Atoi(s[1])
+			if err != nil {
+				return err
+			}
+			envName = s[2]
```
enrichman, line 692: I wouldn't try to spread outside this logic, it looks part of the webhook fieldRef stuff. Would it make sense to have inside the `webhook` package two func to handle this?

Something like:

```
func FormatFieldPathAnnotationKey() {
    // formatting logic here
}

func ParseFieldPathAnnotationKey(annotationKey string) (name string, containerIndex int, envName string, error) {
    // parse logic here
}
```

The cleanest way would be have a `FieldPathAnnotationKey` type, but this is up to you. If it's used only here it's fine enough.

k3k-kubelet/provider/provider.go
```diff
@@ -665,3 +670,37 @@ func getSecretsAndConfigmaps(pod *corev1.Pod) ([]string, []string) {
...
+
+			// re-adding these envs to the pod
```
enrichman, line 694: After readding them, should we cleanup the annotations as well? Do we need them anymore?

enrichman (commented):
A couple of comments!

galal-hussein:
> A couple of comments!

Thanks I will address the comments today

k3k-kubelet/provider/provider.go
galal-hussein: ah okay, I can add the logic in the webhook package

k3k-kubelet/provider/provider.go
galal-hussein: I will remove them from the pods on the host cluster, removing them after they are created on the virtual cluster will require a new controller which I think its an overkill

enrichman (approved):

Issue #168, opened 2024-12-27
Rancher Kubectl shell


jp-gouin:
Once a K3K cluster in shared mode is imported in Rancher, you cannot start a `Kubectl Shell` 

The `dashboard-shell` pod fails in the k3k cluster with :

```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2024-12-27T14:59:31Z"
    status: "True"
    type: PodScheduled
  message: 'unable to sync volumes for pod cattle-system/dashboard-shell-dpbf6: unable
    to sync configmap volume admin-kubeconfig: unable to add configmap to sync cattle-system/impersonation-shell-admin-kubeconfig-ldjwk:
    unable to reconcile new object cattle-system/impersonation-shell-admin-kubeconfig-ldjwk:
    unable to create host configmap test2/impersonation-shell-admin-kubeconfig-ldjwk-cattle-system-d9a25e
    for virtual configmap cattle-system/impersonation-shell-admin-kubeconfig-ldjwk:
    %!w(<nil>)'
```  

`impersonation-shell-admin-kubeconfig-ldjwk-cattle-system-d9a25e` is created on the host. But it's not present in the k3k cluster.

VestigeJ:
Are you running rancher on the host cluster or external to the nodes running the host/virtual/shared cluster for this?

VestigeJ:
Able to import k3k clusters of both mode types still unable to get a kubectl shell into one though..

jp-gouin:
@VestigeJ sorry I forgot to reply

Rancher is running externally in my case and when I try to start the shell from Rancher on a virtual cluster in shared mode it fails

rossella:
let's retest after conformance tests are passing.

PR #188, opened 2025-01-16
Add retry for k3k-kubelet provider functions

galal-hussein:
the PR will add a retry logic to create/update/delete implementations of k3k-kubelet to allow creation of single pods

- https://github.com/rancher/k3k/issues/168

### DIFF
--- a/k3k-kubelet/provider/provider.go
+++ b/k3k-kubelet/provider/provider.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/pkg/errors"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/rancher/k3k/k3k-kubelet/controller"
 	"github.com/rancher/k3k/k3k-kubelet/controller/webhook"
@@ -32,6 +31,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes/scheme"
 	cv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"errors"
 
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/portforward"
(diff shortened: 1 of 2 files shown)


VestigeJ (approved):
LGTM

k3k-kubelet/provider/provider.go
```diff
@@ -364,6 +376,26 @@ func (p *Provider) CreatePod(ctx context.Context, pod *corev1.Pod) error {
...
+		})
+	if err != nil {
+		const message = "failed to create pod with provider"
+		if lastErr != nil {
+			return errors.Wrap(lastErr, message)
+		}
+		return errors.Wrap(err, message)
+	}
+	return nil
+}
```
enrichman, line 397: Is ther any reason on why we are not using the same original `ctx`? If that one was already canceled it means we are not going to respect it.

Regarding the errors I was thinking that we could keep all the errors, instead of just the last one, and if not nil we could log them, just to see that even if the operation finally succeeded, there were some retries.

Also it's nice to have the final retry error exported, so when wrapped a caller could eventually check if the error was due to a retry timeout.

```go
var ErrRetryTimeout = errors.New("provider retry timed out")
```

The last bit is that it looks like we are not going to change the interval or timeout, so we can avoid to pass them as arguments.

Maybe something like:

```suggestion
// withRetry retries passed function with interval and timeout
func (p *Provider) withRetry(ctx context.Context, f func(context.Context, *v1.Pod) error, pod *v1.Pod) error {
	const (
		interval = 2 * time.Second
		timeout  = 10 * time.Second
	)

	var allErrors error

	// retryFn will retry until the operation succeed, or the timeout occurs
	retryFn := func(ctx context.Context) (bool, error) {
		if lastErr = f(ctx, pod); lastErr != nil {
			// log that the retry failed?
			allErrors = errors.Join(allErrors, lastErr)
			return false, nil
		}
		return true, nil
	}

	if err := wait.PollUntilContextTimeout(ctx, interval, timeout, true, retryFn); err != nil {
		return errors.Join(allErrors, ErrRetryTimeout)
	}
	return nil
}
```

enrichman (commented):
Do we have often this issue? I left a comment, I'm not sure if 2s within 10s is too much time. What about a quicker polling of 1s in 5s?

k3k-kubelet/provider/provider.go
galal-hussein: That makes more sense, I will change it, also I think util/errors in k8s.io has a helper method called aggregate errors `errorsutil.AggregateError` which can help instead of doing them manually

as for the context background I think it was just a mistake on my part, I will address the comments

enrichman: Thank you! I'll try to review it ASAP to get it merged before EOD :smile:

enrichman (approved):

Issue #166, opened 2024-12-24
CI testing


jp-gouin:
Our current CI pipeline lacks sufficient end-to-end test coverage for virtual cluster deployments, in shared and virtual modes. This deficiency increases the risk of undetected regressions and integration issues impacting production deployments.

We should considered including : 
- deployment tests for `virtual` and `shared` mode
- accessibility checks
- isolation tests 
  - cluster in a same `clusterset` are visible
  - cluster in a different `clusterset` are isolated
- Conformance tests

jp-gouin:
Conformance test should run nightly

PR #192, opened 2025-01-17
Add Cluster creation test

enrichman:
Ref:
- #166 
---

This PR adds a first Cluster creation test.

It will check for the readiness state of the server and kubelet, and it will wait for the bootstrap secret to be created.

Note that we are not yet checking for it to be "functional". A next PR will be made, after a small check and refactor of the kubeconfig generation.

:warning: **Warning** :warning: 

The `PullPolicy` for the kubelet was removed, with the default `IfNotPresent`. This is because we are pushing dev images locally, and they do not exists in the remote registry.

For testing development you should be aware that the image in your environment could be cached, and other tags should be used instead. Using the `latest` tag should enable the `Always` policy.
We could think adding a dedicated environment variable, or enable this strategy when the `debug` is true.


tests/cluster_test.go
```diff
@@ -0,0 +1,88 @@
...
+				case "rancher/k3k-kubelet":
+					kubeletRunning = pod.Status.Phase == corev1.PodRunning
+				}
+
+				if serverRunning && kubeletRunning {
+					return true
+				}
+			}
+
+			return false
```
jp-gouin, line 71: Apologies for the noob question as I'm not super familiar with go test :) 
If it's false , do we want to give more context (e.g logs?)

enrichman: Probably not in this case. This is a continuous loop that will check for the readiness state, until the timeout occurs.

If the timed out we know that the pods were not Ready, and at the end of the tests the cluster logs will be available to download: https://github.com/rancher/k3k/actions/runs/12827942200/artifacts/2446381844

Those are very convenient to check why the tests failed!

image

galal-hussein (approved):

PR #193, opened 2025-01-20
Simplify Cluster spec

enrichman:
This PR removes some required parameters, adding defaults. Now it's possible to create a Cluster with an empty spec:

```yaml
apiVersion: k3k.io/v1alpha1
kind: Cluster
metadata:
  name: mycluster
  namespace: k3k-mycluster
```

**mode**: default  to `shared`
**servers**: default to `1`
**agents**: default to `0`
**version**: if not provided the version used will be the same of the host. To do so the Reconciler will use the DiscoveryClient, and it will update the Cluster **Status** subresource. A field `HostVersion` will be used, this will prevent us to overwrite the submitted spec, keeping track the original submitted resource.

### DIFF
--- a/pkg/apis/k3k.io/v1alpha1/types.go
+++ b/pkg/apis/k3k.io/v1alpha1/types.go
@@ -15,28 +15,34 @@ type Cluster struct {
 	metav1.TypeMeta   `json:",inline"`
 
 	// +kubebuilder:default={}
+	// +optional
 	Spec   ClusterSpec   `json:"spec"`
 	Status ClusterStatus `json:"status,omitempty"`
 }
 
 type ClusterSpec struct {
 	// Version is a string representing the Kubernetes version to be used by the virtual nodes.
-	Version string `json:"version,omitempty"`
+	//
+	// +optional
+	Version string `json:"version"`
 
 	// Servers is the number of K3s pods to run in server (controlplane) mode.
 	//
 	// +kubebuilder:default=1
 	// +kubebuilder:validation:XValidation:message="cluster must have at least one server",rule="self >= 1"
-	Servers *int32 `json:"servers,omitempty"`
+	// +optional
+	Servers *int32 `json:"servers"`
 
 	// Agents is the number of K3s pods to run in agent (worker) mode.
 	//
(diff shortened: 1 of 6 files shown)


pkg/apis/k3k.io/v1alpha1/types.go
```diff
@@ -14,21 +14,26 @@ type Cluster struct {
...
 
+	// +kubebuilder:default={}
 	Spec   ClusterSpec   `json:"spec"`
 	Status ClusterStatus `json:"status,omitempty"`
 }
 
 type ClusterSpec struct {
 	// Version is a string representing the Kubernetes version to be used by the virtual nodes.
-	Version string `json:"version"`
+	Version string `json:"version,omitempty"`
```
bigkevmcd, line 24: What happens if you don't set the `Version`?

It looks like https://github.com/rancher/k3k/blob/e7671134d26c94e950bde553e307702d0ad90ffd/pkg/controller/controller.go#L30 would generate an invalid image reference?

So, this should probably be `// +required` and **not** `omitempty` ?

enrichman: @bigkevmcd that's correct. The idea was to simplify all the possible fields, and the empty version will fallback to the same host k3s version. To handle this the controller will check this, update, and requeue the resource.

https://github.com/rancher/k3k/blob/a00c9e243a0e35e8a0b0bec6ee9c6379d9379561/pkg/controller/cluster/cluster.go#L90-L102

I'm not completely sold on this, so I'm happy to discuss any possible issue with that.

On the UX perspective this is something that can be handled via the UI, or with the `k3kcli`, keeping the version as the only required field.

bigkevmcd: Feels like it's better to prevent it being created/updated, than to fail to reconcile?

bigkevmcd: I think what I'd do here, is leave the spec.version empty, but populate the status with the version.

galal-hussein (approved):
LGTM, Thank you!

PR #197, opened 2025-01-20
Added `golangci-lint` action

enrichman:
This PR adds the `golangci-lint` action to perform a lint with tests.

It enables only the "default" linters (fixing an issue). Other linters will be added in the future.


andreas-kupries (approved):

k3k-kubelet/controller/handler.go
```diff
@@ -109,7 +109,7 @@ func (c *ControllerHandler) RemoveResource(ctx context.Context, obj client.Objec
 	ctrl, ok := c.controllers[obj.GetObjectKind().GroupVersionKind()]
 	c.RUnlock()
 	if !ok {
-		return fmt.Errorf("no controller found for gvk" + obj.GetObjectKind().GroupVersionKind().String())
+		return fmt.Errorf("no controller found for gvk %s", obj.GetObjectKind().GroupVersionKind().String())
```
bigkevmcd, line 112: You don't need `.String()` here because `%s` will implicitly call it anyway.

bigkevmcd (approved):

k3k-kubelet/controller/handler.go
enrichman: Ah, correct! Thanks!

PR #200, opened 2025-01-22
Fix secret tokens and DNS translation

galal-hussein:
This PR fixes few bugs as well as adds few features:

bug fixes:
- fix token translation in init containers
- fix dns translation in init containers
- fix dns translation fo kubernetes.defaul service
- fixes cluster imports in rancher

Features:
- Adds node version to the shared agent node registration
- Add SKIP_TEST var to facilitate dev testing
- Add image pull policy var to the shared agent image

### DIFF
--- a/k3k-kubelet/config.go
+++ b/k3k-kubelet/config.go
@@ -18,6 +18,7 @@ type config struct {
 	VirtualConfigPath string `yaml:"virtualConfigPath,omitempty"`
 	KubeletPort       string `yaml:"kubeletPort,omitempty"`
 	ServerIP          string `yaml:"serverIP,omitempty"`
+	Version           string `yaml:"version,omitempty"`
 }
 
 func (c *config) unmarshalYAML(data []byte) error {
@@ -53,6 +54,9 @@ func (c *config) unmarshalYAML(data []byte) error {
 	}
 	if c.ServerIP == "" {
 		c.ServerIP = conf.ServerIP
+	}
+	if c.Version == "" {
+		c.Version = conf.Version
 	}
 	return nil
 }
--- a/k3k-kubelet/provider/provider.go
+++ b/k3k-kubelet/provider/provider.go
@@ -369,7 +369,7 @@ func (p *Provider) createPod(ctx context.Context, pod *corev1.Pod) error {
 		return fmt.Errorf("unable to transform tokens for pod %s/%s: %w", pod.Namespace, pod.Name, err)
 	}
 	// inject networking information to the pod including the virtual cluster controlplane endpoint
-	p.configureNetworking(pod.Name, pod.Namespace, tPod)
+	p.configureNetworking(pod.Name, pod.Namespace, tPod, p.serverIP)
 
 	p.logger.Infow("Creating pod", "Host Namespace", tPod.Namespace, "Host Name", tPod.Name,
 		"Virtual Namespace", pod.Namespace, "Virtual Name", "env", pod.Name, pod.Spec.Containers[0].Env)
@@ -475,6 +475,7 @@ func (p *Provider) syncConfigmap(ctx context.Context, podNamespace string, confi
 
 // syncSecret will add the secret object to the queue of the syncer controller to be synced to the host cluster
 func (p *Provider) syncSecret(ctx context.Context, podNamespace string, secretName string, optional bool) error {
+	p.logger.Infow("Syncing secret", "Name", secretName, "Namespace", podNamespace, "optional", optional)
 	var secret corev1.Secret
 	nsName := types.NamespacedName{
 		Namespace: podNamespace,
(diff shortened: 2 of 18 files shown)


main.go
```diff
@@ -50,6 +51,12 @@ var (
 			Value:       "rancher/k3k:k3k-kubelet-dev",
 			Destination: &sharedAgentImage,
 		},
+		cli.StringFlag{
+			Name:        "shared-agent-pull-policy",
+			EnvVar:      "SHARED_AGENT_PULL_POLICY",
+			Usage:       "K3K Virtual Kubelet image pull policy",
+			Destination: &sharedAgentImagePullPolicy,
+		},
```
enrichman, line 59: Since this is an enum, I would put at least a quick check of the provided value. With `urfave/cli/v2` there is a Flag Action we can use, but for now we can stick for a manual check.

pkg/controller/cluster/agent/shared.go
```diff
@@ -26,18 +26,20 @@ const (
...
 
 type SharedAgent struct {
-	cluster          *v1alpha1.Cluster
-	serviceIP        string
-	sharedAgentImage string
-	token            string
+	cluster                    *v1alpha1.Cluster
+	serviceIP                  string
+	sharedAgentImage           string
+	sharedAgentImagePullPolicy string
```
enrichman, line 32: We can omit the redundant prefix:

```suggestion
	image           string
	imagePullPolicy string
```

pkg/controller/cluster/agent/shared.go
```diff
@@ -60,13 +62,18 @@ func (s *SharedAgent) Config() ctrlruntimeclient.Object {
 }
 
 func sharedAgentData(cluster *v1alpha1.Cluster, token, nodeName, ip string) string {
+	version := cluster.Spec.Version
+	if cluster.Spec.Version == "" {
+		version = cluster.Status.HostVersion
+	}
```
enrichman, line 68: Looking a bit for a different solution I found a very interesting one I didn't know about!

```suggestion
	version := cmp.Or(cluster.Spec.Version, cluster.Status.HostVersion) 
```

k3k-kubelet/provider/provider.go
```diff
@@ -733,6 +740,31 @@ func (p *Provider) configureNetworking(podName, podNamespace string, pod *corev1
...
+			},
+			corev1.EnvVar{
+				Name:  "KUBERNETES_SERVICE_HOST",
+				Value: p.serverIP,
+			},
+			corev1.EnvVar{
+				Name:  "KUBERNETES_SERVICE_PORT",
+				Value: "6443",
+			},
+		)
```
enrichman, line 766: I've seen we have already a problem with something like this. If the env vars are already set then we can have an issue.

Instead of just adding the EnvVar we should check if the same var already exists. If exists we should override it. Having multiple env vars with the same name could be risky, and unclear.

I would create a dedicated func for this, and fix also the other environment variables in the service syncer (IIRC).

k3k-kubelet/provider/provider.go
```diff
@@ -707,7 +708,13 @@ func (p *Provider) GetPods(ctx context.Context) ([]*corev1.Pod, error) {
 // configureNetworking will inject network information to each pod to connect them to the
 // virtual cluster api server, as well as confiugre DNS information to connect them to the
 // synced coredns on the host cluster.
-func (p *Provider) configureNetworking(podName, podNamespace string, pod *corev1.Pod) {
+func (p *Provider) configureNetworking(podName, podNamespace string, pod *corev1.Pod, serverIP string) {
+	// inject serverIP to hostalias for the pod
+	KubernetesHostAlias := corev1.HostAlias{
+		IP:        serverIP,
+		Hostnames: []string{"kubernetes", "kubernetes.default", "kubernetes.default.svc.cluster.local"},
```
enrichman, line 715: I'm not sure if we should need to add also the two other possible hostnames, such as `"kubernetes.default.svc"` and `"kubernetes.default.svc.cluster"`.

k3k-kubelet/config.go
```diff
@@ -54,6 +55,9 @@ func (c *config) unmarshalYAML(data []byte) error {
 	if c.ServerIP == "" {
 		c.ServerIP = conf.ServerIP
 	}
+	if c.Version == "" {
+		c.Version = conf.Version
+	}
```
enrichman, line 60: I would use the new `cmp.Or` func to simplify this very long chain of ifs:

```suggestion
	c.Version = cmp.Or(c.Version, conf.Version)
```

enrichman (commented):
Thanks, I left some comments!

k3k-kubelet/config.go
galal-hussein: I am actually trying to think of a different way altogether to simplify this function, it just piled up with a lot of options to virtual kubelet

k3k-kubelet/provider/provider.go
galal-hussein: yeah I think we should, I will add them

k3k-kubelet/provider/provider.go
galal-hussein: I think for the KUBERNETES_* envs, we shouldn't check if they exist, because they are always being set in any pod, maybe for other vars but we need to override the KUBERNETES_* vars anyway

pkg/controller/cluster/agent/shared.go
galal-hussein: okidoke

main.go
galal-hussein: we have a validation function, we should implement some kind of conditions to make sure we have a the right values

pkg/controller/cluster/agent/shared.go
galal-hussein: yeah it seems that they have added that in go 1.22, however I am not sure it is the best use case for our situation, since its just a simple comparison of strings and it really doesnt need a new layer of abstraction

k3k-kubelet/provider/provider.go
enrichman: What I meant is that we need to override the already existing value. Otherwise we get a WARN (info) in the logs:

```
{"level":"info","timestamp":"2025-01-22T17:46:30.655Z","logger":"KubeAPIWarningLogger","msg":"spec.containers[0].env[8]: hides previous definition of \"KUBERNETES_PORT_443_TCP\", which may be dropped when using apply"}
{"level":"info","timestamp":"2025-01-22T17:46:30.655Z","logger":"KubeAPIWarningLogger","msg":"spec.containers[0].env[9]: hides previous definition of \"KUBERNETES_PORT\", which may be dropped when using apply"}
{"level":"info","timestamp":"2025-01-22T17:46:30.655Z","logger":"KubeAPIWarningLogger","msg":"spec.containers[0].env[10]: hides previous definition of \"KUBERNETES_PORT_443_TCP_ADDR\", which may be dropped when using apply"}
{"level":"info","timestamp":"2025-01-22T17:46:30.655Z","logger":"KubeAPIWarningLogger","msg":"spec.containers[0].env[11]: hides previous definition of \"KUBERNETES_SERVICE_HOST\", which may be dropped when using apply"}
{"level":"info","timestamp":"2025-01-22T17:46:30.655Z","logger":"KubeAPIWarningLogger","msg":"spec.containers[0].env[12]: hides previous definition of \"KUBERNETES_SERVICE_PORT\", which may be dropped when using apply"}
```

image

galal-hussein: ah I see

enrichman (approved):
Thanks, LGTM!

Issue #170, opened 2025-01-03
[EPIC] Resiliency in shared mode


jp-gouin:
The `kubelet` and the `server` should be resilient to pod restart.
The `kubelet` has issue restarting due to the webhook cert
The `server` has issue restarting due to the `boostrap` secret

### Issue encountered while using `persistence.type: dynamic` to persist ETCD data: 

If the `server` pod IP changes (pod killed, reschedules, node restarted,...) the `server` fail to come up again:

```
time="2025-01-03T15:29:35Z" level=info msg="Failed to test data store connection: this server is a not a member of the etcd cluster. Found [k3k-mycluster-server-0-dc22a3d8=https://10.244.3.241:2380], expect: k3k-mycluster-server-0-dc22a3d8=https://10.244.3.243:2380"
```

The shared mode use the embedded ETCD which use the local IP (Pod IP) to register the member. When the Pod IP changes , ETCD fails to start.

### Set the proper ETCD startup config using the headless service or services
```
etcd-arg:
  - --initial-cluster=... 
  - --advertise-client-urls=...
  - --initial-advertise-peer-urls=...
```
This remains challenging due to the embedded nature of ETCD. The `server` pod is not considered running until ETCD is running and ETCD won't start because the dns resolution will fail. 

### Use Sqlite in shared mode instead of ETCD
This will require some rework of the `boostrap` part which `kubelet` currently rely on to connect to the cluster.
We could use ` --write-kubeconfig` to directly store the `kubeconfig` for the `kubelet`

enrichman:
While investigating this issue I've seen one of the problem is related to the `bootstrap` secret. When the server pod is restarted the old secret are lost, and recreated.

Another issue is that deleting the `bootstrap` secret will not recreate it.

I'm trying to see if mounting an external secret where to store the certs is possible, and if it works. If yes this should solve most of the issues. We will need to think about a way to provide already existing, or custom certs though.

enrichman:
Trying to watch an owned resource (Secret) I'm facing an infinite reconciliation loop.

After a bit of investigation I've seen that this is because of the Webhook certs being updated at every loop. At the moment the reconciliation is pretty convoluted, and adding the fix is not trivial, or possible, without a small refactor.

I'll open a small PR with some of them. Nice related article: https://ahmet.im/blog/controller-pitfalls/

PR #208, opened 2025-01-23
Logs and organization cleanup

enrichman:
Trying to following some best practices from this article I tried to cleanup a bit the Reconciliation logic, trying to apply the folliwing pattern:

```go
func (m *FooController) Reconcile(..., req ctrl.Request) (ctrl.Result, error) {
    log := ctrl.LoggerFrom(ctx)
    obj := new(apiv1.Foo)
    // 1. Fetch the resource from cache: r.Client.Get(ctx, req.NamespacedName, obj)
    // 2. Finalize object if it's being deleted
    // 3. Add finalizers + r.Client.Update() if missing

    orig := foo.DeepCopy() // take a copy of the object since we'll update it below
    foo.InitializeConditions() // set all conditions to "Unknown", if missing

   // 4. Reconcile the resource and its child resources (the magic happens here).
   //    Calculate the conditions and other "status" fields along the way.
   reconcileErr := r.reconcileKind(ctx, kcp) // magic happens here!

    // 5. if (orig.Status != foo.Status): client.Status.Patch(obj)
}
```

From the snippet I've seen we can use the logger set in the controller-runtime (`ctrl.LoggerFrom(ctx)`), and I moved the finalize/cleanup in a different file (`cluster_finalize.go`).

All the logic was kept as before.

This was done to try to simplify a bit the flow before adding some fixes related to the #170.

### DIFF
--- a/main.go
+++ b/main.go
@@ -116,18 +116,18 @@ func run(clx *cli.Context) error {
 	}
 
 	logger.Info("adding etcd pod controller")
-	if err := cluster.AddPodController(ctx, mgr, logger); err != nil {
+	if err := cluster.AddPodController(ctx, mgr); err != nil {
 		return fmt.Errorf("failed to add the new cluster controller: %v", err)
 	}
 
 	logger.Info("adding clusterset controller")
-	if err := clusterset.Add(ctx, mgr, clusterCIDR, logger); err != nil {
+	if err := clusterset.Add(ctx, mgr, clusterCIDR); err != nil {
 		return fmt.Errorf("failed to add the clusterset controller: %v", err)
 	}
 
 	if clusterCIDR == "" {
 		logger.Info("adding networkpolicy node controller")
-		if err := clusterset.AddNodeController(ctx, mgr, logger); err != nil {
+		if err := clusterset.AddNodeController(ctx, mgr); err != nil {
 			return fmt.Errorf("failed to add the clusterset node controller: %v", err)
 		}
 	}
--- a/main.go
+++ b/main.go
@@ -109,8 +109,9 @@ func run(clx *cli.Context) error {
 	}
 
 	ctrlruntimelog.SetLogger(zapr.NewLogger(logger.Desugar().WithOptions(zap.AddCallerSkip(1))))
+
 	logger.Info("adding cluster controller")
-	if err := cluster.Add(ctx, mgr, sharedAgentImage, sharedAgentImagePullPolicy, logger); err != nil {
+	if err := cluster.Add(ctx, mgr, sharedAgentImage, sharedAgentImagePullPolicy); err != nil {
 		return fmt.Errorf("failed to add the new cluster controller: %v", err)
 	}
 
--- a/pkg/controller/cluster/cluster.go
+++ b/pkg/controller/cluster/cluster.go
@@ -13,8 +13,6 @@ import (
 	"github.com/rancher/k3k/pkg/controller/cluster/agent"
 	"github.com/rancher/k3k/pkg/controller/cluster/server"
 	"github.com/rancher/k3k/pkg/controller/cluster/server/bootstrap"
-	"github.com/rancher/k3k/pkg/log"
-	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -51,12 +49,10 @@ type ClusterReconciler struct {
 	Scheme                     *runtime.Scheme
 	SharedAgentImage           string
 	SharedAgentImagePullPolicy string
-	logger                     *log.Logger
 }
 
 // Add adds a new controller to the manager
-func Add(ctx context.Context, mgr manager.Manager, sharedAgentImage, sharedAgentImagePullPolicy string, logger *log.Logger) error {
-
+func Add(ctx context.Context, mgr manager.Manager, sharedAgentImage, sharedAgentImagePullPolicy string) error {
 	discoveryClient, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
 	if err != nil {
 		return err
@@ -69,7 +65,6 @@ func Add(ctx context.Context, mgr manager.Manager, sharedAgentImage, sharedAgent
 		Scheme:                     mgr.GetScheme(),
 		SharedAgentImage:           sharedAgentImage,
 		SharedAgentImagePullPolicy: sharedAgentImagePullPolicy,
-		logger:                     logger.Named(clusterController),
 	}
 
 	return ctrl.NewControllerManagedBy(mgr).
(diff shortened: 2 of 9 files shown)


pkg/controller/cluster/cluster.go
```diff
@@ -81,78 +76,62 @@ func Add(ctx context.Context, mgr manager.Manager, sharedAgentImage, sharedAgent
...
-	}
+	return reconcile.Result{}, reconcilerErr
+}
 
-	if err := c.unbindNodeProxyClusterRole(ctx, &cluster); err != nil {
-		return reconcile.Result{}, err
-	}
+func (c *ClusterReconciler) reconcileCluster(ctx context.Context, cluster *v1alpha1.Cluster) error {
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("reconcileCluster")
```
galal-hussein, line 115: can we add more informative massage instead of adding the function name, maybe reconciling cluster and adding a field with the cluster name/namespace

pkg/controller/cluster/cluster.go
```diff
@@ -81,78 +76,62 @@ func Add(ctx context.Context, mgr manager.Manager, sharedAgentImage, sharedAgent
 }
 
 func (c *ClusterReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
-	var (
-		cluster v1alpha1.Cluster
-		podList v1.PodList
-	)
-	log := c.logger.With("Cluster", req.NamespacedName)
+	log := ctrl.LoggerFrom(ctx)
```
galal-hussein, line 79: so the reason I think I pass the logger is that this logger is initialized in the main, to this controller, since we have too many controllers in the same process I initialize the logger as follows 
```
logger: logger.Named(clusterController)
```
This will basically initialize the logger with the name of the controller as well as adding fields related to that controller

pkg/controller/cluster/cluster.go
enrichman: I haven't add any more information because when you get the logger from the controller-runtime it will be already filled with many information, so you just need to take care about the message:

```
{"level":"info","timestamp":"2025-01-27T23:04:59.854Z","msg":"reconcileCluster","controller":"cluster","controllerGroup":"k3k.io","controllerKind":"Cluster","Cluster":{"name":"cluster1","namespace":"default"},"namespace":"default","name":"cluster1","reconcileID":"eb32a4f6-ca66-49e1-8bf5-2946f83f5e49"}
```
```json
{
    "level": "info",
    "timestamp": "2025-01-27T23:04:59.854Z",
    "msg": "reconcileCluster",
    "controller": "cluster",
    "controllerGroup": "k3k.io",
    "controllerKind": "Cluster",
    "Cluster": {
        "name": "cluster1",
        "namespace": "default"
    },
    "namespace": "default",
    "name": "cluster1",
    "reconcileID": "eb32a4f6-ca66-49e1-8bf5-2946f83f5e49"
}
```

Also the next log is the one related to the underlying function called (`reconcileCluster`):

```json
{
    "level": "info",
    "timestamp": "2025-01-27T23:04:59.854Z",
    "msg": "reconcileCluster",
    "controller": "cluster",
    "controllerGroup": "k3k.io",
    "controllerKind": "Cluster",
    "Cluster": {
        "name": "cluster1",
        "namespace": "default"
    },
    "namespace": "default",
    "name": "cluster1",
    "reconcileID": "eb32a4f6-ca66-49e1-8bf5-2946f83f5e49"
}
```

If the object was going to be finalized then we will see the other `log.Info("finalizeCluster")` log.

I agree that the message should be more readable. What about adding a new key-value to match the function caller to easily identify it? Something like:

```
log := ctrl.LoggerFrom(ctx).WithValues("caller", "finalizeCluster")
log.Info("starting cluster deletion")
```

I think I like this approach!

pkg/controller/cluster/cluster.go
enrichman: I actually forgot about the other controllers we have. Not using the controller-runtime logger will result in even less info though, and a very long logger name:

```
{
    "level": "info",
    "timestamp": "2025-01-27T23:04:59.792Z",
    "logger": "k3k-pod-controller.k3k-clusterset-controller.k3k-node-controller",
    "msg": "Handle etcd server pod",
    "Pod": "default/k3k-cluster1-server"
}
```

Changing the current implementation using the one from the controller runtime instead:

```go
log := ctrl.LoggerFrom(ctx)
// log := p.logger.With("Pod", req.NamespacedName)
```

will result in this log:

```json
{
    "level": "info",
    "timestamp": "2025-01-27T23:55:17.564Z",
    "msg": "Handle etcd server pod",
    "controller": "k3k-pod-controller",
    "namespace": "default",
    "name": "k3k-cluster1-server",
    "reconcileID": "06cc4bdf-4e7a-4f92-9606-01ccdf0384a9"
}
```

As you can see we have the controller already injected (`"controller": "k3k-pod-controller"`), and a very nice thing is the `reconcileID`. With that you can follow all the events that are part of the same reconciliation request.

enrichman:
@galal-hussein thanks for the review! As a wrap up using the logger injected in the context will provide already all the core information about the controller, namespace and name:

```
{
    "level": "info",
    "timestamp": "2025-01-27T23:55:17.564Z",
    "msg": "Handle etcd server pod",
    "controller": "k3k-pod-controller",
    "namespace": "default",
    "name": "k3k-cluster1-server",
    "reconcileID": "06cc4bdf-4e7a-4f92-9606-01ccdf0384a9"
}
```

I think I will need to find out where the `"Cluster": {"name": "cluster1","namespace": "default"}` field is populated, because I'm not really sure. Also I will change the messages with something more human readable, and add the `"caller":"<func>"` at the beginning.

What do you think about this?

galal-hussein:
> @galal-hussein thanks for the review! As a wrap up using the logger injected in the context will provide already all the core information about the controller, namespace and name:
> 
> ```
> {
>     "level": "info",
>     "timestamp": "2025-01-27T23:55:17.564Z",
>     "msg": "Handle etcd server pod",
>     "controller": "k3k-pod-controller",
>     "namespace": "default",
>     "name": "k3k-cluster1-server",
>     "reconcileID": "06cc4bdf-4e7a-4f92-9606-01ccdf0384a9"
> }
> ```
> 
> I think I will need to find out where the `"Cluster": {"name": "cluster1","namespace": "default"}` field is populated, because I'm not really sure. Also I will change the messages with something more human readable, and add the `"caller":"<func>"` at the beginning.
> 
> What do you think about this?

okay as long as we have a structured logging with informative log messages then I am okay with it, one of the reason tho that I went with the first approach is to plan to add more structure loggers to the interface, basically will have the option for example to use zap or logrus, or logr.

As for the cluster name/namespace I just missed to add it, but its being added to other controllers for example https://github.com/rancher/k3k/blob/main/pkg/controller/clusterset/clusterset.go#L124

pkg/controller/cluster/cluster.go
galal-hussein: sorry if I am not very knowledgeable about this, but if I understand correctly the LoggerFrom is using logr? will that result into two sets of loggers in k3k controllers?

main.go
```diff
@@ -109,8 +109,9 @@ func run(clx *cli.Context) error {
 	}
 
 	ctrlruntimelog.SetLogger(zapr.NewLogger(logger.Desugar().WithOptions(zap.AddCallerSkip(1))))
+
 	logger.Info("adding cluster controller")
-	if err := cluster.Add(ctx, mgr, sharedAgentImage, sharedAgentImagePullPolicy, logger); err != nil {
+	if err := cluster.Add(ctx, mgr, sharedAgentImage, sharedAgentImagePullPolicy); err != nil {
```
galal-hussein, line 114: we would also need to remove it from other controllers

pkg/controller/cluster/cluster.go
galal-hussein: yeah that works!

pkg/controller/cluster/cluster.go
enrichman: Basically `LoggerFrom` will fetch the logger set in the controller-runtime.

In `main.go` we were setting it like this:
```go
ctrlruntimelog.SetLogger(zapr.NewLogger(logger.Desugar().WithOptions(zap.AddCallerSkip(1))))
```
so we will get a Zap logger. `logr` is just an API (or "interface"?), and it's working with whichever logger implementation you set. As you can see we are using `zapr.NewLogger` to use zap within the logr interface, but still the underlying logger will be the same we created before (zap).

The only difference is that now we can get it directly from the controllers, already filled with some information about the reconciliation loop.

enrichman:
@galal-hussein I've implemented the requested changes, PTAL!

One thing: I have not added the "caller" value because of 2 reasons. First one is that adding everytime the caller function will cause the log to have multiple values, if we reinject the logger in the context. But the main reason is that this value is already provided by default in the Config, it just needs to be enabled.

Should I enable it?

pkg/controller/cluster/cluster.go
galal-hussein: Thats cool, thank you!

galal-hussein (approved):

pkg/controller/clusterset/node.go
```diff
@@ -56,26 +56,33 @@ func (n *NodeReconciler) Reconcile(ctx context.Context, req reconcile.Request) (
...
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("ensuring network policies")
+
 	var setNetworkPolicy *networkingv1.NetworkPolicy
 	for _, cs := range clusterSetList.Items {
 		if cs.Spec.DisableNetworkPolicy {
 			continue
 		}
+
+		clusterSetLog := log.WithValues("clusterset", cs.Namespace+"/"+cs.Name)
```
galal-hussein, line 76: just a nit here, we can just override the log variable here, instead of creating a new one:

log = log.WithValues(....

enrichman: Yep! I thought about this and I wasn't sure. Happy to change it :)

PR #226, opened 2025-02-03
Fix Webhook certificate recreate

enrichman:
Ref:
- #170 
---

## Description

The webhook certificate was created more than once, this was causing some issues related to it not being aligned between the different resources. This was because the reconciliation of the object was done in a generic function of the Controller (the `(c *ClusterReconciler) ensure(...)`), that was not aware of the specific resource. And since the webhook certificate is created with a time.Now() validity then it was always considered different from the previous one.

To solve this I moved the resource reconciliation responsibility to the agents. In this way it's possible to "customize" the reconciliation, and in the case of the Webhook certs we are checking for their exisistance, and eventually create them, without updating them. In the future we could check for the validity, but this was out of scope, at the moment.

## Changes

Now the `ClusterReconciler` will create the Virtual or Shared agent with a `Config`, that will be used to perform the CreateOrUpdate of the resources. The Agents implement the `ResourceEnsurer` interface.

```go
type ResourceEnsurer interface {
	EnsureResources(context.Context) error
}

func (c *ClusterReconciler) ensureAgent(ctx context.Context, cluster *v1alpha1.Cluster, serviceIP, token string) error {
	config := agent.NewConfig(cluster, c.Client, c.Scheme)

	var agentEnsurer ResourceEnsurer
	if cluster.Spec.Mode == agent.VirtualNodeMode {
		agentEnsurer = agent.NewVirtualAgent(config, serviceIP, token)
	} else {
		agentEnsurer = agent.NewSharedAgent(config, serviceIP, c.SharedAgentImage, c.SharedAgentImagePullPolicy, token)
	}

	return agentEnsurer.EnsureResources(ctx)
}
```

The `EnsureResources(ctx)` of each agent will call every resource func, that will know how to "ensure" every resource. Generally they will call the `ensureObject(ctx context.Context, obj ctrlruntimeclient.Object)`

```
func (v *VirtualAgent) ensureObject(ctx context.Context, obj ctrlruntimeclient.Object) error {
	return ensureObject(ctx, v.Config, obj)
}
```

but when needed it's now possible to do a different update logic.

### DIFF
--- a/pkg/controller/cluster/agent/agent.go
+++ b/pkg/controller/cluster/agent/agent.go
@@ -1,28 +1,44 @@
 package agent
 
 import (
+	"context"
+	"fmt"
+
 	"github.com/rancher/k3k/pkg/apis/k3k.io/v1alpha1"
 	"github.com/rancher/k3k/pkg/controller"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 const (
 	configName = "agent-config"
 )
 
-type Agent interface {
-	Name() string
-	Config() ctrlruntimeclient.Object
-	Resources() ([]ctrlruntimeclient.Object, error)
-}
-
-func New(cluster *v1alpha1.Cluster, serviceIP, sharedAgentImage, sharedAgentImagePullPolicy, token string) Agent {
-	if cluster.Spec.Mode == VirtualNodeMode {
-		return NewVirtualAgent(cluster, serviceIP, token)
-	}
-	return NewSharedAgent(cluster, serviceIP, sharedAgentImage, sharedAgentImagePullPolicy, token)
-}
-
 func configSecretName(clusterName string) string {
 	return controller.SafeConcatNameWithPrefix(clusterName, configName)
+}
+
+func (s *SharedAgent) ensureObject(ctx context.Context, obj ctrlruntimeclient.Object) error {
+	return ensureObject(ctx, s.cluster, obj, s.client, s.scheme)
+}
+
+func (v *VirtualAgent) ensureObject(ctx context.Context, obj ctrlruntimeclient.Object) error {
+	return ensureObject(ctx, v.cluster, obj, v.client, v.scheme)
+}
+
+func ensureObject(ctx context.Context, cluster *v1alpha1.Cluster, obj ctrlruntimeclient.Object, ctrlClient ctrlruntimeclient.Client, scheme *runtime.Scheme) error {
+	result, err := controllerutil.CreateOrUpdate(ctx, ctrlClient, obj, func() error {
+		fmt.Printf("call FN mut %v - %s\n", cluster, client.ObjectKeyFromObject(obj))
+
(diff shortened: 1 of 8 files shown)


pkg/controller/cluster/agent/agent.go
```diff
@@ -1,28 +1,51 @@
...
+func ensureObject(ctx context.Context, cfg *Config, obj ctrlruntimeclient.Object) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	result, err := controllerutil.CreateOrUpdate(ctx, cfg.client, obj, func() error {
+		return controllerutil.SetControllerReference(cfg.cluster, obj, cfg.scheme)
+	})
+
+	if result != controllerutil.OperationResultNone {
+		key := client.ObjectKeyFromObject(obj)
+		log.Info(fmt.Sprintf("ensureObject: %T object %s was %s", obj, key, result))
```
galal-hussein, line 47: nit: I think we can either initialize the log with values, or use key/values in the Info
```suggestion
		log.Info("ensure Object",  "Object", obj, "Key", key, "result", result)
```

Also I wouldn't print the object maybe just the name and namespace

pkg/controller/cluster/cluster.go
```diff
@@ -368,16 +373,21 @@ func (c *ClusterReconciler) bindNodeProxyClusterRole(ctx context.Context, cluste
 	return c.Client.Update(ctx, clusterRoleBinding)
 }
 
-func (c *ClusterReconciler) agent(ctx context.Context, cluster *v1alpha1.Cluster, serviceIP, token string) error {
-	agent := agent.New(cluster, serviceIP, c.SharedAgentImage, c.SharedAgentImagePullPolicy, token)
-	agentsConfig := agent.Config()
-	agentResources, err := agent.Resources()
-	if err != nil {
-		return err
+type ResourceEnsurer interface {
```
galal-hussein, line 376: can we add this method to the agent interface thats already exists in agent.go

pkg/controller/cluster/token.go
```diff
@@ -63,14 +64,17 @@ func (c *ClusterReconciler) ensureTokenSecret(ctx context.Context, cluster *v1al
...
+	key = client.ObjectKeyFromObject(&tokenSecret)
 
-	if err := c.ensure(ctx, &tokenSecret, false); err != nil {
-		return "", err
+	result, err := controllerutil.CreateOrUpdate(ctx, c.Client, &tokenSecret, func() error {
+		return controllerutil.SetControllerReference(cluster, &tokenSecret, c.Scheme)
+	})
+
+	if result != controllerutil.OperationResultNone {
+		log.Info(fmt.Sprintf("ensureObject: tokenSecret %s was %s", key, result))
```
galal-hussein, line 74: same as https://github.com/rancher/k3k/pull/226/files#r1941833363

pkg/controller/cluster/agent/agent.go
enrichman: Yes, I can do that, but still we need the object type. I added it because we have multiple resources with the same name (i.e. services, roles, and others with `k3k-mycluster-kubelet`).

pkg/controller/cluster/cluster.go
enrichman: I removed that interface, but I can move this in the `agent` package if you prefer. 👍

galal-hussein: yeah we can move it to the agent, the reason I did that interface so that we can add to it later if needed, but I think its fine for now

galal-hussein (approved):

Issue #206, opened 2025-01-22
`k3kcli` UX enhancement


enrichman:
Now we have improved a bit the Cluster spec, and we should think about aligning the `k3kcli` to the new changes.

Some proposed changes:

### `k3kcli cluster create`

Currently the Cluster name is a flag (`--name value`). This is now the only required field, together with the `namespace`. It would make sense to move it as the only required argument instead:

```
k3kcli cluster create mycluster
```

Regarding the `--namespace`, if not specified it looks cleaner to have a dedicated namespace. I.e. the above `k3kcli cluster create mycluster` command will create the Cluster inside a new `k3k-mycluster` namespace. If the `--namespace` flag is defined we will try to create the resource in that namespace, without checking its existence. We can think about adding a `--create-namespace` flag eventually (like Helm).

The `--version` flag help should be updated to indicate that the k8s host version will be used by default.

As per other issue the default kubeconfig should be taken into account, it the KUBECONFIG var is empty.

Also we should document how to embed the automplete (I wasn't able to).

Related issues:
- https://github.com/rancher/k3k/issues/153
- https://github.com/rancher/k3k/issues/191

VestigeJ:
+1 for mirroring how existing kubectl functions for naming of a resource ie 
```
k3k cluster create my_cluster_name
```

+1 for mirroring the use of a default namespace as its also an existing expectation from anyone familiar with kubectl commands against a cluster
```
k3k cluster create my_cluster_name
-> deploying my_cluster_name to k3k-default...
```

As it's an expectation for k3k to run under an existing cluster it would make the most sense to create a directory adjacent to the existing config 
```
sudo ls /etc/rancher/k3s/k3k.d/my_cluster_name.yaml

sudo ls /etc/rancher/rke2/k3k.d/my_cluster_name.yaml

ls ~/.kube/k3k.d/my_cluster_name.yaml
```

If KUBECONFIG isn't set the command should fail the same way it does against a regular cluster. 

```
k3k cluster list == kubectl get clusters (assuming no clusters are created with cluster-api here maybe check their type)
```

$ kubectl config --help
```
Modify kubeconfig files using subcommands like "kubectl config set
current-context my-context".

 The loading order follows these rules:

  1.  If the --kubeconfig flag is set, then only that file is loaded. The flag
may only be set once and no merging takes place.
  2.  If $KUBECONFIG environment variable is set, then it is used as a list of
paths (normal path delimiting rules for your system). These paths are merged.
[...]

Issue #153, opened 2024-11-29
Set `--kubeconfig` flag as optional in `k3kcli cluster <cmd>` commands


enrichman:
The `k3kcli cluster` commands need the `--kubeconfig` flag.

This should not be mandatory, and it should follow the same logic of `kubectl` ([see doc](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_config/#synopsis)):

> - If the --kubeconfig flag is set, then only that file is loaded. The flag may only be set once and no merging takes place.
>
> - If $KUBECONFIG environment variable is set, then it is used as a list of paths (normal path delimiting rules for your system). These paths are merged. When a value is modified, it is modified in the file that defines the stanza. When a value is created, it is created in the first file that exists. If no files in the chain exist, then it creates the last file in the list.
> 
> - Otherwise, ${HOME}/.kube/config is used and no merging takes place.

It should work by default, but if not I think we can avoid the merging of multiple files, if complicated, and just add the fallback to the default location `${HOME}/.kube/config`.

VestigeJ:
@enrichman this is looking good! 

$ unset KUBECONFIG
$ echo $KUBECONFIG
$  ls .kube
```
cache  config
```

$  k3kcli cluster create --namespace default --servers 1 --agents 1 --mode virtual --persistence-type ephemeral defaultkubefolder
```
INFO[0000] Creating a new cluster [defaultkubefolder]
INFO[0000] Extracting Kubeconfig for [defaultkubefolder] cluster
INFO[0000] waiting for cluster to be available..
INFO[0065] certificate CN=system:admin,O=system:masters signed by CN=k3s-client-ca@1741045771: notBefore=2025-03-03 23:49:31 +0000 UTC notAfter=2026-03-03 23:50:34 +0000 UTC
INFO[0065] You can start using the cluster with:

		export KUBECONFIG=/home/ec2-user/defaultkubefolder-kubeconfig.yaml
		kubectl cluster-info
```



$ export KUBECONFIG=/home/ec2-user/rke2.yaml
$ k3kcli cluster create --namespace default --servers 1 --mode virtual --persistence-type ephemeral doozy
```
INFO[0000] Creating a new cluster [doozy]
INFO[0000] Extracting Kubeconfig for [doozy] cluster
INFO[0000] waiting for cluster to be available..
[...]

PR #230, opened 2025-02-04
Small cli refactor (`cluster name` as arg, default kubeconfig path)

enrichman:
Ref:
- #206 
- #153
---

As discussed in the other issue this PR changes the behaviour of the CLI to accept an argument for the cluster name, instead of the `--name` flag:

```
k3kcli cluster create my_cluster_name
```

It also default the `--kubeconfig` value to the `$HOME/.kube/config` if not set as a flag nor environment variable (the same as `kubectl`).

As a small refactor I've moved the flags in a separate file (`cli/cmds/cluster/create_flags.go`), using also the new `Action` func to validate the flags (from `urfave/cli/v2`), removing the old validation func.

There are also other changes that we need to do, but I'll leave them for other PRs.


cli/cmds/cluster/create_cmd.go
```diff
@@ -0,0 +1,202 @@
+package cluster
```
galal-hussein, line 1: is there a reason to rewrite the file instead of changes to the create.go?

galal-hussein: if the reason because of the flags file, we can just leave the file as it is "create.go" and the flags file can be create_flags.go or flags.go

enrichman: Yeah I did a back-and-forth actually. I wasn't sure, and I left it like this because this way the file `create_cmd.go` was ordered before the `create_flags.go` lol. But I agree that is probably better to have it just as `create.go`. 😅

galal-hussein (approved):

Issue #202, opened 2025-01-22
Docs


jp-gouin:
We should review the current documentation and add more to support the early adopters

- [ ] Refresh/cleanup the readme
- [ ] Quickstart in the readme 
- [ ] docs/Example to setup dev environment
- [ ] docs/HowTos

jp-gouin:
Closed by #231

PR #233, opened 2025-02-06
Fixing etcd pod controller

galal-hussein:
etcd pod controller does the following
- reconcile k3s in HA state, i.e whenever a pod is deleted and assigned a new IP, then the controller will remove the etcd member from the cluster allowing the new pod to be added to the cluster

The fix added the namespace lookup to the clutser, it was missing and that caused the controller not able to reconcile the cluster
#170

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -23,7 +23,7 @@ RUN go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest && \
     cp $ENVTEST_BIN/* /usr/local/kubebuilder/bin
 
 ENV GO111MODULE on
-ENV DAPPER_ENV REPO TAG DRONE_TAG CROSS GITHUB_TOKEN SKIP_TESTS
+ENV DAPPER_ENV REPO TAG DRONE_TAG CROSS GITHUB_TOKEN SKIP_TESTS GIT_TAG
 ENV DAPPER_SOURCE /go/src/github.com/rancher/k3k/
 ENV DAPPER_OUTPUT ./bin ./dist ./deploy ./charts
 ENV DAPPER_DOCKER_SOCKET true
--- a/pkg/controller/cluster/pod.go
+++ b/pkg/controller/cluster/pod.go
@@ -59,7 +59,7 @@ func AddPodController(ctx context.Context, mgr manager.Manager) error {
 }
 
 func (p *PodReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
-	log := ctrl.LoggerFrom(ctx).WithValues("pod", req.NamespacedName)
+	log := ctrl.LoggerFrom(ctx).WithValues("statefulset", req.NamespacedName)
 	ctx = ctrl.LoggerInto(ctx, log) // enrich the current logger
 
 	s := strings.Split(req.Name, "-")
@@ -71,7 +71,7 @@ func (p *PodReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	}
 	clusterName := s[1]
 	var cluster v1alpha1.Cluster
-	if err := p.Client.Get(ctx, types.NamespacedName{Name: clusterName}, &cluster); err != nil {
+	if err := p.Client.Get(ctx, types.NamespacedName{Name: clusterName, Namespace: req.Namespace}, &cluster); err != nil {
 		if !apierrors.IsNotFound(err) {
 			return reconcile.Result{}, err
 		}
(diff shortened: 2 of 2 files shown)


VestigeJ (approved):
LGTM

Dockerfile.dapper
```diff
@@ -23,7 +23,7 @@ RUN go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest && \
     cp $ENVTEST_BIN/* /usr/local/kubebuilder/bin
 
 ENV GO111MODULE on
-ENV DAPPER_ENV REPO TAG DRONE_TAG CROSS GITHUB_TOKEN SKIP_TESTS
+ENV DAPPER_ENV REPO TAG DRONE_TAG CROSS GITHUB_TOKEN SKIP_TESTS GIT_TAG
```
enrichman, line 26: Is this something coming from some custom script? Not a big deal, we will drop Dapper one day.

pkg/controller/cluster/pod.go
```diff
@@ -59,7 +59,7 @@ func AddPodController(ctx context.Context, mgr manager.Manager) error {
 }
 
 func (p *PodReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
-	log := ctrl.LoggerFrom(ctx).WithValues("pod", req.NamespacedName)
+	log := ctrl.LoggerFrom(ctx).WithValues("statefulset", req.NamespacedName)
```
enrichman, line 62: Ah, so this is actually receiving StatefulSets because of the `EnqueueRequestForOwner`. Gotcha. :+1:

pkg/controller/cluster/pod.go
```diff
@@ -84,8 +84,10 @@ func (p *PodReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	if err := p.Client.List(ctx, &podList, listOpts); err != nil {
 		return reconcile.Result{}, ctrlruntimeclient.IgnoreNotFound(err)
 	}
+	if len(podList.Items) == 1 {
+		return reconcile.Result{}, nil
+	}
```
enrichman, line 89: Why we are now skipping the reconciliation of the single pod?

Dockerfile.dapper
galal-hussein: yeah, this is just to facilitate the dev testing, this is how I test my changes:

build_push.sh
```
#!/bin/bash
SKIP_TESTS=true GIT_TAG=dev REPO=husseingalal make
docker push husseingalal/k3k:dev
docker push husseingalal/k3k:k3k-kubelet
sleep 10
helm delete k3k -n k3k-system
helm install --namespace k3k-system --set image.repository=husseingalal/k3k --set image.tag=dev --set sharedAgent.image.repository=husseingalal/k3k --set sharedAgent.image.tag=k3k-kubelet k3k k3k/k3k --devel --set-string debug=true --set image.pullPolicy=Always --set sharedAgent.image.pullPolicy=Always
```

pkg/controller/cluster/pod.go
galal-hussein: yup

pkg/controller/cluster/pod.go
galal-hussein: single pod doesnt need reconciliation for etcd, because its a brand new cluster, this will probably will need some changes tho in dynamic persistence mode, but that will be another pr

Dockerfile.dapper
enrichman: Yes, I too have a "custom script" for my development flow.

Since I will try to create the Development documentation for the #202 (and #231) it's probably better to have something we agree, altough the real "local" development could still diverge a bit. :smile:

pkg/controller/cluster/pod.go
enrichman: Hm, but we don't have any restriction about the scaling of the server pods. So you could eventually start from 1, then scale up to 3, and then scale down to 1 again. Maybe this particular case will work, but what if I have 3 pods, kill the `server-0` pod, that will be rescheduled with a new ip (I suppose), and then scale down to 1 pod?

I'm not sure, just thinking about some possible scenario.

galal-hussein: thats why actually I am listing the pods, basically everytime any scale happen to the statefulset, this controller will be triggered, listing the pod, and if the pod is marked to be deleted from the statefulset then it will be removed first from the etcd cluster member list, its a standard operation to allow a stable etcd cluster

Dockerfile.dapper
galal-hussein: we will eventually just remove the dapper altogether, this is just temporary

pkg/controller/cluster/pod.go
galal-hussein: however if there was only one pod, and this is marked for deletion, then the controller will do nothing because its the only member in etcd member list

enrichman: Ok, I just wanted to be sure about the allowed number of replicas. I wasn't sure if we are actually allowing a change on that part, but it makes sense in my opinion.

Let's merge it, thanks!

enrichman (approved):

Issue #246, opened 2025-02-11
Changing the `expose` configuration after cluster creation has no effect


enrichman:
After initial Cluster creation, changes to the expose setting are not applied to the running Cluster.

ShylajaDevadiga:
Can we have the steps to validate this issue? @enrichman

enrichman:
@ShylajaDevadiga sure. This was a bug that was affecting the Expose configuration. When you create a Virtual Cluster with the NodePort expose (for example), then it was not possible to change it afterwards. Now the update of the ExposeConfig is actually respected. To check all the possible configuration you can refer to the doc: https://github.com/enrichman/k3k/blob/main/docs/crds/crd-docs.md#exposeconfig

To test this you can try to create a Virtual Cluster with an empty expose config, or with one of your preferences, and then update it. You should see that for example a Service with the NodePort type is created when specyfing the NodePort expose, or when a LoadBalancer is specified, then the NodePort is then deleted, and the LoadBalancer created, and so on.

PR #248, opened 2025-02-12
Fix and enhancements to IngressExposeConfig (`annotations`)

enrichman:
Ref:
- #246
---

This PR fixes some issues with the current Ingress implementation adding some additional configurations field.

Like the NodePort expose the `enabled` field was removed (#247). The `annotations` field was added.

The `tlsSANs` field will be used as Path to add the relevant IngressRules, while the `annotations` will be applied to the Ingress. This is because those annotations are needed to enable some behaviors to the proxy, i.e. when using an Nginx ingress-controller the SSL passthrough option is needed, and the following annotations need to be used:

```yaml
spec:
  tlsSANs:
    - my-cluster.140.150.160.199.sslip.io
  expose:
    ingress:
      ingressClassName: "nginx"
      annotations:
        nginx.ingress.kubernetes.io/ssl-passthrough: "true"
        nginx.ingress.kubernetes.io/backend-protocol: "true"
        nginx.ingress.kubernetes.io/ssl-redirect: "HTTPS"
```

### DIFF
--- a/charts/k3k/crds/k3k.io_clusters.yaml
+++ b/charts/k3k/crds/k3k.io_clusters.yaml
@@ -117,10 +117,6 @@ spec:
                         description: Annotations is a key value map that will enrich
                           the Ingress annotations
                         type: object
-                      hosts:
-                        items:
-                          type: string
-                        type: array
                       ingressClassName:
                         type: string
                     type: object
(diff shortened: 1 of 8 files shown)


jp-gouin (approved):
lgtm - thanks !

Issue #245, opened 2025-02-10
Make dynamic persistent the default persistence mode


jp-gouin:
The default persistence mode should be `dynamic` to ensure resiliency by default 

**Optional** we should detect and use the  default `storageclass` is it's configured in the host cluster

PR #250, opened 2025-02-12
Default to dynamic persistence and fix HA restarts

galal-hussein:
#170 
#245 

- Default to dynamic persistence instead of ephemeral
- Use the default storage cluster on the host cluster if none is specified
- Fix the resiliency issue, by adding cluster-reset to the single server mode when it restarts
- Fix resiliency issue in HA, by pointing the single server to the load balancer service instead of creating a new cluster


tests/tests_suite_test.go
```diff
@@ -150,3 +166,31 @@ func buildScheme() *runtime.Scheme {
...
+
+	buf := new(bytes.Buffer)
+	_, err = io.Copy(buf, podLogs)
+	if err != nil {
+		return "", err
+	}
+	str := buf.String()
+
+	return str, nil
+}
```
enrichman, line 196: This can be simplified a bit. I.e. you don't need to export the kubeconfig, because we already have a "plain" Kubernetes client, so you can just do `k8s.CoreV1().Pods` etc. Also since we have the Expect stuff you can drop the error checking.

I would also create a dedicated func for both of the "logs", something like this (not tested!!):

```suggestion
var _ = AfterSuite(func() {
	// dump k3s logs
	readCloser, err := k3sContainer.Logs(context.Background())
	Expect(err).To(Not(HaveOccurred()))
	writeLogs("k3s.log", readCloser)
	
	// dump k3k logs
	writeK3kLogs()

	testcontainers.CleanupContainer(GinkgoTB(), k3sContainer)
})

func writeK3kLogs() {
	ctx := context.Background()
	
	var podList v1.PodList
	err = k8sClient.List(ctx, &podList, &client.ListOptions{Namespace: "k3k-system"})
	Expect(err).To(Not(HaveOccurred()))
	
	k3kPod := podList.Items[0]
	req := k8s.CoreV1().Pods(k3kPod.Namespace).GetLogs(k3kPod.Name, &corev1.PodLogOptions{})
	podLogs, err := req.Stream(ctx)
	Expect(err).To(Not(HaveOccurred()))

	writeLogs("k3k.log", podLogs)
}

func writeLogs(filename string, logs io.ReadCloser) {
	logsStr, err := io.ReadAll(logs)
	Expect(err).To(Not(HaveOccurred()))
	defer logs.Close()

	tempfile := path.Join(os.TempDir(), filename)
	err = os.WriteFile(tempfile, []byte(logsStr), 0644)
	Expect(err).To(Not(HaveOccurred()))

	fmt.Fprintln(GinkgoWriter, "logs written to: "+filename)
}
```

pkg/controller/cluster/server/template.go
```diff
@@ -0,0 +1,15 @@
...
+	/bin/k3s server --cluster-reset --config  {{.INIT_CONFIG}} {{.EXTRA_ARGS}}
+fi
+/bin/k3s server --config {{.INIT_CONFIG}} {{.EXTRA_ARGS}}`
+
+var HAServerTemplate string = ` 
+if [ ${POD_NAME: -1} == 0 ] && [ ! -d "{{.ETCD_DIR}}" ]; then
+	/bin/k3s server --config {{.INIT_CONFIG}} {{.EXTRA_ARGS}}
+else 
+	/bin/k3s server --config {{.SERVER_CONFIG}} {{.EXTRA_ARGS}}
+fi`
```
enrichman, line 15: This is getting quite complex, since K3s can be configured in many ways (env vars, config files, commands, etc) I feel there is a better way.

pkg/controller/cluster/server/server.go
```diff
@@ -352,16 +326,79 @@ func (s *Server) StatefulServer(ctx context.Context) (*apps.StatefulSet, error)
...
+	if storageClassName == "" {
+		if err := s.client.List(ctx, &storageClassList); err != nil {
+			return v1.PersistentVolumeClaim{}, err
+		}
+		for _, sc := range storageClassList.Items {
+			if util.IsDefaultAnnotation(sc.ObjectMeta) {
+				storageClassName = sc.Name
+			}
+		}
+	}
```
enrichman, line 361: While testing this I had a small issue, because an empty string is treated differently from the nil: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1

> PVCs don't necessarily have to request a class. A PVC with its storageClassName set equal to "" is always interpreted to be requesting a PV with no class, so it can only be bound to PVs with no class (no annotation or one set equal to ""). A PVC with no storageClassName is not quite the same and is treated differently by the cluster, depending on whether the [DefaultStorageClass admission plugin](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#defaultstorageclass) is turned on.

Basically we don't need to do this check, but just changin the `s.cluster.Spec.Persistence.StorageClassName` to be a pointer, and use the same. If not provided it will use the default storage class.

pkg/controller/cluster/server/template.go
galal-hussein: unfortunately this is because of raft problems with etcd, we are forced to handle different cases where the cluster is misconfigured

enrichman (approved):
LGTM, thanks!

Issue #220, opened 2025-01-29
Version String passed to cluster creation should prevent creation if malformed


VestigeJ:
Passing the below string to the version flag continues with cluster creation but fails to run because the version string isn't formatted correctly. This should return either an error or correct the flag before creation to prevent the below errors from the deployed nodes.
 
$ k3kcli cluster create --namespace k3k-cilium --name k3k-k3s --servers 1 --agents 1 --mode virtual --version v1.29.13+k3s1

```
INFO[0000] Creating a new cluster [k3k-k3s]
INFO[0000] Cluster [k3k-k3s] already exists
INFO[0000] Extracting Kubeconfig for [k3k-k3s] cluster
INFO[0000] waiting for cluster to be available..
FATA[0120] secrets "k3k-k3k-k3s-bootstrap" not found
```

$ kgp
```
NAMESPACE     NAME                                                    READY   STATUS             RESTARTS       AGE
default       k3k-7847894d9-fz96q                                     1/1     Running            0              101m
k3k-cilium    k3k-k3k-k3s-agent-5585b96978-8c2js                      0/1     InvalidImageName   0              20m
k3k-cilium    k3k-k3k-k3s-server-0                                    0/1     InvalidImageName   0              20m
```

$ kgp k3k-k3k-k3s-server-0 -n k3k-cilium -o yaml
```
state:
      waiting:
        message: 'Failed to apply default image tag "rancher/k3s:v1.29.13+k3s1": couldn''t
          parse image name "rancher/k3s:v1.29.13+k3s1": invalid reference format'
        reason: InvalidImageName

```

$ kgn
```
NAME               STATUS   ROLES                       AGE    VERSION
ip                 Ready    control-plane,etcd,master   117m   v1.32.1+rke2r1
```

$ rke2 -v
```
rke2 version v1.32.1+rke2r1 (c0f7be4407cf2c437cacfe735e5c943e827f2ff8)
go version go1.23.4 X:boringcrypto
```

$ get_figs
```
=========== rke2 config ===========
node-external-ip: 3.3.3.125
node-ip: 3.3.3.205
write-kubeconfig-mode: 644
debug: true
token: YOUR_TOKEN_HERE
cni: multus,cilium
```

enrichman:
Ouch, thanks. Yes, this should be validated. I wouldn't correct and change the value, because it could be used also directly in a Cluster spec resource, so it's probably better to validate it, and fail with an error.

[...]

PR #256, opened 2025-02-18
Fix k3s version metadata

enrichman:
Fix:
- #220 
---

The K3s release tags don't match the Docker tags:

When the user provides the release tag, i.e `v1.29.13+k3s1`, the docker pull of the K3s image will fail:

```
couldn't parse image name "rancher/k3s:v1.29.13+k3s1": invalid reference format
```

This PR adds a fallback when the user provides a version with the metadata, trying to replace the `+` with the `-`, warning the user about this behavior:

```
-> % k3kcli cluster create --version v1.29.13+k3s1 mycluster
WARN[0000] Invalid K3s docker reference version: 'v1.29.13+k3s1'. Using 'v1.29.13-k3s1' instead 
```

### DIFF
--- a/cli/cmds/cluster/create.go
+++ b/cli/cmds/cluster/create.go
@@ -89,6 +89,12 @@ func createAction(config *CreateConfig) cli.ActionFunc {
 			return err
 		}
 
+		if strings.Contains(config.version, "+") {
+			orig := config.version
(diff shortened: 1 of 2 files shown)


galal-hussein (approved):

VestigeJ:
This was tested and that was closed - https://github.com/rancher/k3k/issues/220

PR #263, opened 2025-02-21
Added ServiceCIDR lookup, and changed default

enrichman:
In shared mode, the service CIDR must match the host's service CIDR.

This PR adds a lookup mechanism to retrieve the correct service CIDR instead of relying on the default.

It first attempts to locate the kube-apiserver pod and extract the `--service-cluster-ip-range` flag from its arguments.

If the flag is not found, it attempts to create a failing service with an invalid cluster IP. This will trigger an error message containing the valid service CIDR.

If both methods fail, the default service CIDR is used as a fallback. This PR also updates the default service CIDR to match the K3s/RKE2 default (`10.43.0.0/16`).

### DIFF
--- a/pkg/controller/cluster/agent/agent.go
+++ b/pkg/controller/cluster/agent/agent.go
@@ -48,7 +48,7 @@ func ensureObject(ctx context.Context, cfg *Config, obj ctrlruntimeclient.Object
 
 	if result != controllerutil.OperationResultNone {
 		key := client.ObjectKeyFromObject(obj)
-		log.Info(fmt.Sprintf("ensuring %T", obj), "key", key, "result, result")
+		log.Info(fmt.Sprintf("ensuring %T", obj), "key", key, "result", result)
 	}
 
 	return err
--- a/pkg/controller/cluster/cluster.go
+++ b/pkg/controller/cluster/cluster.go
@@ -427,9 +427,10 @@ func (c *ClusterReconciler) validate(cluster *v1alpha1.Cluster) error {
 	return nil
 }
 
-// lookupServiceCIDR will try to lookup the serviceCIDR of the cluster.
-// It will first look for the 'kube-apiserver' pod, looking for the --service-cluster-ip-range flag.
-// Otherwise
+// lookupServiceCIDR attempts to determine the cluster's service CIDR.
+// It first searches the 'kube-apiserver' pod's arguments for the --service-cluster-ip-range flag.
+// If the flag is not found, it attempts to create a failing Service (with an invalid cluster IP)
+// and extracts the expected CIDR from the resulting error.
 func (c *ClusterReconciler) lookupServiceCIDR(ctx context.Context) (string, error) {
 	log := ctrl.LoggerFrom(ctx)
 
--- a/pkg/controller/cluster/cluster.go
+++ b/pkg/controller/cluster/cluster.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"reflect"
 	"strings"
 	"time"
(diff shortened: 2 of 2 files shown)


pkg/controller/cluster/agent/agent.go
```diff
@@ -48,7 +48,7 @@ func ensureObject(ctx context.Context, cfg *Config, obj ctrlruntimeclient.Object
 
 	if result != controllerutil.OperationResultNone {
 		key := client.ObjectKeyFromObject(obj)
-		log.Info(fmt.Sprintf("ensuring %T", obj), "key", key, "result, result")
+		log.Info(fmt.Sprintf("ensuring %T", obj), "key", key, "result", result)
```
galal-hussein, line 51: can we use `WithValues` instead of sprintf just to be consistent

pkg/controller/cluster/cluster.go
```diff
@@ -413,3 +426,76 @@ func (c *ClusterReconciler) validate(cluster *v1alpha1.Cluster) error {
...
+			// validate serviceCIDR
+			_, serviceCIDRAddr, err := net.ParseCIDR(serviceCIDR)
+			if err != nil {
+				return "", err
+			}
+			return serviceCIDRAddr.String(), nil
+		}
+	}
+
+	log.Info("cannot find serviceCIDR from lookup")
```
galal-hussein, line 499: nit: I think this should be a warning

pkg/controller/cluster/cluster.go
```diff
@@ -413,3 +426,76 @@ func (c *ClusterReconciler) validate(cluster *v1alpha1.Cluster) error {
...
 }
+
+// lookupServiceCIDR will try to lookup the serviceCIDR of the cluster.
+// It will first look for the 'kube-apiserver' pod, looking for the --service-cluster-ip-range flag.
+// Otherwise
+func (c *ClusterReconciler) lookupServiceCIDR(ctx context.Context) (string, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	// find the kube-apiserver pod, and look for the '--service-cluster-ip-range' flag
+	matchingLabels := ctrlruntimeclient.MatchingLabels(map[string]string{
```
galal-hussein, line 437: we may as well just use the fallback method, since that not all distros (host clusters) run the api server as a static pod

galal-hussein (commented):
just some few comments

pkg/controller/cluster/agent/agent.go
enrichman: `WithValues` is not the same though, it returns a new logger with the key-value pairs.

The Sprintf is just used to format the first argument (the msg): `Info(msg string, keysAndValues ...any)` :slightly_smiling_face:

pkg/controller/cluster/cluster.go
enrichman: Yes, I need to investigate a bit because it looks like there is no Warn in the logr.Logger interface.. :/

pkg/controller/cluster/cluster.go
enrichman: Gotcha!

pkg/controller/cluster/cluster.go
enrichman: Hm, I think I'll leave this here as an info, just because it looks like there is a "philosophical" discussion going here with logr.Logger (https://github.com/go-logr/logr/issues/335#issuecomment-2521775704)

Basically their argument is all around the concepts explained here: https://ntietz.com/blog/the-only-two-log-levels-you-need-are-info-and-error/

It's a bit opinionated but because of the interface we are using here I'll leave it like this, and maybe we can see if it makes sense to do some changes.

galal-hussein (approved):

PR #269, opened 2025-02-24
Support for multi node in shared mode

galal-hussein:
The PR adds the following:

- Converting the shared agent deployment to deamonset so that it can be scheduled on all nodes of the host cluster
- Using the cluster.Spec.NodeSelector to only schedule the pods of the agent daemonset to specific nodes selected by the user
- Using leader election for controllers running by the k3k-kubelet
- Refactoring the PV/PVC support by:
  - for each pod that consumes a pvc, we create a fake PV (psuedoPV) in the virtual cluster to simulate persistent volume, this will enable the pod to be picked up (scheduled) by the virtual-kubelet and then created on the host cluster


pkg/controller/cluster/agent/shared.go
```diff
@@ -106,14 +105,14 @@ version: %s`,
...
 	labels := map[string]string{
 		"cluster": s.cluster.Name,
 		"type":    "agent",
 		"mode":    "shared",
 	}
 
-	deploy := &apps.Deployment{
+	deploy := &apps.DaemonSet{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Deployment",
```
enrichman, line 117:
```suggestion
			Kind:       "DaemonSet",
```

k3k-kubelet/controller/pod.go
```diff
@@ -0,0 +1,165 @@
...
+	return ctrl.NewControllerManagedBy(virtMgr).
+		For(&v1.Pod{}).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: maxConcurrentReconciles,
+		}).
+		Complete(&reconciler)
+}
+
+func (r *PodReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := ctrl.LoggerFrom(ctx).WithValues("Cluster", r.clusterName, "Pod", req.NamespacedName)
```
enrichman, line 63: This is a bit overloaded, because the logger already has some of the info.

Since it's a PodReconciler then it has a `name` and `namespace` keys with the Pod references. Also the `Cluster` field is missing the `namespace`, so if we have two Clusters in two different namespaces we don't know which one it is. Also a small NIT: keys and messages should be lowercase for consistency. :)

```json
{
    "level": "info",
    "timestamp": "2025-02-25T12:13:37.350Z",
    "msg": "Handling pod with pvc",
    "controller": "pod",
    "controllerGroup": "",
    "controllerKind": "Pod",
    "Pod": {
        "name": "nginx",
        "namespace": "default"
    },
    "namespace": "default",
    "name": "nginx",
    "reconcileID": "1f7b57ad-8b6d-41b7-b61f-edbbec272e76",
    "Cluster": "mycluster"
}
```

pkg/controller/cluster/agent/virtual.go
```diff
@@ -41,7 +41,7 @@ func (v *VirtualAgent) EnsureResources(ctx context.Context) error {
 		v.config(ctx),
 		v.deployment(ctx),
 	); err != nil {
-		return fmt.Errorf("failed to ensure some resources: %w\n", err)
+		return fmt.Errorf("failed to ensure some resources: %w", err)
```
enrichman, line 44: :+1:

enrichman (commented):
A couple of small comments, maybe we need to go through all the logging part to define a coding style.

galal-hussein:
> A couple of small comments, maybe we need to go through all the logging part to define a coding style.

Will fix, yeah I think the logging should be standardized throughout the repo.

enrichman (approved):
LGTM! Thanks, great work :clap:

Issue #268, opened 2025-02-24
Virtual Mode connectivity issue


jp-gouin:
#### **Environment**  
- **K3k Version**: `v0.3.0`  
- **RKE2 Version**: 
```
❯ k get nodes
NAME            STATUS   ROLES                              AGE   VERSION
ip-10-0-0-112   Ready    control-plane,etcd,master,worker   3d    v1.31.5+rke2r1
```
- **Rancher Version**: `2.11.0-alpha5`  
- **Installation Method**: Rancher-managed custom cluster  
- **K3k Deployment Method**: Running K3k on RKE2  

#### **Description**  
When deploying a virtual cluster in `virtual` mode, workloads are stuck in `containerCreating` with the following error
```
 Warning  FailedCreatePodSandBox  12s   kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to get sandbox image "rancher/mirrored-pause:3.6": failed to pull image "rancher/mirrored-pause:3.6": failed to pull and unpack image "docker.io/rancher/mirrored-pause:3.6": failed to resolve reference "docker.io/rancher/mirrored-pause:3.6": failed to do request: Head "https://registry-1.docker.io/v2/rancher/mirrored-pause/manifests/3.6": dial tcp: lookup registry-1.docker.io: Try again
```

#### **Steps to Reproduce**  
1. run on the host `k3kcli cluster create --mode virtual  test`
2. connect to the rancher virtual cluster
3. run `k run --image nginx nginx`
4. describe the pod `k describe po nginx`

#### **Expected Behavior**  
The pod is running

galal-hussein:
okay so after investigating this issue, it seems that we actually need to have different `cluster-cidr` and `service-cidr` than the host cluster for virtual mode, it doesnt really matter if you have multiple virtual clusters with the same cidrs but they still need to be different than the host cluster, I will create a PR clarifying that in the docs

PR #271, opened 2025-02-25
Fix the default CIDRs for both modes

galal-hussein:
#268 

This will fix the defaults of the CIDRs for both modes:

- Virtual mode: 
  - clusterCIDR: 10.52.0.0/16
  - serviceCIDR: 10.53.0.0/16
- Shared mode:
  - clusterCIDR: 10.42.0.0/16
  - serviceCIDR: 10.43.0.0/16

It will also only lookup the host's cluster CIDR in shared mode

### DIFF
--- a/pkg/controller/cluster/cluster.go
+++ b/pkg/controller/cluster/cluster.go
@@ -180,21 +180,21 @@ func (c *ClusterReconciler) reconcileCluster(ctx context.Context, cluster *v1alp
 	}
 
 	cluster.Status.ServiceCIDR = cluster.Spec.ServiceCIDR
-	// only lookup service cidr in shared mode
-	if cluster.Status.ServiceCIDR == "" && cluster.Spec.Mode == v1alpha1.SharedClusterMode {
-		log.Info("looking up service cidr for shared mode")
-
-		cluster.Status.ServiceCIDR, err = c.lookupServiceCIDR(ctx)
-		if err != nil {
-			log.Error(err, "error while looking up Cluster ServiceCIDR")
-		}
-	}
-	// set the default cidrs if not specified
 	if cluster.Status.ServiceCIDR == "" {
-		log.Info("setting default ServiceCIDR")
-		cluster.Status.ServiceCIDR = defaultVirtualServiceCIDR
+		// in shared mode try to lookup the serviceCIDR
 		if cluster.Spec.Mode == v1alpha1.SharedClusterMode {
-			cluster.Status.ServiceCIDR = defaultSharedServiceCIDR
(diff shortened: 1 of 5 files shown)


pkg/controller/cluster/cluster.go
```diff
@@ -171,24 +173,29 @@ func (c *ClusterReconciler) reconcileCluster(ctx context.Context, cluster *v1alp
...
-			serviceCIDR = defaultClusterServiceCIDR
+	}
+	// set the default cidrs if not specified
+	if cluster.Status.ServiceCIDR == "" {
+		log.Info("setting default ServiceCIDR")
+		cluster.Status.ServiceCIDR = defaultVirtualServiceCIDR
+		if cluster.Spec.Mode == v1alpha1.SharedClusterMode {
+			cluster.Status.ServiceCIDR = defaultSharedServiceCIDR
 		}
-		cluster.Status.ServiceCIDR = serviceCIDR
```
enrichman, line 191: > [!WARNING]
> **Untested**

What about putting them in the same block? Imho it looks simpler to read:

```go
if cluster.Status.ServiceCIDR == "" {
	// in shared mode try to lookup the serviceCIDR
	if cluster.Spec.Mode == v1alpha1.SharedClusterMode {
		log.Info("looking up service cidr for shared mode")
		cluster.Status.ServiceCIDR, err = c.lookupServiceCIDR(ctx)
		if err != nil {
			log.Error(err, "error while looking up Cluster ServiceCIDR")
			cluster.Status.ServiceCIDR = defaultSharedServiceCIDR
		}
	}

	// in virtual mode assign a default serviceCIDR
	if cluster.Spec.Mode == v1alpha1.VirtualClusterMode { {
		cluster.Status.ServiceCIDR = defaultVirtualServiceCIDR
	}
}
```

The if-else looks a bit more "crowded", but I'll let you choose :smile: 

```go
if cluster.Status.ServiceCIDR == "" {
	// only lookup service cidr in shared mode
	if cluster.Spec.Mode == v1alpha1.SharedClusterMode {
		log.Info("looking up service cidr for shared mode")
		cluster.Status.ServiceCIDR, err = c.lookupServiceCIDR(ctx)
		if err != nil {
			log.Error(err, "error while looking up Cluster ServiceCIDR")
			cluster.Status.ServiceCIDR = defaultSharedServiceCIDR
		}
	} else {
		// assign a default serviceCIDR in virtual mode
		cluster.Status.ServiceCIDR = defaultVirtualServiceCIDR
	}
}
```

enrichman (approved):
Thanks, LGTM!

PR #279, opened 2025-02-28
Update CRDs documentation

enrichman:
This PR revisit and complete the documentation of the CRD types. I added the docs to all the missing fields, corrected a couple, and regenerated the markdown. I also dropped the `set_types.go` file for consistency, and readability.


pkg/apis/k3k.io/v1alpha1/set_types.go
```diff
@@ -1,86 +0,0 @@
-package v1alpha1
```
galal-hussein, line 1: Are these files mistakenly deleted?

galal-hussein: nvm, I missed the description, LGTM

galal-hussein (approved):

pkg/apis/k3k.io/v1alpha1/set_types.go
enrichman: :smile: thanks!

Issue #164, opened 2024-12-20
Deleting a Cluster will not delete the created resources


enrichman:
When a Cluster is deleted all the created resources within the cluster should be cleaned up.

The workloads, secrets and possibly other resources are still present after a deletion.

```
k get all,cm,secret,ing -l 'k3k.io/clusterName=example-cluster'
```

PR #281, opened 2025-03-03
Fix delete cli

galal-hussein:
- Adds --keep-data flag to the delete cli.
  - when this flag is used both webhook secret and token secret will not be removed.
  - when this flag is used the PVCs of the cluster will not be removed.
- Default the delete cli to remove the entire cluster.

- #164

### DIFF
--- a/cli/cmds/cluster_delete.go
+++ b/cli/cmds/cluster_delete.go
@@ -112,11 +112,6 @@ func RemoveOwnerReferenceFromSecret(ctx context.Context, name string, cl client.
 		return err
 	}
 
-	owners := secret.GetOwnerReferences()
-	if len(owners) < 1 {
-		return nil
-	}
-
 	if controllerutil.HasControllerReference(&secret) {
 		if err := controllerutil.RemoveOwnerReference(&cluster, &secret, cl.Scheme()); err != nil {
 			return err
--- a/cli/cmds/cluster_delete.go
+++ b/cli/cmds/cluster_delete.go
@@ -117,8 +117,10 @@ func RemoveOwnerReferenceFromSecret(ctx context.Context, name string, cl client.
 		return nil
 	}
 
-	if err := controllerutil.RemoveOwnerReference(&cluster, &secret, cl.Scheme()); err != nil {
-		return err
+	if controllerutil.HasControllerReference(&secret) {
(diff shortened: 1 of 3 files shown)


cli/cmds/cluster_delete.go
```diff
@@ -55,6 +65,52 @@ func delete(clx *cli.Context) error {
...
+		Namespace: cluster.Namespace,
+	}
+
+	if err := cl.Get(ctx, key, &secret); err != nil {
+		return err
+	}
+
+	if err := controllerutil.RemoveOwnerReference(&cluster, &secret, cl.Scheme()); err != nil {
+		return client.IgnoreNotFound(err)
+	}
```
enrichman, line 113: The `RemoveOwnerReference` does not fail with a ErrNotFound, but the `Get` yes. We should do the check there, and log. Also we should check if the object has the owner reference, or it will return an error.

https://github.com/kubernetes-sigs/controller-runtime/blob/v0.20.3/pkg/controller/controllerutil/controllerutil.go#L142

cli/cmds/cluster_delete.go
```diff
@@ -55,6 +65,52 @@ func delete(clx *cli.Context) error {
...
+
+		// skip removing webhook secret
+		if err := RemoveOwnerReferenceFromSecret(ctx, agent.WebhookSecretName(cluster.Name), ctrlClient, cluster); err != nil {
+			return err
+		}
+	}
+
+	if err := ctrlClient.Delete(ctx, &cluster); err != nil {
+		return client.IgnoreNotFound(err)
+	}
```
enrichman, line 83: Maybe if not found we should still try to delete the data. Should we swap the deletions?

```go
if keepData {
    removeOwnerRefs
} else {
    deletePVCs
}
deleteCluster
```

cli/cmds/cluster_delete.go
```diff
@@ -55,6 +66,62 @@ func delete(clx *cli.Context) error {
...
+			return nil
+		}
+
+		return err
+	}
+
+	owners := secret.GetOwnerReferences()
+	if len(owners) < 1 {
+		return nil
+	}
```
enrichman, line 118:
```suggestion
```

enrichman (approved):
A small improvement, LGTM btw. :)

PR #288, opened 2025-03-04
cli - fix bug with Kubeconfig value resolution

jp-gouin:
Fix a regression with the cli.
The location of the `kubeconfig` is not resolved anymore and lead an error when the `KUBECONFIG` variable is not defined and the `--kubeconfig` is not passed

```
-> % k3kcli kubeconfig generate --name mycluster2 --namespace k3k-mycluster2 --kubeconfig-server localhost
FATA[0000] stat $HOME/.kube/config: no such file or directory
```

This fix solve the issue by reverting the value to 
```
Value:       os.Getenv("HOME") + "/.kube/config",
```
and using `DefaultText ` to describe the value and provide a static output for the cli documentation

### DIFF
--- a/cli/cmds/root.go
+++ b/cli/cmds/root.go
@@ -2,6 +2,7 @@ package cmds
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/rancher/k3k/pkg/apis/k3k.io/v1alpha1"
 	"github.com/rancher/k3k/pkg/buildinfo"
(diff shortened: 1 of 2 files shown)


galal-hussein (approved):

VestigeJ:
This was validated here but there's a new sub-issue elsewhere opting to move this to done. It really ideally should only be issues moved over into to test instead of the PRs themselves.

PR #299, opened 2025-03-06
Add documentation to install k3k on k3d

rossella:
k3d is a valid local option for testing k3k. Adding documentation on how to install k3k on top of k3d.


docs/install-k3d.md
```diff
@@ -0,0 +1,60 @@
...
+
+Since k3d uses docker under the hood, we need to expose the ports on the host that we'll then use for the NodePort in virtual cluster creation.
+
+Create the k3d cluster in the following way:
+
+```bash
+k3d cluster create k3k -p "30000-30010:30000-30010@server:0"
+```
+
+With this syntax ports from 30000 to 20010 will be exposed on the host.
```
jp-gouin, line 15:
```suggestion
With this syntax ports from 30000 to 30010 will be exposed on the host.
```

docs/install-k3d.md
```diff
@@ -0,0 +1,60 @@
...
+Check when the cluster is ready:
+
+```bash
+kubectl get po -n k3k-mycluster
+```
+
+Last thing to do is to get the kubeconfig to connect to the virtual cluster we've just created:
+
+```bash
+k3kcli kubeconfig generate --name mycluster --namespace k3k-mycluster --kubeconfig-server localhost:30001
```
jp-gouin, line 59: You should omit the port-> `--kubeconfig-server localhost`  
otherwise I think you'll end up with a Kubeconfig looking like : 

```
...
  server: "https://localhost:30001:30001"
```

rossella: you need it to overwrite the kubeconfig with the correct port value. It's true that there's a bug and the port gets only added and doesn't remove the old one. @enrichman shall I file a bug or you fixed it already?

jp-gouin: The port is automatically get from the deployed service so you don't need to specify it. 
Now maybe we also want to be able to override it

rossella: Yes but the port that you get from the service it's not going to be reachable. In k3d the host is a docker container, we need to expose the ports that we want to use, that's done in the k3d cluster creation.

enrichman: @jp-gouin as Rossella said we have the port from the service, while we want to use the one from the API server.

@rossella the bug still exist. We can easily fix it, but we probably need some more thoughts about the generated kubeconfig. I.e. how to handle the Rancher URLs, or any other domains.

enrichman:
Since K3d is probably meant to be used only locally for some development maybe it makes sense to have this doc in the `docs/development.md` section. WDYT?

enrichman (approved):
LGTM!

jp-gouin:
Lgtm thanks @rossella

jp-gouin (approved):

PR #308, opened 2025-03-17
Initial support for ResourceQuotas in clustersets

galal-hussein:
The PR changes the following:

- Removes the old Limits/MaxLimits spec in both clustersets and in clusters.
- Adds a `Quota` field in the clusterset spec which will translates to a k8s ResourceQuota Object per namespace.


pkg/controller/clusterset/clusterset.go
```diff
@@ -340,3 +329,100 @@ func (c *ClusterSetReconciler) reconcileClusters(ctx context.Context, clusterSet
...
+		items = []v1.LimitRangeItem{
+			{
+				Type: v1.LimitTypeContainer,
+				DefaultRequest: v1.ResourceList{
+					v1.ResourceCPU:    resource.MustParse(defaultCPURequest),
+					v1.ResourceMemory: resource.MustParse(defaultMemoryRequest),
+				},
+			},
+		}
+	}
```
enrichman, line 413: Hm, I have the feeling that this check could be overcome. We are setting a default value only if the `clusterSet.Spec.DefaultLimits` is nil, but since this is now a `[]v1.LimitRangeItem` then the user could set a different type if `LimitRangeItem` (i.e. if he set just a LimitTypePersistentVolumeClaim), and he will have an unbound limit for the containers.

I think I'm fine with this, maybe we should not provide a default at all, and let the users handle their limits. Obviously documenting the behavior.

I think setting defaults could sometimes being kind of unexpected. Not sure.

pkg/controller/clusterset/clusterset.go
```diff
@@ -340,3 +329,100 @@ func (c *ClusterSetReconciler) reconcileClusters(ctx context.Context, clusterSet
...
+}
+
+func limitRange(clusterSet *v1alpha1.ClusterSet) v1.LimitRange {
+	items := clusterSet.Spec.DefaultLimits
+	if items == nil {
+		// if defaultlimits is not specified we attempt to have a default value for
+		// requests for cpu/memory for the pods within the clusterset
+		items = []v1.LimitRangeItem{
+			{
+				Type: v1.LimitTypeContainer,
```
enrichman, line 406: Shuld this be `LimitTypePod` or `LimitTypeContainer`?

pkg/controller/clusterset/clusterset_test.go
```diff
@@ -666,5 +671,127 @@ var _ = Describe("ClusterSet Controller", Label("controller"), Label("ClusterSet
...
+								v1.ResourceMemory: resource.MustParse("1Gi"),
+							},
+						},
+					},
+				}
+
+				err := k8sClient.Create(ctx, clusterSet)
+				Expect(err).To(Not(HaveOccurred()))
+				time.Sleep(3 * time.Second)
+				// wait for a bit for the ResourceQuota to be created
```
enrichman, line 695: The Eventually is used exactly for this, so I would remove the Sleep. It could also cause some flaky tests, what if the ResourceQuota takes more than 3s?

In the next Eventually I would just poll for the ResourceQuota. Once ready we can check for the values.

pkg/controller/clusterset/clusterset.go
galal-hussein: The issue is that without a default limit the resourceQuota will not work, because Memory/CPU must be specified for all pods within a resourceQuota's namespace:

From Docs:

> If quotas are enabled in a namespace for compute resources like cpu and memory, users must specify requests or limits for those values; otherwise, the quota system may reject pod creation. Hint: Use the LimitRanger admission controller to force defaults for pods that make no compute resource requirements.

pkg/controller/clusterset/clusterset.go
galal-hussein: I think the request/limit is added to the containers and not on pods

charts/k3k/crds/k3k.io_clustersets.yaml
```diff
@@ -60,29 +60,72 @@ spec:
...
-                    type: object
-                  workerLimit:
-                    additionalProperties:
-                      anyOf:
-                      - type: integer
-                      - type: string
-                      pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
-                      x-kubernetes-int-or-string: true
-                    description: WorkerLimit specifies resource limits for agent nodes.
-                    type: object
```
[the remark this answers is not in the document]
galal-hussein: I can revert this back and add functionality to workerLimit as well because from what I can see it was unusable so far

pkg/apis/k3k.io/v1alpha1/types.go
```diff
@@ -338,10 +334,16 @@ type ClusterSet struct {
...
 
-	// DefaultLimits specifies the default resource limits for servers/agents when a cluster in the set doesn't provide any.
+	// Quota specifies the resource limits for clusters within a clusterset.
 	//
 	// +optional
-	DefaultLimits *ClusterLimit `json:"defaultLimits,omitempty"`
+	Quota *v1.ResourceQuotaSpec `json:"quota,omitempty"`
+
+	// Limit specifies the default value for a limitRange that will be applied to all pods within
+	// the clusterSet, if not specified a default value will be applied.
```
enrichman, line 343: Maybe we need to update the doc, because the LimitRange can be used also to set the min/max resource requests.

Wdyt about something like this?

```suggestion
	// Limit specifies the LimitRange that will be applied to all pods within the ClusterSet
	// to set defaults and constraints (min/max)
```

pkg/controller/clusterset/clusterset.go
```diff
@@ -143,27 +144,14 @@ func (c *ClusterSetReconciler) Reconcile(ctx context.Context, req reconcile.Requ
...
-	// 				},
-	// 			},
-	// 		},
-	// 	}
-	// 	quota.Spec.Hard = clusterSet.Spec.MaxLimits
-	// 	if err := c.Client.Create(ctx, &quota); err != nil {
-	// 		return reconcile.Result{}, fmt.Errorf("unable to create resource quota from cluster set: %w", err)
-	// 	}
-	// }
+	if err := c.reconcileDefaultLimits(ctx, &clusterSet); err != nil {
```
enrichman, line 147:
```suggestion
	if err := c.reconcileLimit(ctx, &clusterSet); err != nil {
```

pkg/controller/clusterset/clusterset.go
```diff
@@ -340,3 +328,100 @@ func (c *ClusterSetReconciler) reconcileClusters(ctx context.Context, clusterSet
...
+		},
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ResourceQuota",
+			APIVersion: "v1",
+		},
+		Spec: *clusterSet.Spec.Quota,
+	}
+}
+
+func (c *ClusterSetReconciler) reconcileDefaultLimits(ctx context.Context, clusterSet *v1alpha1.ClusterSet) error {
```
enrichman, line 379:
```suggestion
func (c *ClusterSetReconciler) reconcileLimit(ctx context.Context, clusterSet *v1alpha1.ClusterSet) error {
```

pkg/controller/clusterset/clusterset.go
```diff
@@ -340,3 +328,100 @@ func (c *ClusterSetReconciler) reconcileClusters(ctx context.Context, clusterSet
...
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("Reconciling Default Limits")
```
enrichman, line 381:
```suggestion
	log.Info("Reconciling Limit")
```

pkg/controller/clusterset/clusterset_test.go
```diff
@@ -8,18 +8,23 @@ import (
...
 	"k8s.io/utils/ptr"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
+const (
+	CPUResourceTest    = "800m"
+	MemoryResourceTest = "1Gi"
+)
```
enrichman, line 26: It looks like these are used only once. Maybe we can drop them altogether, or use them in different tests. :+1:

pkg/controller/clusterset/clusterset_test.go
```diff
@@ -666,5 +671,137 @@ var _ = Describe("ClusterSet Controller", Label("controller"), Label("ClusterSet
...
+				// wait for a bit for the resourceQuota to be deleted
+				Eventually(func() bool {
+					key := types.NamespacedName{
+						Name:      k3kcontroller.SafeConcatNameWithPrefix(clusterSet.Name),
+						Namespace: namespace,
+					}
+					err := k8sClient.Get(ctx, key, &resourceQuota)
+					return apierrors.IsNotFound(err)
+				}).
+					MustPassRepeatedly(5).
```
enrichman, line 759: Do we need the `MustPassRepeatedly`?

pkg/controller/clusterset/clusterset_test.go
```diff
@@ -666,5 +671,137 @@ var _ = Describe("ClusterSet Controller", Label("controller"), Label("ClusterSet
...
+						Namespace: namespace,
+					}
+					return k8sClient.Get(ctx, key, &limitRange)
+				}).
+					WithTimeout(time.Minute).
+					WithPolling(time.Second).
+					Should(BeNil())
+
+				// make sure that default limit range has the default requet values.
+				Expect(len(limitRange.Spec.Limits) > 0).To(BeTrue())
```
enrichman, line 801:
```suggestion
				Expect(limitRange.Spec.Limits).ShouldNot(BeEmpty())
```

enrichman (commented):
Thanks @galal-hussein, it looks great! I left few comments, but overall it looks good.

pkg/apis/k3k.io/v1alpha1/types.go
galal-hussein: Yes you are correct, I will update the docs as well

pkg/controller/clusterset/clusterset_test.go
galal-hussein: I dont think so, but I added as an ensurance because it may take some time

enrichman (approved):

Issue #312, opened 2025-03-25
Create Cluster in dedicated namespace


enrichman:
Currently new Clusters are created in the targeted namespace. This will lead having multiple virtual cluster in the same namespace, and it could be a bit unexpected.

Has a default it's better to create a dedicated namespace for each Cluster (`k3k-clustername`), and let the user specify the namespace if they want to use a particular one (with the `--namespace` flag).

The UI already implemented this behavior.

PR #314, opened 2025-03-25
Create dedicated namespace for new clusters

### LINKED ISSUE #313: Kubeconfig files conflicts
The cli use just the Cluster name  for the kubeconfig files. This means that if two clusters with the same name but in two different namespace are created then the two kubeconfig files will conflicts.

enrichman:
This PR updates the cli to create a dedicated namespace for new virtual clusters, matching the UI behavior. (#312 )

New Clusters will be created in a `k3k-<CLUSTER NAME>` namespace, to isolate better its resources (easing the cleanup), and to avoid unexpected result. The user will still be able to create a cluster in a specific namespace with the `--namespace` flag.

This PR also fixes #313 adding the namespace to the generated kubeconfig file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,9 +89,8 @@ lint:	## Find any linting issues in the project
 validate: build-crds docs ## Validate the project checking for any dependency or doc mismatch
 	$(GINKGO) unfocus
 	go mod tidy
-	git --no-pager diff go.mod go.sum
 	git status --porcelain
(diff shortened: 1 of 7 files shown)


jp-gouin (approved):
Functionally speaking - LGTM 
Thanks @enrichman

galal-hussein (approved):

PR #316, opened 2025-03-28
Added ClusterSet as singleton

enrichman:
This PR enforce the existence of only one ClusterSet per namespace.

To enforce this without implementing a Webhook we are enforcing the resource to have a `default` name, and also we create a ResourceQuota with a `"count/clustersets.k3k.io": "1"` in the namespace. 

To have also a human readable name a new DisplayName field (like Projects) was added. This displayName is shown in the output thanks to the `additionalPrinterColumns` property:

```
-> % kubectl get clustersets.k3k.io -A                   
NAMESPACE           NAME        DISPLAY NAME    AGE
another             default     my-clusterset   24h
k3k-my-clusterset   default     my-clusterset   77m
```

### DIFF
--- a/pkg/controller/clusterset/clusterset.go
+++ b/pkg/controller/clusterset/clusterset.go
@@ -48,6 +48,7 @@ func Add(ctx context.Context, mgr manager.Manager, clusterCIDR string) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.ClusterSet{}).
 		Owns(&networkingv1.NetworkPolicy{}).
+		Owns(&v1.ResourceQuota{}).
 		WithOptions(controller.Options{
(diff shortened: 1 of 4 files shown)


pkg/controller/clusterset/clusterset.go
```diff
@@ -340,3 +331,31 @@ func (c *ClusterSetReconciler) reconcileClusters(ctx context.Context, clusterSet
 
 	return err
 }
+
+func (c *ClusterSetReconciler) reconcileResourceQuota(ctx context.Context, clusterSet *v1alpha1.ClusterSet) error {
```
galal-hussein, line 335: Can we use a different name here since I am using similar name to reconcile the quota set in the clusterset.spec.quota

galal-hussein (approved):

PR #319, opened 2025-04-01
Add `clusterset` commands

enrichman:
This PR add the `k3kcli clusterset create/delete` commands.

It also refactor the commands a bit to add an `AppContext` where the Client is initialized and stored for every command.

### DIFF
--- a/cli/cmds/cluster.go
+++ b/cli/cmds/cluster.go
@@ -4,13 +4,13 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
-func NewClusterCmd() *cli.Command {
+func NewClusterCmd(appCtx *AppContext) *cli.Command {
 	return &cli.Command{
 		Name:  "cluster",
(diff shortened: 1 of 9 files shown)


andreas-kupries (approved):

galal-hussein (approved):
LGTM, Thanks! @enrichman, are we considering adding a `join cluster` command and `remove cluster` to add/remove clusters from clusterset, or is it implied if clusterset exists in the namespace?

enrichman:
Thanks @galal-hussein. I haven't thought about that, but I since there is no link in the ClusterSet resource about the Clusters I guess it's implied to exists in the same namespace.

I'm thinking a bit about the deletion or creation when there are already Clusters. Should we prompt the user about this, use some `--force` option, delete also the Clusters within the namespace?

If you are ok I'll merge this, and maybe we can think about this in a next iteration! :slightly_smiling_face:

Issue #305, opened 2025-03-13
How do I configure HTTP_PROXY for the virtual mode k3s instance?


BroderPeters:
**Environmental Info:**
Host Cluster Version: 
k3s version v1.31.6+k3s1 (6ab750f9)
go version go1.22.12

k3k version: v0.3.0

Node(s) CPU architecture, OS, and Version: 
24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Feb 17 11:51:52 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

Host Cluster Configuration:
1 server, 1 agent

K3K Cluster Configuration:
```yaml
apiVersion: k3k.io/v1alpha1
kind: Cluster
metadata:
  name: kcp
  namespace: kcp
spec:
  mode: virtual
```

**Describe the bug:**
I'm running k3k behind a proxy that I need to configure inside the virtual mode spawned k3s instance using e.g. HTTP_PROXY as env.

**Steps To Reproduce:**
- Created a cluster with `k3k create` behind a proxy

**Expected behavior:**
- I have an option to provide custom env variables

**Actual behavior:**
- k3k server ran into failed to get image from docker hub

**Additional context / logs:**
- For now I manually adjusted the created statefulset which deploys the k3s images to expose the needed envs.
- Am I missing an option?

jp-gouin:
Hi @BroderPeters , thank you for opening the issue. This is indeed a missing feature. 
User should be able to configure `env` variable for the K3s server

BroderPeters:
@jp-gouin Thanks, good to know!

The most "comfortable" setup I have for now is to spawn a configmap alongside the namespace containing the proxy environment variables and to manually inject that into the server statefulset after that is created.

```bash
kubectl set env --from=configmap/my-k3k-proxy-config -n my-k3k sts/k3k-my-k3k-server
```

Should I rephrase the titel into a feature request or is there already something for it?
Does it need more information?

jp-gouin:
It's alright, I added the labels and correct type. You're all set 🚀

galal-hussein:
This feature is implemented using the extra env feature for both servers/agents

Validation Steps:

- Create a virtual cluster with the following using the cli:

```
k3kcli cluster create --namespace k3k-ns-1 --mode virtual --persistence-type ephemeral --server-envs FOO=BAR test-1
```
you should be able to see the environment variable populated withing the server/agent pod

```
[...]

PR #324, opened 2025-04-08
Add extra env for servers/agents

galal-hussein:
Adding extra env arguments to the cluster spec:

cluster.Spec.ServerEnvs
cluster.Spec.AgentEnvs

The extra envs also can  be passed as a cli flag:
--server-envs
--agent-envs

#305


enrichman (approved):
Looks ok to me, just a small comment

Issue #334, opened 2025-04-29
DownwardAPI volumes are not being updated on pod updates


galal-hussein:
**Environmental Info:**
Host Cluster Version: 

v1.32.3+k3s1
Node(s) CPU architecture, OS, and Version: 

1 node ubuntu 24.04

Host Cluster Configuration:

1 server

K3K Cluster Configuration:

1 server shared mode

**Describe the bug:**

In shared mode when you create a pod with downward api volume annotation, and then update the pod's annotation the volume will not be updated.

**Steps To Reproduce:**
- Created a cluster with `k3k create`: 
```
k3kcli cluster create --namespace shared --mode shared --servers 1 sharedcluster
```
- create a pod with downwardAPI volume in the virtual cluster
```
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  namespace: default
  annotations:
    foo: bar
spec:
  volumes:
  - name: downward
    downwardAPI:
      items:
      - path: "annotaions"
        fieldRef:
          fieldPath: metadata.annotations
  containers:
  - name: nginx
    image: nginx:latest
    volumeMounts:
    - name: downward
      mountPath: /opt/annotations
```
- exec to the pod and make sure that annotation file does have foo bar annotation
```
root@test-pod:/# cat /opt/annotations/annotaions 
foo="bar"
k3k.io/name="test-pod"
k3k.io/namespace="default"
kubectl.kubernetes.io/last-applied-configuration="{\"apiVersion\":\"v1\",\"kind\":\"Pod\",\"metadata\":{\"annotations\":{\"foo\":\"bar\"},\"name\":\"test-pod\",\"namespace\":\"default\"},\"spec\":{\"containers\":[{\"image\":\"nginx:latest\",\"name\":\"nginx\",\"volumeMounts\":[{\"mountPath\":\"/opt/annotations\",\"name\":\"downward\"}]}],\"volumes\":[{\"downwardAPI\":{\"items\":[{\"fieldRef\":{\"fieldPath\":\"metadata.annotations\"},\"path\":\"annotaions\"}]},\"name\":\"downward\"}]}}\n"
kubernetes.io/config.seen="2025-04-29T09:44:07.951891663Z"
```
- update the annotations in the pod and add `foo2: bar2`

**Expected behavior:**

you should see the file is updated in /opt/annotations/annotations with the new annotation

**Actual behavior:**

the file is not updated

**Additional context / logs:**

ShylajaDevadiga:
[...]

PR #335, opened 2025-04-29
Fix annotation update

galal-hussein:
- Fix annotation/label update in shared mode
- Fix a panic that happens in shared mode when running conformance test
- Make sure that hostname doesnt exceed 64 characters

#334

### DIFF
--- a/k3k-kubelet/provider/provider.go
+++ b/k3k-kubelet/provider/provider.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"maps"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


k3k-kubelet/provider/provider.go
```diff
@@ -608,6 +612,10 @@ func (p *Provider) updatePod(ctx context.Context, pod *v1.Pod) error {
 	currentHostPod.Spec.ActiveDeadlineSeconds = pod.Spec.ActiveDeadlineSeconds
 	currentHostPod.Spec.Tolerations = pod.Spec.Tolerations
 
+	// in the virtual cluster we can update also the labels and annotations
+	currentHostPod.Annotations = pod.Annotations
+	currentHostPod.Labels = pod.Labels
+
```
enrichman, line 618: It's probably safer to keep the previous annotations/labels, merging the two maps:

```go
maps.Copy(currentHostPod.Annotations, pod.Annotations)
maps.Copy(currentHostPod.Labels, pod.Labels)
```

enrichman (approved):

Issue #264, opened 2025-02-24
How to specify the image repository for `rancher/k3s:v1.31.3-k3s1` when using Helm install?


kingsd041:
**Description:**  
I installed K3k using Helm with the following command, specifying the image repositories for `image.repository` and `sharedAgent.image.repository`:  

```sh
helm install --namespace k3k-system --create-namespace k3k k3k/k3k --devel \
  --set image.repository=registry.cn-hangzhou.aliyuncs.com/kingsd/k3k \
  --set sharedAgent.image.repository=registry.cn-hangzhou.aliyuncs.com/kingsd/k3k-kubelet
```

However, when I run `k3kcli cluster create k3k-7`, the `k3k-k3k-7-server-0` pod still tries to pull the image from Docker Hub instead of my specified registry.  

I couldn't find a configuration option for setting the repository of `rancher/k3s:v1.31.3-k3s1` in the [[values.yaml](https://github.com/rancher/k3k/blob/v0.3.0/charts/k3k/values.yaml)](https://github.com/rancher/k3k/blob/v0.3.0/charts/k3k/values.yaml).  

**Expected Behavior:**  
The cluster should pull the `rancher/k3s:v1.31.3-k3s1` image from my specified registry.  

**Actual Behavior:**  
The cluster still pulls `rancher/k3s:v1.31.3-k3s1` from Docker Hub.  

**Steps to Reproduce:**  
1. Install K3k using the above Helm command.  
2. Create a cluster with `k3kcli cluster create k3k-7`.  
3. Check the image pull behavior of `k3k-k3k-7-server-0`.  

**Question:**  
How can I specify the registry for `rancher/k3s:v1.31.3-k3s1` when installing K3k via Helm?

enrichman:
Hi @kingsd041, this is not supported (yet?). We can think about having a global flag provided during the K3k installation, that can be eventually overridden on a per-cluster basis (if it makes sense).

PR #359, opened 2025-05-26
add air-gap support

jp-gouin:
Related issue: https://github.com/rancher/k3k/issues/264 

This PR add the air-gap support.
User can specify the registry of the `k3s` server (and agent) in the helm chart `values` using 

```
k3sServer:
  image:
    repository: rancher/k3s
    pullPolicy: "" 
```

When deploying a cluster the specified repository will be used. 
If not specified in the `values` it'll default to `rancher/k3s`

- Add support of `pullPolicy` for the `k3s` server pod 
- Add a `how-to` guide to explain how to use air-gap functionality


enrichman (approved):
We will probably need to refactor a bit some struct, because some functions are getting lot of arguments! LGTM

PR #368, opened 2025-05-30
Added `k3kcli cluster list` and `k3kcli policy list` commands

### LINKED ISSUE #352: Policy integration in k3kcli
User should be able to use the `k3kcli` to 
- Create/Delete policy
- List policies also show which cluster/namespace are bound to the policy

enrichman:
This PR adds the `list` commands to the clusters and policies.

```
-> % k3kcli cluster list                       
NAMESPACE        NAME         MODE
k3k-mycluster2   mycluster2   shared
k3k-mycluster3   mycluster3   shared
```
```
-> % k3kcli policy list 
NAME     MODE
mypol2   shared
simple   shared
```

It leverages the already imported `k8s.io/cli-runtime` package (`k8s.io/cli-runtime/pkg/printers`) so it doesn't add any extra dependency, and it uses the CRD printerColumn definition to print the columns (with the `jsonpath`).

This will give the option to just update the CRD spec in order to have additional columns, and it will work either with the `k3kcli` and `kubectl`:

```
-> % kubectl get vcp     
NAME     MODE
mypol2   shared
simple   shared
```
```
-> % kubectl get clusters.k3k.io -A
NAMESPACE        NAME         MODE
k3k-mycluster2   mycluster2   shared
k3k-mycluster3   mycluster3   shared
```

**Note:** since the `vcp` alias was added we can get policies with `kubectl get vcp`

### DIFF
--- a/cli/cmds/policy_list.go
+++ b/cli/cmds/policy_list.go
@@ -6,11 +6,8 @@ import (
 	"github.com/rancher/k3k/pkg/apis/k3k.io/v1alpha1"
 	"github.com/urfave/cli/v2"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/cli-runtime/pkg/printers"
-	"k8s.io/client-go/util/jsonpath"
 )
 
 func NewPolicyListCmd(appCtx *AppContext) *cli.Command {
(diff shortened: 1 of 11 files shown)


cli/cmds/policy_list.go
```diff
@@ -0,0 +1,129 @@
...
+		})
+	}
+
+	return &metav1.Table{
+		TypeMeta:          metav1.TypeMeta{APIVersion: "meta.k8s.io/v1", Kind: "Table"},
+		ColumnDefinitions: columnDefinitions,
+	}
+}
+
+func makeRow(obj runtime.Object, printerColumns []apiextensionsv1.CustomResourceColumnDefinition) metav1.TableRow {
```
galal-hussein, line 93: nit: Since this is used by policy list and cluster list can we move it to separate file including makerow,makecells, etc.

galal-hussein (approved):

PR #372, opened 2025-06-04
[CLI] add storage-request-size flag

jp-gouin:
This PR adds `storage-request-size` flag in the cli .

User can specify the size of the storage using the cli 
```
k3kcli cluster create cluster-persistent \
  --persistence-type=dynamic \
  --storage-class-name=local-path
  --storage-request-size=2Gi
```

If the size is invalid the cli will throw an error : 
```
./k3kcli cluster create  --storage-request-size wrong-value test-persistent
FATA[0000] invalid storage size, should be a valid resource quantity e.g "10Gi"
```

### DIFF
--- a/cli/cmds/cluster_create_flags.go
+++ b/cli/cmds/cluster_create_flags.go
@@ -67,9 +67,6 @@ func NewCreateFlags(config *CreateConfig) []cli.Flag {
 			Destination: &config.storageRequestSize,
 			Action: func(ctx *cli.Context, value string) error {
 				if _, err := resource.ParseQuantity(value); err != nil {
-				  return errors.New(`invalid storage size, should be a valid resource quantity e.g "10Gi"`) 
-				}
-				if err != nil {
 					return errors.New(`invalid storage size, should be a valid resource quantity e.g "10Gi"`)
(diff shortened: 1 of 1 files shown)


cli/cmds/cluster_create_flags.go
```diff
@@ -60,6 +61,18 @@ func NewCreateFlags(config *CreateConfig) []cli.Flag {
 			Usage:       "storage class name for dynamic persistence type",
 			Destination: &config.storageClassName,
 		},
+		&cli.StringFlag{
+			Name:        "storage-request-size",
+			Usage:       "storage size for dynamic persistence type",
+			Destination: &config.storageRequestSize,
+			Action: func(ctx *cli.Context, value string) error {
+				_, err := resource.ParseQuantity(value)
```
galal-hussein: sinec we dont need the value we can wrap the error
```suggestion
				if _, err := resource.ParseQuantity(value); err != nil {
				  return errors.New(`invalid storage size, should be a valid resource quantity e.g "10Gi"`) 
				}
```

enrichman (approved):
LGTM!

PR #373, opened 2025-06-04
[DOC] add how-to create virtual clusters

jp-gouin:
This how-to docs add various example to create a virtual cluster using both the CRD and the CLI equivalent (if possible)

### DIFF
--- a/docs/howtos/create-virtual-clusters.md
+++ b/docs/howtos/create-virtual-clusters.md
@@ -22,7 +22,6 @@ kind: Cluster
 metadata:
   name: k3kcluster-ingress
 spec:
-  mode: shared
   tlsSANs:
     - my-cluster.example.com
   expose:
@@ -30,11 +29,11 @@ spec:
       ingressClassName: nginx
       annotations:
         nginx.ingress.kubernetes.io/ssl-passthrough: "true"
-        nginx.ingress.kubernetes.io/backend-protocol: "true"
+        nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
         nginx.ingress.kubernetes.io/ssl-redirect: "HTTPS"
 ```
 
-This will create a virtual cluster and expose it via an ingress with the specified hostname.
+This will create a virtual cluster in `shared` mode and expose it via an ingress with the specified hostname.
 
 ### CLI Method
 
@@ -42,7 +41,7 @@ This will create a virtual cluster and expose it via an ingress with the specifi
 
 ---
 
-## Use Case: Create a Virtual Cluster with Persistent Storage
+## Use Case: Create a Virtual Cluster with Persistent Storage (**Default**)
 
 ### CRD Method
 
@@ -60,7 +59,7 @@ spec:
 
 This ensures that the virtual cluster stores its state persistently with a 30Gi volume.  
 If `storageClassName` is not set it will default to the default StorageClass.  
-If `storageRequestSize` is not set it will request a 1Gi volume.  
+If `storageRequestSize` is not set it will request a 1Gi volume by default.  
 
 ### CLI Method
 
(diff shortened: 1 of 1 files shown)


docs/howtos/create-virtual-clusters.md
```diff
@@ -0,0 +1,295 @@
...
+apiVersion: k3k.io/v1alpha1
+kind: Cluster
+metadata:
+  name: k3kcluster-persistent
+spec:
+  persistence:
+    type: dynamic
+    storageClassName: local-path
+    storageRequestSize: 30Gi
+```
```
enrichman: Maybe we should say that this is the default.

docs/howtos/create-virtual-clusters.md
```diff
@@ -0,0 +1,295 @@
...
+If `storageClassName` is not set it will default to the default StorageClass.  
+If `storageRequestSize` is not set it will request a 1Gi volume.  
+
+### CLI Method
+
+```sh
+k3kcli cluster create \
+  --persistence-type dynamic \
+  --storage-class-name local-path \
+  --storage-request-size 30Gi \
```
enrichman: If we want to merge this doc before the rc we need to drop this part, and update it again once #372 is merged.

docs/howtos/create-virtual-clusters.md
```diff
@@ -0,0 +1,295 @@
...
+### CLI Method
+
+```sh
+k3kcli cluster create \
+  --agents 1 \
+  --mode virtual \
+  k3kcluster-virtual
+```
+
+---
```
enrichman: Maybe we can add a more interesting case, like 3 servers and 3 agents in virtual mode. And also stating that agents make sense only in virtual mode.

docs/howtos/create-virtual-clusters.md
```diff
@@ -0,0 +1,295 @@
...
+For `virtual` mode, make sure to set `agents` to `3` or more.
+
+### CLI Method
+
+```sh
+k3kcli cluster create \
+  --servers 3 \
+  --persistence-type dynamic \
+  k3kcluster-ha
+```
```
enrichman: If the `dynamic` persistence mode is the default I would skip that part.

docs/howtos/create-virtual-clusters.md
```diff
@@ -0,0 +1,295 @@
...
+This will create an ephemeral virtual cluster with no persistence.  
+
+### CLI Method
+
+```sh
+k3kcli cluster create \
+  --servers 1 \
+  --persistence-type ephemeral \
+  k3kcluster-ephemeral
+```
```
enrichman: I would skip the servers: 1, if it's the default.

docs/howtos/create-virtual-clusters.md
```diff
@@ -0,0 +1,295 @@
...
+
+This sets the virtual cluster's Kubernetes version explicitly.
+
+### CLI Method
+
+```sh
+k3kcli cluster create \
+  --version v1.33.1-k3s1 \
+  k3kcluster-custom-k8s
```
enrichman: Maybe we can add few lines to say we are supporting `k3s` and because of that we need the `version` to be one of the K3s one.

docs/howtos/create-virtual-clusters.md
```diff
@@ -0,0 +1,295 @@
...
+apiVersion: k3k.io/v1alpha1
+kind: Cluster
+metadata:
+  name: k3kcluster-node-placed
+spec:
+  nodeSelector:
+    disktype: ssd
+```
+
+This places the virtual cluster on nodes with the label `disktype: ssd`.
```
enrichman: It's probably worth to say that, in shared mode, also the workloads will run in the selected nodes.

docs/howtos/create-virtual-clusters.md
```diff
@@ -0,0 +1,295 @@
...
+This configures an HTTP proxy for all nodes in the virtual cluster.
+
+### CLI Method
+
+```sh
+k3kcli cluster create  \
+  --server-envs HTTP_PROXY=http://abc.xyz \
+  --agent-envs HTTP_PROXY=http://abc.xyz \
+  k3kcluster-http-proxy
+```
```
enrichman: Maybe it's worth to mention that the same vars can be used to provide custom env variables to the server and agents.

docs/howtos/create-virtual-clusters.md
jp-gouin: Alright, I'll rewrite it to  *Create a Highly Available Virtual Cluster in `virtual` mode* with a disclaimer about `agents`

PR #377, opened 2025-06-11
Add PriorityClass reconciler (virtual cluster -> host)

enrichman:
This PR adds the PriorityClass sync from the virtual cluster to the host cluster.

With this a user will be able to create and assign priority classes also in shared mode, otherwise the priority class for the pod scheduled in the host would failed because of the missing priority class.

If a priorityClass is set as global default then this flag will need to be removed in the host, because only one priority class could be set as default. An annotation to keep track of this will be used (`priorityclass.k3k.io/globalDefault`).

This PR adds a `k3k-priorityclass` ClusterRole, needed by the k3k-kubelet to manage PriorityClasses in the host cluster.

It also add tests to the virtual kubelet, to test the sync of the resources from the virtual cluster to the host.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,10 @@ test-unit:	## Run the unit tests (skips the e2e)
 test-controller:	## Run the controller tests (pkg/controller)
 	$(GINKGO) -v -r pkg/controller
 
+.PHONY: test-kubelet-controller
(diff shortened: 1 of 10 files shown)


galal-hussein (approved):
LGTM Thanks!

PR #385, opened 2025-06-19
fix-howto usage of serverEnvs and agentEnvs

jp-gouin:
Fix `serverEnvs` and `agentEnvs` usage in the how-to doc


galal-hussein (approved):
Looks good, thanks! LGTM

Issue #347, opened 2025-05-07
Node labels/annotations are not reflected to virtual nodes in shared mode


galal-hussein:
**Environmental Info:**
Host Cluster Version: 

v1.32.3+k3s1
Node(s) CPU architecture, OS, and Version: 

1 node ubuntu 24.04

Host Cluster Configuration:

1 server

K3K Cluster Configuration:

1 server shared mode

**Describe the bug:**

In shared mode there are psuedo nodes that holds the same node names of the host cluster, however any annotation and labels added to that node are not reflected to the virtual cluster, this causes a failure in conformance test 

```
 [FAIL] [sig-scheduling] SchedulerPredicates [Serial] [It] validates resource limits of pods that are allowed to run [Conformance] [sig-scheduling, Serial, Conformance]
  k8s.io/kubernetes/test/e2e/scheduling/predicates.go:414
```

Related to #338

galal-hussein:
blocked, see https://github.com/rancher/k3k/issues/338#issuecomment-2979760618

galal-hussein:
Sorry there has been slight mistake in the description of the issue, The validation steps should be as following:
- On the host cluster label and annotate the node. for example `foo=bar`
- Create a virtual cluster with the following flag: `--mirror-host-nodes`
- make sure that the node/annotation added in step 1 are reflected to the virtual node
- make sure that the conformance test for this test passes
```
validates resource limits of pods that are allowed to run
```

mdrahman-suse:
# Validated on v0.3.4-rc3

`--mirror-host-nodes` is visible on help

```
$ k3kcli -v
k3kcli version v0.3.4-rc3

$ k3kcli cluster create --help
Create new cluster

Usage:
  k3kcli cluster create [flags]

Examples:
k3kcli cluster create [command options] NAME

Flags:
      --agent-args strings            agents extra arguments
      --agent-envs strings            agents extra Envs
      --agents int                    number of agents
      --cluster-cidr string           cluster CIDR
      --custom-certs string           The path for custom certificate directory
  -h, --help                          help for create
      --kubeconfig-server string      override the kubeconfig server host
      --mirror-host-nodes             Mirror Host Cluster Nodes
      --mode string                   k3k mode type (shared, virtual) (default "shared")
  -n, --namespace string              namespace of the k3k cluster
      --persistence-type string       persistence mode for the nodes (dynamic, ephemeral, static) (default "dynamic")
      --policy string                 The policy to create the cluster in
[...]

Issue #338, opened 2025-05-01
NodePort being unreachable from k3k-kubelet


galal-hussein:
**Environmental Info:**
Host Cluster Version: 

v1.32.3+k3s1
Node(s) CPU architecture, OS, and Version: 

1 node ubuntu 24.04

Host Cluster Configuration:

1 server

K3K Cluster Configuration:

1 server shared mode

**Describe the bug:**

Services in shared mode are synced to the host, however since k3k-kubelet acts as a node, NodePorts are expected to be reachable from the k3k-kubelet IP as well as the host node's IP, the current behavior is causing some network conformance tests to fail:

> [FAIL] [sig-network] Services [It] should have session affinity work for NodePort service [LinuxOnly] [Conformance] [sig-network, Conformance]
  k8s.io/kubernetes/test/e2e/network/service.go:4208

> [FAIL] [sig-network] Services [It] should be able to change the type from ExternalName to NodePort [Conformance] [sig-network, Conformance]
  k8s.io/kubernetes/test/e2e/network/service.go:1463

> [FAIL] [sig-network] Services [It] should be able to switch session affinity for NodePort service [LinuxOnly] [Conformance] [sig-network, Conformance]
  k8s.io/kubernetes/test/e2e/network/service.go:4208

**Steps To Reproduce:**
- Created a cluster with `k3k create`: 
```
k3kcli cluster create --namespace shared --mode shared --servers 1 sharedcluster
```
- create a pod and expose it to a nodeport service
```
kubectl run nginx --image=nginx
kubectl expose --type=NodePort --port 80 --target-port 80 pod/nginx
```
- get the logs of the pod
```
kubectl logs downward-api-limits
```

- get the service IP of k3k-kubelet
```
kubectl get service -n shared k3k-sharedcluster-kubelet
k3k-sharedcluster-kubelet                                         ClusterIP   10.43.17.145    <none>        10250/TCP,9443/TCP             13m
```

- try to reach the nodeport through this IP

Suggested solution:

One solution is to have the ability to use real node's information from the host cluster, that will include the IP/Hostname of the actual host nodes, so that node ports can be reachable through the node's IP

galal-hussein:
In order to fix this issue, we have to enable node synchronization, which will allow k3k-kubelet to register a node with real information including the node IP and Hostname, this can be fixed easily however the consequence of this is that it causes logs/exec of the pods fail with this error:

```
[...]

PR #389, opened 2025-06-20
Mirror host nodes

galal-hussein:
The PR adds mirror host nodes feature in shared mode, each k3k-kubelet pod will copy real node information and register a node with the virtual cluster, in this mode the PR also changes the k3k-kubelet to use the `hostNetwork: true` so that features like logs/exec can work normally.

### Mirror host nodes feature:

1- A flag is added to cli --mirror-host-nodes
2- A spec field has been added to the cluster `cluster.spec.MirrorHostNodes`
3- k3k-kubelet has been modified to use the real node information when the field is set to the cluster spec

### Port allocation/deallocation in mirror-host-nodes
When mirror-host-nodes is passed to the virtual cluster the following will happen
1- k3k-controller will create two configmaps called `k3k-kubelet-port-range` and `k3k-webhook-port-range` to store ports allocated for the clusters
2- when a cluster is created k3k-controller will use the `PortAllocator` to allocate a k3k-kubelet port and webhook port for each cluster from a ranges passed to the k3k-controller binary (can be passed through the chart as well)

k3k-kubelet-port-range configmap
```
  apiVersion: v1
  data:
    sharedcluster-shared-x: "50000"
  kind: ConfigMap
  metadata:
    name: k3k-kubelet-port-range
    namespace: k3k-system
```

k3k-webhook-port-range configmap
```
  apiVersion: v1
  data:
    sharedcluster-shared-x: "51001"
  kind: ConfigMap
  metadata:
    name: k3k-webhook-port-range
    namespace: k3k-system
```
3- when a cluster is deleted the controller will deallocate the port from configmaps 
4- the port ranges are defaulted to:
  - kubelet: 50000-51000
  - webhook: 51001-52000

- #338 
- #347


pkg/controller/cluster/agent/shared.go
```diff
@@ -472,3 +516,50 @@ func newWebhookCerts(commonName string, subAltNames []string, caPrivateKey, caCe
...
+			{
+				Kind:      "ServiceAccount",
+				Name:      s.Name(),
+				Namespace: s.cluster.Namespace,
+			},
+		},
+	}
+
+	return s.ensureObjectWithoutOwner(ctx, roleBinding)
+}
```
enrichman: This ClusterRole is probably not needed. We already have a `k3k-node-proxy` role and binding, used to fetch metrics for the virtual kubelet ([here](https://github.com/rancher/k3k/blob/57263bd10ec355f875e41a2c4b52b96e908fb4a8/k3k-kubelet/provider/configure.go#L119)).

https://github.com/rancher/k3k/blob/57263bd10ec355f875e41a2c4b52b96e908fb4a8/charts/k3k/templates/rbac.yaml#L16-L37

The bindings with the service account of the Kubelet happens [here](https://github.com/rancher/k3k/blob/57263bd10ec355f875e41a2c4b52b96e908fb4a8/pkg/controller/cluster/cluster.go#L605).

Basically we can drop this part, and just adding the missing `watch` verb to the `k3k-node-proxy` ClusterRole.

Maybe we can think about renaming these roles to `k3k-kubelet-node` and `k3k-kubelet-priorityclass`. They are separate because ideally if you don't want the priorityclass to be synced, then the binding with the `k3k-kubelet-priorityclass` can be skipped.

pkg/controller/cluster/agent/agent.go
```diff
@@ -39,15 +39,17 @@ func configSecretName(clusterName string) string {
...
 
 	key := ctrlruntimeclient.ObjectKeyFromObject(obj)
 
 	log.Info(fmt.Sprintf("ensuring %T", obj), "key", key)
 
-	if err := controllerutil.SetControllerReference(cfg.cluster, obj, cfg.scheme); err != nil {
-		return err
+	if setClusterOwner {
+		if err := controllerutil.SetControllerReference(cfg.cluster, obj, cfg.scheme); err != nil {
+			return err
```
enrichman: If we do the ClusterRoleBinding outside with the existing `k3k-kubelet-node` ClusterRole we can revert this part.

pkg/controller/cluster/agent/shared.go
```diff
@@ -64,15 +68,28 @@ func (s *SharedAgent) EnsureResources(ctx context.Context) error {
...
+
 	return nil
 }
 
 func (s *SharedAgent) ensureObject(ctx context.Context, obj ctrlruntimeclient.Object) error {
-	return ensureObject(ctx, s.Config, obj)
+	return ensureObject(ctx, s.Config, obj, true)
+}
+
+func (s *SharedAgent) ensureObjectWithoutOwner(ctx context.Context, obj ctrlruntimeclient.Object) error {
```
enrichman: Same, this can be dropped.

enrichman (changes requested):
Thank you, the PR is working great! We can reuse something that already exists, and tweak some parts.

I.e. I left a couple of comments to reuse the already existing `k3k-node-proxy` ClusterRole. We can rename it to `k3k-kubelet-node` if it sounds better. :smile:

enrichman:
As discussed it should be possible to use the Kubernetes [portallocator](https://github.com/kubernetes/kubernetes/blob/eb1b603cda3b956e52bddf3b51748191e80a59a6/pkg/registry/core/service/portallocator/allocator.go). I did some tests, and even though it's not an exposed resource we can reuse the same concept about having a ConfigMap to store the data:

```go
cm := v1.ConfigMap{
	ObjectMeta: metav1.ObjectMeta{
		Name:      "kubelet-allocator",
		Namespace: "k3k-system",
	},
	Data: map[string]string{
		"range": "50000-60000",
	},
	BinaryData: map[string][]byte{
		"snapshotData": []byte(""),
	},
}

err := s.Config.client.Get(ctx, ctrlruntimeclient.ObjectKeyFromObject(&cm), &cm)
fmt.Println("Get", err)

	err = s.Config.client.Create(ctx, &cm)
	fmt.Println("Create", err)
}

fmt.Println("Data", string(cm.BinaryData["snapshotData"]))

snapshot := core.RangeAllocation{
	Range: cm.Data["range"],
	Data:  cm.BinaryData["snapshotData"],
}

pa, err := portallocator.NewFromSnapshot(&snapshot)
fmt.Println("NewFromSnapshot", pa, err)

next, err := pa.AllocateNext()
fmt.Println("AllocateNext", next, err)
fmt.Println("Used", pa.Used())
fmt.Println("Snapshot", pa.Snapshot(&snapshot))
fmt.Println("Data", string(snapshot.Data))

cm.BinaryData["snapshotData"] = snapshot.Data
cm.Data["range"] = snapshot.Range

err = s.Config.client.Update(ctx, &cm)
fmt.Println("Update", err)
```

k3k-kubelet/kubelet.go
```diff
@@ -272,22 +284,22 @@ func (k *kubelet) start(ctx context.Context) {
...
 
 		return utilProvider, &provider.Node{}, nil
 	}
 }
 
-func (k *kubelet) nodeOpts(ctx context.Context, srvPort, namespace, name, hostname, agentIP string) nodeutil.NodeOpt {
+func (k *kubelet) nodeOpts(ctx context.Context, srvPort int, namespace, name, hostname, agentIP string) nodeutil.NodeOpt {
 	return func(c *nodeutil.NodeConfig) error {
-		c.HTTPListenAddr = fmt.Sprintf(":%s", srvPort)
+		c.HTTPListenAddr = fmt.Sprintf(":%s", strconv.Itoa(srvPort))
```
enrichman:
```suggestion
		c.HTTPListenAddr = fmt.Sprintf(":%d", srvPort)
```

k3k-kubelet/main.go
```diff
@@ -128,7 +139,7 @@ func run(clx *cli.Context) error {
 		logger.Fatalw("failed to create new virtual kubelet instance", zap.Error(err))
 	}
 
-	if err := k.registerNode(ctx, k.agentIP, cfg.KubeletPort, cfg.ClusterNamespace, cfg.ClusterName, cfg.AgentHostname, cfg.ServerIP, k.dnsIP, cfg.Version); err != nil {
+	if err := k.registerNode(ctx, k.agentIP, cfg.KubeletPort, cfg.ClusterNamespace, cfg.ClusterName, cfg.AgentHostname, cfg.ServerIP, k.dnsIP, cfg.Version, cfg.MirrorHostNodes); err != nil {
```
enrichman: Maybe we can just pass the `cfg` instead of adding one more argument (10 are a lot!).

pkg/controller/cluster/agent/ports.go
```diff
@@ -0,0 +1,253 @@
...
+const (
+	kubeletPortRangeConfigMapName = "k3k-kubelet-port-range"
+	webhookPortRangeConfigMapName = "k3k-webhook-port-range"
+)
+
+type PortAllocator struct {
+	KubeletCM        *v1.ConfigMap
+	WebhookCM        *v1.ConfigMap
+	KubeletPortRange string
+	WebhookPortRange string
```
enrichman: I would use the `net.PortRange` from the `k8s.io/apimachinery/pkg/util/net` package, instead of string. This way we can also validate the string passed with the `net.ParsePortRange` method.

k3k-kubelet/controller/webhook/pod.go
```diff
@@ -112,7 +113,7 @@ func (w *webhookHandler) configuration(ctx context.Context, hostClient ctrlrunti
 		return nil, errors.New("webhook CABundle does not exist in secret")
 	}
 
-	webhookURL := "https://" + w.serviceName + ":" + webhookPort + webhookPath
+	webhookURL := "https://" + w.serviceName + ":" + strconv.Itoa(w.webhookPort) + webhookPath
```
enrichman: Maybe we can use a `fmt.Sprintf`?

```suggestion
	webhookURL := fmt.Sprintf("https://%s:%d%s", w.serviceName, w.webhookPort, webhookPath)
```

pkg/controller/cluster/agent/ports.go
enrichman: Thinking a little bit more about this, I think we can omit completely the ranges, because they will be stored in the two configMaps. So I would drop these fields, after the validation.

pkg/controller/cluster/agent/ports.go
```diff
@@ -0,0 +1,253 @@
...
+)
+
+const (
+	kubeletPortRangeConfigMapName = "k3k-kubelet-port-range"
+	webhookPortRangeConfigMapName = "k3k-webhook-port-range"
+)
+
+type PortAllocator struct {
+	KubeletCM        *v1.ConfigMap
+	WebhookCM        *v1.ConfigMap
```
enrichman: What about adding directly the Client in the struct? It's needed for its operations, and I think is ok to embed it.
```suggestion
type PortAllocator struct {
	ctrlruntimeclient.Client
	KubeletCM	*v1.ConfigMap
	WebhookCM	*v1.ConfigMap
```

pkg/controller/cluster/agent/ports.go
```diff
@@ -0,0 +1,253 @@
...
+			return err
+		}
+
+		if err := a.getOrCreate(ctx, client, a.WebhookCM, a.WebhookPortRange); err != nil {
+			return err
+		}
+
+		return nil
+	})
+}
```
enrichman: Which is the reason of having this as a `manager.Runnable`? Can we just initialize it during its creation, in the `NewPortAllocator`?

pkg/controller/cluster/agent/ports.go
```diff
@@ -0,0 +1,253 @@
...
+			configmap = a.cm(configmap.Name, configmap.Namespace, portRange)
+			if err := client.Create(ctx, configmap); err != nil {
+				return fmt.Errorf("failed to create port range configmap: %v", err)
+			}
+		} else {
+			return err
+		}
+	}
+
+	return nil
```
enrichman: Switching the condition we can remove the else (it looks cleaner to me), wdyt?

```suggestion
	if err := client.Get(ctx, nn, configmap); err != nil {
		if !apierrors.IsNotFound(err) {
			return err
		}
		
		// creating the configMap for the first time
		configmap = a.cm(configmap.Name, configmap.Namespace, portRange)
		if err := client.Create(ctx, configmap); err != nil {
			return fmt.Errorf("failed to create port range configmap: %w", err)
		}
	}

	return nil
```

pkg/controller/cluster/agent/ports.go
```diff
@@ -0,0 +1,253 @@
...
+
+	if err := saveSnapshot(pa, &snapshot, configMap, portsMap); err != nil {
+		return nil, err
+	}
+
+	if err := cfg.client.Update(ctx, configMap); err != nil {
+		return nil, err
+	}
+
+	return ptr.To(next), nil
```
enrichman: Is there a reason on returning this as a pointer?

galal-hussein: yes instead of returning 0 or -1, I think thats what kubernetes does as well for its code when dealing with integers for ports

k3k-kubelet/main.go
galal-hussein: yes this was an old code and we kept adding fields to it, I will fix

pkg/controller/cluster/agent/ports.go
galal-hussein: because we need the manager to start first and initialize the client/informers etc, to be able to create the configmap, if you try to use the client before starting the manager you will get an error

pkg/controller/cluster/agent/ports.go
enrichman: Yes, but we are not checking the nil in any case (I guess it cannot be nil, without the err to be nil). Sometimes it's useful to return the three state, but we are anyway de-referencing the pointer. I think it's safer to just return the zero value.

```go
kubeletPort, err = c.PortAllocator.AllocateKubeletPort(ctx, config)
if err != nil {
	return err
}

cluster.Status.KubeletPort = *kubeletPort

webhookPort, err = c.PortAllocator.AllocateWebhookPort(ctx, config)
if err != nil {
	return err
}

cluster.Status.WebhookPort = *webhookPort
```

pkg/controller/cluster/agent/ports.go
```diff
@@ -0,0 +1,257 @@
...
+	portRange, ok := configMap.Data["range"]
+	if !ok {
+		return 0, fmt.Errorf("port range is not initialized")
+	}
+	// get configMap first to avoid conflicts
+	if err := a.getOrCreate(ctx, cfg.client, configMap, portRange); err != nil {
+		return 0, err
+	}
+
+	clusterNameNamespace := cfg.cluster.Name + "-" + cfg.cluster.Namespace
```
enrichman: To avoid collisions:

```suggestion
	clusterNameNamespace := cfg.cluster.Namespace + "/" + cfg.cluster.Name 
```

pkg/controller/cluster/agent/ports.go
```diff
@@ -0,0 +1,257 @@
...
+	portRange, ok := configMap.Data["range"]
+	if !ok {
+		return fmt.Errorf("port range is not initialized")
+	}
+
+	if err := a.getOrCreate(ctx, client, configMap, portRange); err != nil {
+		return err
+	}
+
+	clusterNameNamespace := clusterName + "-" + clusterNamespace
```
enrichman:
```suggestion
	clusterNameNamespace := clusterNamespace + "/" + clusterName
```

Issue #326, opened 2025-04-09
[DOC] Add fs notify disclaimer


jp-gouin:
Sometimes the `kubelet` can crash with a `too many open files` error

It might also happen for the `virtual` mode - **Not Tested** 

We should add in the doc the following process to follow: 

1. Connect to the host cluster
2. Run the following snippet
```
sudo sysctl -w fs.inotify.max_user_watches=2099999999
sudo sysctl -w fs.inotify.max_user_instances=2099999999
sudo sysctl -w fs.inotify.max_queued_events=2099999999
```

This might not work for every Linux distribution so this should be given as example and pointer to the user

jp-gouin:
Relates to #224

PR #390, opened 2025-06-20
add troubleshoot how to guide

jp-gouin:
Add an initial how to troubleshoot virtual clusters doc

Add troubleshooting steps for `too many open files` from #326

### DIFF
--- a/docs/howtos/troubleshoot.md
+++ b/docs/howtos/troubleshoot.md
@@ -1,0 +1,118 @@
+# How to: Troubleshoot K3K
+
+This guide walks through common troubleshooting steps for working with K3K virtual clusters.
+
+---
+
+## Virtual Cluster Fails with “too many open files” Error
+
+### Symptom
+
+The `k3k-kubelet` or `k3kcluster-server-` run into the following issue:  
+
+```sh
+E0604 13:14:53.369369       1 leaderelection.go:336] error initially creating leader election record: Post "https://k3k-http-proxy-k3kcluster-service/apis/coordination.k8s.io/v1/namespaces/kube-system/leases": context canceled
+{"level":"fatal","timestamp":"2025-06-04T13:14:53.369Z","logger":"k3k-kubelet","msg":"virtual manager stopped","error":"too many open files"}
+```
+
+This typically indicates a low limit on inotify watchers or file descriptors on the host system.
+
+### Solution: Adjust Host System Settings
+
+Connect to the host nodes and increase inotify limits:
+
+```sh
+sudo sysctl -w fs.inotify.max_user_watches=2099999999
+sudo sysctl -w fs.inotify.max_user_instances=2099999999
+sudo sysctl -w fs.inotify.max_queued_events=2099999999
+```
+
+You can persist these settings by adding them to `/etc/sysctl.conf`:
+
+```sh
+fs.inotify.max_user_watches=2099999999
+fs.inotify.max_user_instances=2099999999
+fs.inotify.max_queued_events=2099999999
+```
+
+Apply the changes:
+
+```sh
+sudo sysctl -p
+```
+
+You can find more details in this [KB document](https://www.suse.com/support/kb/doc/?id=000020048).
+
+---
+
+## Inspect Controller Logs for Failure Diagnosis
+
+To view logs for a failed virtual cluster:
+
+```sh
+kubectl logs -n k3k-system -l app.kubernetes.io/name=k3k
+```
+
+This retrieves logs from K3k controller components.
+
+---
+
+## Inspect Cluster Logs for Failure Diagnosis
+
+To view logs for a failed virtual cluster:
+
+```sh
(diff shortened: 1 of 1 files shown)


docs/howtos/troubleshoot.md
```diff
@@ -0,0 +1,118 @@
+# How to: Troubleshoot K3K
```
enrichman: What about a generic "Troubleshooting" title?

```suggestion
# Troubleshooting
```

docs/howtos/troubleshoot.md
```diff
@@ -0,0 +1,118 @@
...
+
+This guide walks through common troubleshooting steps for working with K3K virtual clusters.
+
+---
+
+## Virtual Cluster Fails with “too many open files” Error
```
enrichman: I would be more direct:

```suggestion
## `too many open files` error
```

docs/howtos/troubleshoot.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Symptom
```
enrichman: I would drop this, it looks a little bit "medical" :smile: 

```suggestion
```

docs/howtos/troubleshoot.md
```diff
@@ -0,0 +1,118 @@
...
+```sh
+E0604 13:14:53.369369       1 leaderelection.go:336] error initially creating leader election record: Post "https://k3k-http-proxy-k3kcluster-service/apis/coordination.k8s.io/v1/namespaces/kube-system/leases": context canceled
+{"level":"fatal","timestamp":"2025-06-04T13:14:53.369Z","logger":"k3k-kubelet","msg":"virtual manager stopped","error":"too many open files"}
+```
+
+This typically indicates a low limit on inotify watchers or file descriptors on the host system.
+
+### Solution: Adjust Host System Settings
+
+Connect to the host nodes and increase inotify limits:
```
enrichman: For the same reason I would keep everything in the same section, and have a more linear reading, with something like:
```suggestion
To increase the inotify limits connect to the host nodes and run:
```

Wdyt?

docs/howtos/troubleshoot.md
```diff
@@ -0,0 +1,118 @@
...
+This retrieves logs from K3k controller components.
+
+---
+
+## Inspect Cluster Logs for Failure Diagnosis
+
+To view logs for a failed virtual cluster:
+
+```sh
+kubectl logs -n k3k-<cluster_name> -l cluster=<cluster_name>
```
enrichman: It's probably better to have `<cluster_namespace>` instead of `k3k-<cluster_name>`, because a virtual cluster can be deployed also in a dedicated non-k3k-prefixed namespace

```suggestion
kubectl logs -n <cluster_namespace> -l cluster=<cluster_name>
```

docs/howtos/troubleshoot.md
```diff
@@ -0,0 +1,118 @@
...
+Check the associated pods in the K3K system namespace:
+
+```sh
+kubectl get pods -n <cluster_namespace>
+kubectl describe pod <pod-name> -n k3k-system
+```
+
+Look for events like `FailedScheduling`, `ImagePullBackOff`, or `VolumeBindingFailed`.
+
+---
```
enrichman: This is probably one of the most important section, so we should document it a bit more. I.e. have dedicated sub-sections for every possible issue, and explain how to check if that's the root cause.

Something like:

---

## Virtual Cluster Not Starting or Stuck in Pending

Some of the most common causes are related to missing prerequisites or wrong configuration.

### Storage class not available

When creating a Virtual Cluster in `shared` mode (the default), a PVC is needed. You can check if the PVC was claimed but not bound with `kubectl get pvc -n <cluster_namespace>`. If you see a pending PVC you probably don;t have a default storage class defined, or you have specified a wrong one. To fix.. BLABLABLA

### Insufficient node resources

...etc...

enrichman (commented):
Thanks @jp-gouin , I left few comments!

docs/howtos/troubleshoot.md
jp-gouin: I agree and that is also something that we should build iteratively and probably not in one go.

docs/howtos/troubleshoot.md
```diff
@@ -0,0 +1,147 @@
...
+The `server` is failing:
+
+```bash
+kubectl get po -n k3k-test-registry
+NAME                             READY   STATUS          RESTARTS       AGE
+k3k-test-registry-kubelet-r4zh5   1/1     Running            0          54s
+k3k-test-registry-server-0        0/1     ImagePullBackOff   0          54s
+```
+
+To fix this make sure the failing image is available. you can describe the failing pod to get more details.
```
enrichman:
```suggestion
To fix this make sure the failing image is available. You can describe the failing pod to get more details.
```

enrichman (approved):

PR #396, opened 2025-06-25
Add scheduled Conformance tests for shared mode

### LINKED ISSUE #331: Conformance tests
Add conformance test for 
- virtual cluster in `shared` mode
- virtual cluster in `virtual` mode

Conformance test should be run (at each release or another cadence) - TBD 

Result of the test should be store in the doc directory 

(Later) - if needed create another issue
Run conformance test with a host cluster with Cilium

enrichman:
This PR introduces a new nightly (01:00) job to execute Conformance tests in "shared" mode.

The job will perform the following:
- Run the full Conformance test suite in parallel, skipping `[Serial]` tests.
- Subsequently, run `[Serial].*[Conformance]` tests serially.
- Concurrently, execute Conformance tests categorized by SIGs.

These tests can also be triggered manually with a specified execution type.

For more information, please refer to the [K8s Conformance documentation](https://github.com/cncf/k8s-conformance/tree/master/docs).

Ref: #166 #331

### DIFF
--- a/.github/workflows/test-conformance.yaml
+++ b/.github/workflows/test-conformance.yaml
@@ -28,13 +28,143 @@ permissions:
 jobs:
   conformance:
     runs-on: ubuntu-latest
+    if: inputs.test == '' || inputs.test == 'conformance'
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        fetch-tags: true
+
+    - uses: actions/setup-go@v5
+      with:
+        go-version-file: go.mod
+
+    - name: Install helm
+      uses: azure/setup-helm@v4.3.0
+    
+    - name: Install hydrophone
+      run: go install sigs.k8s.io/hydrophone@latest
+
+    - name: Install k3d and kubectl
+      run: |
+        wget -q -O - https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
+        k3d version
+
+        curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+
+    - name: Setup Kubernetes (k3d)
+      env:
+        REPO_NAME: k3k-registry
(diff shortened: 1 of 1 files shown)


.github/workflows/test-conformance.yaml
```diff
@@ -0,0 +1,296 @@
...
+    - name: Archive conformance logs
+      uses: actions/upload-artifact@v4
+      if: always()
+      with:
+        name: conformance-${{ matrix.type }}-logs
+        path: /tmp/e2e.log
+
+  sigs:
+    runs-on: ubuntu-latest
+    if: inputs.test == '' || inputs.test != 'conformance'
```
galal-hussein, line 159: Should we have input.test == '' ?

.github/workflows/test-conformance.yaml
```diff
@@ -0,0 +1,296 @@
...
+        k3kcli kubeconfig generate --name mycluster
+
+        export KUBECONFIG=${{ github.workspace }}/k3k-mycluster-mycluster-kubeconfig.yaml
+        
+        kubectl cluster-info
+        kubectl get nodes
+        kubectl get pods -A
+
+    - name: Run sigs tests
+      if: inputs.test == '' || inputs.test == matrix.tests.name
```
galal-hussein, line 282: same question

galal-hussein (approved):
LGTM

.github/workflows/test-conformance.yaml
enrichman: If the question is about `input` or `inputs` (plural) the object should be [`inputs`](https://docs.github.com/en/actions/reference/metadata-syntax-for-github-actions#inputsinput_id).

If otherwise the question was about dropping the second part (` || inputs.test != 'conformance'`) that will be needed when running manually the workflow.

Basically for the scheduled run the inputs will be blank, so all the tests will have to run. Otherwise, when the inputs. test is not blank that means a manual selection was made, and so just the selected test should run.

Unfortunately this cannot be tested until the PR is merged, and available on main.

PR #397, opened 2025-06-26
Increase maxConcurrentReconciles to 50

manno:


enrichman:
Hi @manno, I've just opened https://github.com/rancher/k3k/pull/410 to default the value to `50`, but also adding this as a configurable option. :+1:

Issue #406, opened 2025-07-03
Bug with `expose: {}`


jp-gouin:
**Environmental Info:**
Host Cluster Version: 
RKE2 v1.31.5+rke2r1

Host Cluster Configuration:
Single node cluster

K3K Cluster Configuration:
```
apiVersion: k3k.io/v1alpha1
kind: Cluster
metadata:
  name: test0
  namespace: k3k-test0
spec:
  agents: 0
  mode: shared
  expose: {}
  persistence:
    storageClassName: local-path
    storageRequestSize: 5Gi
    type: dynamic
  servers: 1
  version: v1.32.5-k3s1
```
**Describe the bug:**
With `expose: {}` the resource is correctly applied but the cluster fails to provision. 
the controller is assigning the wrong kind of service. 

**Expected behavior:**
It should default to the service type `ClusterIP`

aganesh-suse:
@jp-gouin @enrichman Pls let know if anything else needs to be tested here. Else we can close the issue as verified. Thanks.

PR #408, opened 2025-07-08
Add Conditions and current status to Cluster

enrichman:
This PR adds a generic current status of the Cluster, and the Conditions to monitor the status of it. It also uses the EventRecorder to emit events, useful to debug possible problems.

```go
// Conditions are the individual conditions for the cluster set.
//
// +optional
Conditions []metav1.Condition `json:"conditions,omitempty"`

// Phase is a high-level summary of the cluster's current lifecycle state.
//
// +kubebuilder:default="Unknown"
// +kubebuilder:validation:Enum=Pending;Provisioning;Ready;Failed;Terminating;Unknown
// +optional
Phase ClusterPhase `json:"phase,omitempty"`
```

This will be reflected in the output of the `k3kcli` and also from kubectl:

```
-> % k3kcli cluster list                                                              
NAMESPACE       NAME             MODE     STATUS   POLICY
k3k-mycluster   mycluster        shared   Failed   <none>
shared          sharedcluster2   shared   Ready    <none>

-> % k get clusters.k3k.io -A
NAMESPACE       NAME             MODE     STATUS   POLICY
k3k-mycluster   mycluster        shared   Failed   
shared          sharedcluster2   shared   Ready  
```

The emitted Events:
```
Events:
  Type     Reason              Age                 From                    Message
  ----     ------              ----                ----                    -------
  Warning  ProvisioningFailed  10s (x11 over 32s)  k3k-cluster-controller  Ingress.networking.k8s.io "k3k-mycluster-ingress" is invalid: spec: Invalid value: []networking.IngressRule(nil): either `defaultBackend` or `rules` must be specified
```

After the reconciliation an `updateStatus` was added, that will check the error, and eventually update the status and conditions in the proper way:

```go
func (c *ClusterReconciler) reconcileCluster(ctx context.Context, cluster *v1alpha1.Cluster) error {
	err := c.reconcile(ctx, cluster)
	c.updateStatus(cluster, err)

	return err
}
```

I.e. now the `validate` is checking if the Cluster is violating the Policy, and returning an `ErrClusterValidation`, and the cluster will be stuck in the `pending` state.

### Extra

The `PortAllocator` instantiation was moved outside to reduce the number of needed args, but we'll need probably a Config struct, or some Options.

It also fixes https://github.com/rancher/k3k/issues/406


galal-hussein (approved):
LGTM, this is great thanks!

Issue #414, opened 2025-07-17
CLI is not respecting the `--kubeconfig` flag


enrichman:
It looks like the `--kubeconfig` flag is not taken into account:

```
-> % export KUBECONFIG=/my/kube/config
-> % k3kcli cluster list                                                                                 
NAMESPACE         NAME             MODE     STATUS   POLICY
k3k-mycluster     mycluster        shared   Failed   <none>
k3k-nonexisting   nonexisting      shared   Failed   nonexisting
k3k-nonexisting   nonexists2       shared   Failed   nonexisting
k3k-test          test             shared   Ready    <none>
shared            sharedcluster2   shared   Ready    <none>

-> % unset KUBECONFIG   
-> % k3kcli cluster list --kubeconfig /my/kube/config
no configuration has been provided, try setting KUBERNETES_MASTER environment variable

```

xelab04:
I found that the kubeconfig flag is respected if it's before the command being run.

So, for me, this works `k3kcli --kubeconfig=/my/kube/config cluster list`

I traced this behaviour to root.go:

```
restConfig, err := loadRESTConfig(appCtx.Kubeconfig)
```
Urfave CLI hasn't parsed the subcommand flags yet (when --kubeconfig is after the command) so appCtx.Kubeconfig is undefined.

PR #422, opened 2025-07-19
Fix kubeconfig flag being recognised

xelab04:
This is a fix for #414 
It works by running the app.Before, then Before from cluster_list.go, then runs initialiseClient.

The issue stems from the kubeconfig flag being at the end of the command and thus not having been parsed yet at the app.Before stage. Hence `k3kcli cluster list --kubeconfig=/kubeconfig` failed while `k3kcli --kubeconfig=/kubeconfig cluster list` worked.

The fix is to only initialise the client *after* the kubeconfig flag has been read in the former case.

### DIFF
--- a/cli/cmds/cluster_list.go
+++ b/cli/cmds/cluster_list.go
@@ -12,8 +12,17 @@ import (
 )
 
 func NewClusterListCmd(appCtx *AppContext) *cli.Command {
-	flags := CommonFlags(appCtx)
-	flags = append(flags, FlagNamespace(appCtx))
+	// flags := CommonFlags(appCtx)
+	// flags = append(flags, FlagNamespace(appCtx))
+
+	flags := []cli.Flag{
+		FlagNamespace(appCtx),
+		// prevents overwriting of kubeconfig specified globally
+		&cli.StringFlag{
+			Name:  "kubeconfig",
+			Usage: "kubeconfig path",
+		},
+	}
 
 	return &cli.Command{
 		Name:            "list",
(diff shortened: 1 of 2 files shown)


cli/cmds/cluster_list.go
```diff
@@ -22,6 +22,10 @@ func NewClusterListCmd(appCtx *AppContext) *cli.Command {
 		Action:          list(appCtx),
 		Flags:           flags,
 		HideHelpCommand: true,
+		Before: func(clx *cli.Context) error {
+	        appCtx.Kubeconfig = clx.String("kubeconfig")
+	        return initializeClient(appCtx)
+	    },
```
xelab04, line 40: This part will handle the kubeconfig file if it's passed in the `k3kcli cluster list --kubeconfig=path` format.

cli/cmds/cluster_list.go
```diff
@@ -12,8 +12,17 @@ import (
 )
 
 func NewClusterListCmd(appCtx *AppContext) *cli.Command {
-	flags := CommonFlags(appCtx)
-	flags = append(flags, FlagNamespace(appCtx))
+	// flags := CommonFlags(appCtx)
```
xelab04, line 15: If left, it would interfere with the `k3kcli --kubeconfig=/path cluster list` format as it would not find a value for kubeconfig and thus reset it to empty.

xelab04:
Awesome! I'm looking forward to it (and helping out if I can) !

PR #426, opened 2025-07-22
Migrate from `urfave/cli` to `cobra`

enrichman:
This PR migrates from `urfave/cli` to `cobra`.

The functionalities are the same, so no breaking changes are expected.

Cobra supports GLOBAL OPTIONS, so now it's possible to use the `--debug` and `--kubeconfig` flag in every command.

Some other changes are the one in the `k3k-kubelet`:
- Renamed `--host-config-path` to `--host-kubeconfig`
- Renamed `--virtual-config-path` to `--virt-kubeconfig`
- Renames `--cluster-token` to `--token`
- Changed default of `--config` from `"/etc/rancher/k3k/config.yaml"` to `"/opt/rancher/k3k/config.yaml"` (and consequently removed the explicit `--config` from the args of the kubelet.

The old `--virtual-config-path` had also a wrong env variable (`CLUSTER_NAME`), and all the other env variables are now matching the flag name.

This should be merged after #417, to check that the main commands are working.

## Acceptance Tests

To check the migration is successful we should test at least some of the basic features:

#### Create

Check that a `k3kcli cluster create` behaves normally, trying different flags, like the `--mode`, `--persistence-type`, `--namespace`.

#### Delete

Check that `k3kcli cluster delete` deletes the cluster.

#### List

Check that `k3kcli cluster list` lists the existing clusters.

#### Create

Check that a `k3kcli policy create` behaves normally, creating a policy.

#### Delete

Check that `k3kcli policy delete` deletes the policy.

#### List

Check that `k3kcli policy list` lists the existing policies.

### Kubeconfig

Check that the `k3kcli kubeconfig generate` is able to generate a valid kubeconfig for the target cluster.


k3k-kubelet/config.go
```diff
@@ -2,83 +2,22 @@ package main
...
-	if c.Version == "" {
-		c.Version = conf.Version
-	}
-
-	if !c.MirrorHostNodes {
-		c.MirrorHostNodes = conf.MirrorHostNodes
-	}
-
-	return nil
+	ClusterName      string `mapstructure:"clustername"`
```
galal-hussein: looks like yaml config are not marshalled correctly

galal-hussein (changes requested):
Only small comment on how we handle k3k-kubelet config yaml, otherwise the PR looks really good thanks!

k3k-kubelet/config.go
enrichman: Ah, the `mapstructure` is used to map viper keys to the yaml, but it's case insensitive. I anyway switched to the camelCase because I like it more, and it's more clear!

PR #431, opened 2025-07-28
Add resources sync configuration

galal-hussein:
The PR adds the following to the cluster spec:

```
    sync:
      configmaps:
        activeResources: true
        enabled: true
      ingresses:
        enabled: false
      persistentVolumeClaims:
        enabled: true
      priorityClasses:
        enabled: true
      secrets:
        activeResources: true
        enabled: true
      services:
        enabled: true
```

Currently there are 6 resource types that are being synchronized from virtual cluster to the host cluster in shared mode, the defaults are:

1- configmaps: only active resources are being synced, i.e the configmaps used by pods as a volume or a projected volume, to enable syncing all configmaps `activeResources` must be set to false and `enabled` must be set to true
2- secrets: only active resources are being synced, i.e the secrets used by pods as a volume or a projected volume, to enable syncing all secrets `activeResources` must be set to false and `enabled` must be set to true
3- pvcs: enabled by default, and all pvcs will be synced to the host cluster.
4- services: enabled by default, and all services will be synced to the host cluster.
5- priorityClasses: disabled by default.
6- ingresses: disabled by default.


k3k-kubelet/controller/configmap.go
```diff
@@ -31,18 +38,56 @@ type ConfigMapSyncer struct {
 	TranslateFunc func(*corev1.ConfigMap) (*corev1.ConfigMap, error)
 	// Logger is the logger that the controller will use
 	Logger *k3klog.Logger
+	// Scheme is the scheme of the virtual manager
+	Scheme *runtime.Scheme
```
enrichman: I would use the same `Virtual` prefix, for clarity.
```go
// VirtualScheme is the scheme of the virtual manager
VirtualScheme *runtime.Scheme
```

k3k-kubelet/controller/configmap.go
```diff
@@ -31,18 +38,56 @@ type ConfigMapSyncer struct {
...
+	if labelSelector.Empty() {
+		labelSelector = labels.Everything()
+	}
+
+	return ctrl.NewControllerManagedBy(virtMgr).
+		Named(ConfigMapSyncerName).
+		For(&corev1.ConfigMap{}).WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
+		return labelSelector.Matches(labels.Set(object.GetLabels()))
+	})).
+		Complete(&reconciler)
```
enrichman: Formatting nit:

```suggestion
	return ctrl.NewControllerManagedBy(virtMgr).
		Named(ConfigMapSyncerName).
		For(&corev1.ConfigMap{}).
		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
			return labelSelector.Matches(labels.Set(object.GetLabels()))
		})).
		Complete(&reconciler)
```

k3k-kubelet/controller/ingress.go
```diff
@@ -0,0 +1,145 @@
...
+	ingressFinalizerName    = "ingress.k3k.io/finalizer"
+)
+
+type IngressReconciler struct {
+	clusterName      string
+	clusterNamespace string
+
+	virtualClient ctrlruntimeclient.Client
+	hostClient    ctrlruntimeclient.Client
+	Scheme        *runtime.Scheme
```
enrichman: I would use also here the `Virtual` prefix. Also these are unexported, while the others in the other syncer are exported. If they are not used externally we can probably unexport them.

k3k-kubelet/controller/ingress.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	return ctrl.NewControllerManagedBy(virtMgr).
+		Named(ingressSyncerController).
+		For(&networkingv1.Ingress{}).
+		Complete(&reconciler)
+}
+
+func (r *IngressReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := ctrl.LoggerFrom(ctx).WithValues("cluster", r.clusterName, "clusterNamespace", r.clusterNamespace, "name", req.Name, "namespace", req.Namespace)
```
enrichman: The `name` and `namespace` are already in the logger context
```suggestion
	log := ctrl.LoggerFrom(ctx).WithValues("cluster", r.clusterName, "clusterNamespace", r.clusterNamespace)
```

k3k-kubelet/controller/ingress.go
```diff
@@ -0,0 +1,145 @@
...
+		cluster     v1alpha1.Cluster
+	)
+
+	if err := r.hostClient.Get(ctx, types.NamespacedName{Name: r.clusterName, Namespace: r.clusterNamespace}, &cluster); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if err := r.virtualClient.Get(ctx, req.NamespacedName, &virtIngress); err != nil {
+		return reconcile.Result{}, ctrlruntimeclient.IgnoreNotFound(err)
+	}
```
enrichman: Commenting here, but I guess the logic applies to all the syncers.

What if someone deletes the resource in theHost cluster? The `r.hostClient.Get` will fail, and it will not be recreated. I think we should handle this case, checking before the resource in the Virtual Cluster, and if deleted then deletes also the one in the Host, and if not syncronizing it.

k3k-kubelet/controller/ingress.go
```diff
@@ -0,0 +1,145 @@
...
+	if err := controllerutil.SetControllerReference(&cluster, syncedIngress, r.HostScheme); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// handle deletion
+	if !virtIngress.DeletionTimestamp.IsZero() {
+		// deleting the synced service if exists
+		if err := r.hostClient.Delete(ctx, syncedIngress); err != nil {
+			return reconcile.Result{}, ctrlruntimeclient.IgnoreNotFound(err)
+		}
```
enrichman: IIRC we have a bug because of this check.

If we delete an Ingress but the Ingress on the Host was already deleted, or not found, then we will fall into this case. But we will return a `return reconcile.Result{}, nil`, because of the `IgnoreNotFound`. That means the reconciliation loop will NOT be triggered again, and the Virtual Ingress will not be deleted, and stuck into the deletion phase.

k3k-kubelet/controller/ingress.go
```diff
@@ -0,0 +1,145 @@
...
+
+		// remove the finalizer after cleaning up the synced service
+		if controllerutil.ContainsFinalizer(&virtIngress, ingressFinalizerName) {
+			controllerutil.RemoveFinalizer(&virtIngress, ingressFinalizerName)
+
+			if err := r.virtualClient.Update(ctx, &virtIngress); err != nil {
+				return reconcile.Result{}, err
+			}
+		}
```
enrichman: Commenting here, but it's the same for the other places:

the `controllerutil.RemoveFinalizer` is already checking for the existence of the finalizer, returning a `bool` if it's being removed.

k3k-kubelet/controller/configmap.go
```diff
@@ -31,18 +38,56 @@ type ConfigMapSyncer struct {
...
+
+	reconciler := ConfigMapSyncer{
+		VirtualClient: virtMgr.GetClient(),
+		HostClient:    hostMgr.GetClient(),
+		Scheme:        virtMgr.GetScheme(),
+		HostScheme:    hostMgr.GetScheme(),
+		TranslateFunc: func(cm *corev1.ConfigMap) (*corev1.ConfigMap, error) {
+			translator.TranslateTo(cm)
+			return cm, nil
+		},
```
enrichman: The `TranslateFunc` looks weird. It's proably better to have just the `Translator` as a field, and use that instead. Like we already have in other syncers.

galal-hussein: The idea of the TranslateFunc is Michael's idea, he wanted to eventually create a generic syncer that applies to all resources, and just change the translate func

k3k-kubelet/controller/syncer/configmap.go
```diff
@@ -25,17 +25,16 @@ import (
 )
 
 const (
-	configMapControllerName = "configmap-syncer"
-	configMapFinalizerName  = "configmap.k3k.io/finalizer"
+	configMapControllerName       = "configmap-syncer"
+	globalConfigMapCongrollerName = "global-configmap-syncer"
```
enrichman: Typo

```suggestion
	globalConfigMapControllerName = "global-configmap-syncer"
```

cli/cmds/cluster_create.go
```diff
@@ -196,6 +196,17 @@ func newCluster(name, namespace string, config *CreateConfig) *v1alpha1.Cluster
...
+				ConfigMaps: v1alpha1.SyncConfigOptions{
+					Enabled: true,
+				},
+				Secrets: v1alpha1.SyncConfigOptions{
+					Enabled: true,
+				},
+				Services: v1alpha1.SyncConfigOptions{
+					Enabled: true,
+				},
+			},
```
enrichman: I would drop this, and leverage the default values.

galal-hussein: yeah for some reason, the cli wont respect the default values

k3k-kubelet/controller/syncer/configmap.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// check for configMap Sync Config
+	syncConfig := cluster.Spec.Sync.ConfigMaps
+
+	if !syncConfig.Enabled {
+		if object.GetDeletionTimestamp() != nil {
+			return true
+		}
+		return false
+	}
```
enrichman: The linter is suggesting this, and I would also add a small comment:

```suggestion
	// If syncing is disabled, only process deletions to allow for cleanup.
	if !syncConfig.Enabled {
		return object.GetDeletionTimestamp() != nil
	}
```

k3k-kubelet/controller/syncer/ingress.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// check for ingressConfig
+	syncConfig := cluster.Spec.Sync.Ingresses
+
+	if !syncConfig.Enabled {
+		if object.GetDeletionTimestamp() != nil {
+			return true
+		}
+		return false
+	}
```
enrichman: Same:

```suggestion
	// If syncing is disabled, only process deletions to allow for cleanup.
	if !syncConfig.Enabled {
		return object.GetDeletionTimestamp() != nil
	}
```

k3k-kubelet/controller/syncer/persistentvolumeclaims.go
```diff
@@ -19,77 +20,96 @@ import (
...
+
+	// check for pvc config
+	syncConfig := cluster.Spec.Sync.PersistentVolumeClaims
+
+	if !syncConfig.Enabled {
+		if object.GetDeletionTimestamp() != nil {
+			return true
+		}
+		return false
+	}
```
enrichman: Same:

```suggestion
	// If syncing is disabled, only process deletions to allow for cleanup.
	if !syncConfig.Enabled {
		return object.GetDeletionTimestamp() != nil
	}
```

k3k-kubelet/controller/syncer/priorityclass.go
```diff
@@ -83,20 +73,47 @@ var ignoreSystemPrefixPredicate = predicate.Funcs{
...
+
+	// check for priorityClassConfig
+	syncConfig := cluster.Spec.Sync.PriorityClasses
+
+	if !syncConfig.Enabled {
+		if object.GetDeletionTimestamp() != nil {
+			return true
+		}
+		return false
+	}
```
enrichman: Same:

```suggestion
	// If syncing is disabled, only process deletions to allow for cleanup.
	if !syncConfig.Enabled {
		return object.GetDeletionTimestamp() != nil
	}
```

k3k-kubelet/controller/syncer/secret.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// check for Secrets Sync Config
+	syncConfig := cluster.Spec.Sync.Secrets
+
+	if !syncConfig.Enabled {
+		if object.GetDeletionTimestamp() != nil {
+			return true
+		}
+		return false
+	}
```
enrichman: Same:

```suggestion
	// If syncing is disabled, only process deletions to allow for cleanup.
	if !syncConfig.Enabled {
		return object.GetDeletionTimestamp() != nil
	}
```

k3k-kubelet/controller/syncer/service.go
```diff
@@ -104,28 +97,53 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req reconcile.Request
...
+
+	// check for serviceSyncConfig
+	syncConfig := cluster.Spec.Sync.Services
+
+	if !syncConfig.Enabled {
+		if object.GetDeletionTimestamp() != nil {
+			return true
+		}
+		return false
+	}
```
enrichman: Same:

```suggestion
	// If syncing is disabled, only process deletions to allow for cleanup.
	if !syncConfig.Enabled {
		return object.GetDeletionTimestamp() != nil
	}
```

k3k-kubelet/controller/syncer/pod.go
```diff
@@ -19,60 +19,71 @@ import (
...
+	// check for pvc config
+	syncConfig := cluster.Spec.Sync.PersistentVolumeClaims
+
+	if !syncConfig.Enabled {
+		if object.GetDeletionTimestamp() != nil {
+			return true
+		}
+		return false
+	}
+	return true
```
enrichman:
```suggestion
	// If PVC syncing is disabled, only process deletions to allow for cleanup.
	return syncConfig.Enabled || object.GetDeletionTimestamp() != nil
```

k3k-kubelet/controller/syncer/configmap.go
```diff
@@ -0,0 +1,152 @@
...
+
+	labelSelector := labels.SelectorFromSet(syncConfig.Selector)
+	if labelSelector.Empty() {
+		labelSelector = labels.Everything()
+	}
```
enrichman: If the labelSelector is empty we can just return true:

```suggestion
	if labelSelector.Empty() {
		return true
	}
```

k3k-kubelet/controller/syncer/ingress.go
```diff
@@ -0,0 +1,163 @@
...
+
+	labelSelector := labels.SelectorFromSet(syncConfig.Selector)
+	if labelSelector.Empty() {
+		labelSelector = labels.Everything()
+	}
```
enrichman: Same:
```suggestion
	if labelSelector.Empty() {
		return true
	}
```

k3k-kubelet/controller/syncer/persistentvolumeclaims.go
```diff
@@ -19,77 +20,96 @@ import (
...
+
+	labelSelector := labels.SelectorFromSet(syncConfig.Selector)
+	if labelSelector.Empty() {
+		labelSelector = labels.Everything()
+	}
```
enrichman: Same:
```suggestion
	if labelSelector.Empty() {
		return true
	}
```

k3k-kubelet/controller/syncer/priorityclass.go
```diff
@@ -83,20 +73,47 @@ var ignoreSystemPrefixPredicate = predicate.Funcs{
...
+
+	labelSelector := labels.SelectorFromSet(syncConfig.Selector)
+	if labelSelector.Empty() {
+		labelSelector = labels.Everything()
+	}
```
enrichman: Same:
```suggestion
	if labelSelector.Empty() {
		return true
	}
```

k3k-kubelet/controller/syncer/secret.go
```diff
@@ -0,0 +1,157 @@
...
+
+	labelSelector := labels.SelectorFromSet(syncConfig.Selector)
+	if labelSelector.Empty() {
+		labelSelector = labels.Everything()
+	}
```
enrichman: Same:
```suggestion
	if labelSelector.Empty() {
		return true
	}
```

k3k-kubelet/controller/syncer/service.go
```diff
@@ -104,28 +97,53 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req reconcile.Request
...
+
+	labelSelector := labels.SelectorFromSet(syncConfig.Selector)
+	if labelSelector.Empty() {
+		labelSelector = labels.Everything()
+	}
```
enrichman: Same:
```suggestion
	if labelSelector.Empty() {
		return true
	}
```

cli/cmds/cluster_create.go
enrichman: Hm, that's strange. Maybe it's because of some `omitempty` or default kubebuilder tags? Does it work sending the yaml with an empty sync?

enrichman (approved):
Thanks, LGTM!

fykyx521:
why virutal mode no support ?

enrichman:
> why virutal mode no support ?

Hi @fykyx521, the resource sync feature is specific to shared mode because of how it integrates with the host cluster.

In shared mode, the virtual cluster shares some resources directly with the host cluster. The "sync" mechanism is necessary to manage and reconcile these shared components.

In virtual mode, the entire k3s cluster runs self-contained inside pods on the host. It doesn't share the host's resources in the same way, providing complete isolation. Since there are no resources being shared from the host, a sync mechanism isn't needed.

So, the lack of support in virtual mode is by design, as it's a fully isolated environment.

Issue #345, opened 2025-05-07
Pod stuck in CreateContainerConfigError after updating subpathexpr


galal-hussein:
**Environmental Info:**
Host Cluster Version: 

v1.32.3+k3s1
Node(s) CPU architecture, OS, and Version: 

1 node ubuntu 24.04

Host Cluster Configuration:

1 server

K3K Cluster Configuration:

1 server shared mode

**Describe the bug:**

In shared mode when creating a pod with a missing subpathexpr, and then updating it, it will still be stuck in `CreateContainerConfigError`

**Steps To Reproduce:**
- Created a cluster with `k3k create`: 
```
k3kcli cluster create --namespace shared --mode shared --servers 1 sharedcluster
```
- create a a pod with subpathExpr
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    notmysubpath: mypath
  labels:
    name: var-expansion-test
  name: var-expansion-test
  namespace: default
spec:
  containers:
  - command:
    - sh
    - -c
    - tail -f /dev/null
    env:
    - name: POD_NAME
      value: foo
    - name: ANNOTATION
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.annotations['mysubpath']
    image: registry.k8s.io/e2e-test-images/busybox:1.36.1-1
    imagePullPolicy: IfNotPresent
    name: dapi-container
    volumeMounts:
    - mountPath: /subpath_mount
      name: workdir1
      subPathExpr: $(ANNOTATION)/$(POD_NAME)
    - mountPath: /volume_mount
      name: workdir1
  volumes:
  - emptyDir: {}
    name: workdir1
```

- After this pod is created it will be in createcontainerconfigerror as **expected** however after editing the pod and adding annotation
```
mysubpath: mypath
```
it will still be in error

**Expected behavior:**

The pod should be updated successfully

**Actual behavior:**

Failure

**Additional context / logs:**

PR #434, opened 2025-08-08
Fix pod fieldpath annotation translation

galal-hussein:
This looks like a a bug in virtual-kubelet, an example for this problem:

1- create a pod in virtual cluster with the following yaml:

```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    notmysubpath: mypath
  labels:
    name: var-expansion-test
  name: var-expansion-test
  namespace: default
spec:
  containers:
  - command:
    - sh
    - -c
    - tail -f /dev/null
    env:
    - name: POD_NAME
      value: foo
    - name: ANNOTATION
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.annotations['mysubpath']
    image: registry.k8s.io/e2e-test-images/busybox:1.36.1-1
    imagePullPolicy: IfNotPresent
    name: dapi-container
    volumeMounts:
    - mountPath: /subpath_mount
      name: workdir1
      subPathExpr: $(ANNOTATION)/$(POD_NAME)
    - mountPath: /volume_mount
      name: workdir1
  volumes:
  - emptyDir: {}
    name: workdir1
```
when printing the Pod in createPod() function you can see the following:

```
debug: target pod before configure fieldpath env: []v1.EnvVar{....v1.EnvVar{Name:"ANNOTATION", Value:"", ValueFrom:(*v1.EnvVarSource)(nil)}, ....
```

to workaround this problem for now this PR will fetch the pod from the virtual cluster before attempting the sync

#345

### DIFF
--- a/k3k-kubelet/provider/provider.go
+++ b/k3k-kubelet/provider/provider.go
@@ -340,7 +340,14 @@ func (p *Provider) CreatePod(ctx context.Context, pod *corev1.Pod) error {
 
 // createPod takes a Kubernetes Pod and deploys it within the provider.
 func (p *Provider) createPod(ctx context.Context, pod *corev1.Pod) error {
-	tPod := pod.DeepCopy()
+	// fieldPath envs are not being translated correctly using the virtual kubelet pod controller
+	// as a workaround we will try to fetch the pod from the virtual cluster and copy over the envSource
+	var sourcePod corev1.Pod
+	if err := p.VirtualClient.Get(ctx, types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, &sourcePod); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


dereknola (approved):

enrichman:
Are these related?
- https://github.com/virtual-kubelet/virtual-kubelet/pull/999
- https://github.com/virtual-kubelet/virtual-kubelet/issues/1032
- https://github.com/virtual-kubelet/virtual-kubelet/issues/998

I would track the issue in the comment if yes, or I would open a new issue in the upstream project if not. We should/could contribute on this, if possible.

The workaround is ok btw. Currently I don't see any other option.

enrichman (approved):

PR #441, opened 2025-08-19
Fix subpath field

galal-hussein:
https://github.com/rancher/k3k/issues/345

### DIFF
--- a/k3k-kubelet/provider/provider.go
+++ b/k3k-kubelet/provider/provider.go
@@ -340,7 +340,14 @@ func (p *Provider) CreatePod(ctx context.Context, pod *corev1.Pod) error {
 
 // createPod takes a Kubernetes Pod and deploys it within the provider.
 func (p *Provider) createPod(ctx context.Context, pod *corev1.Pod) error {
-	tPod := pod.DeepCopy()
+	// fieldPath envs are not being translated correctly using the virtual kubelet pod controller
+	// as a workaround we will try to fetch the pod from the virtual cluster and copy over the envSource
+	var sourcePod corev1.Pod
+	if err := p.VirtualClient.Get(ctx, types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, &sourcePod); err != nil {
+		return err
+	}
+
+	tPod := sourcePod.DeepCopy()
 	p.Translator.TranslateTo(tPod)
 
 	// get Cluster definition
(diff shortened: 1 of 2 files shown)


k3k-kubelet/provider/provider.go
```diff
@@ -869,26 +881,42 @@ func configureNetworking(pod *corev1.Pod, podName, podNamespace, serverIP, dnsIP
...
 		return orig
 	}
 
+	// add env vars if not found
+	var found bool
+
+	for _, newEnv := range updated {
+		for _, env := range orig {
+			if newEnv.Name == env.Name {
+				found = true
```
enrichman: There is a small bug, the `found` is never reset to `false`. The test are passing because one was not updated to match the expected behavior:

### Old test
```go
{
	name: "orig and new have multiple elements and some not matching",
	args: args{
		orig: []corev1.EnvVar{{Name: "FOO_0", Value: "old_val_0"}, {Name: "FOO_1", Value: "old_val_1"}},
		new:  []corev1.EnvVar{{Name: "FOO_1", Value: "new_val_1"}, {Name: "FOO_2", Value: "val_1"}},
	},
	want: []corev1.EnvVar{{Name: "FOO_0", Value: "old_val_0"}, {Name: "FOO_1", Value: "new_val_1"}},
},
```

### Updated test
```go
{
	name: "orig and new have multiple elements and some not matching",
	args: args{
		orig: []corev1.EnvVar{{Name: "FOO_0", Value: "old_val_0"}, {Name: "FOO_1", Value: "old_val_1"}},
		new:  []corev1.EnvVar{{Name: "FOO_1", Value: "new_val_1"}, {Name: "FOO_2", Value: "val_1"}},
	},
	want: []corev1.EnvVar{{Name: "FOO_0", Value: "old_val_0"}, {Name: "FOO_1", Value: "new_val_1"}, {Name: "FOO_2", Value: "val_1"}},
},
```

Another approach to the implementation would be updating the `updatedEnvVarMap` when the env is found. We can delete it from the map, so we know that the remaining envVars needs to be added, because they were simply missing.

```go
	for i, env := range orig {
		if updatedEnv, ok := updatedEnvVarMap[env.Name]; ok {
			orig[i] = updatedEnv
			// Remove the updated variable from the map
			delete(updatedEnvVarMap, env.Name)
		}
	}

	// Any variables remaining in the map are new and should be appended to the original slice.
	for _, env := range updatedEnvVarMap {
		orig = append(orig, env)
	}
```

I'm not going to take credits of this smart approach though. :smile:

enrichman (changes requested):
Small bug in the implementation.

Issue #365, opened 2025-05-28
Sync service LoadBalancer IP back


jp-gouin:
**Environmental Info:**
Host Cluster Version: 
RKE2 v1.31.5 

Host Cluster Configuration:
Single node cluster

K3K Cluster Configuration:
`shared` mode virtual cluster with `1` server

**Describe the bug:**
When a virtual cluster user creates a service type `LoadBalancer` the `EXTERNAL-IP` stay <pending>

On the host cluster the 'EXTERNAL-IP' is assigned 
```
k get svc -n k3k-my-cluster
NAME                                                              TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                     AGE
nginx-default-my-cluster-6e67696e782b64656661756c742b6d79-01074   LoadBalancer   10.43.189.2     10.10.10.10   80:32044/TCP                41m
``` 
But not in the virtual cluster
```
k get svc
NAME         TYPE           CLUSTER-IP    EXTERNAL-IP   PORT(S)        AGE
nginx        LoadBalancer   10.43.189.2   <pending>     80:32044/TCP   42m
```

**Expected behavior:**
External IP is also available on the service in the virtual cluster

**Actual behavior:**
External IP is only available on the host cluster service

PR #451, opened 2025-09-02
Lb service status sync

galal-hussein:
- #365

### DIFF
--- a/pkg/controller/cluster/service.go
+++ b/pkg/controller/cluster/service.go
@@ -53,6 +53,7 @@ func (r *ServiceReconciler) filterResources(object ctrlruntimeclient.Object) boo
 func (r *ServiceReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	log := ctrl.LoggerFrom(ctx)
 	log.Info("ensuring service status to virtual cluster")
+
 	var (
 		virtServiceName, virtServiceNamespace string
 		hostService, virtService              v1.Service


main.go
```diff
@@ -123,6 +123,12 @@ func run(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to add the new cluster controller: %v", err)
 	}
 
+	logger.Info("adding service status controller")
+
+	if err := cluster.AddServiceStatusController(ctx, mgr, maxConcurrentReconciles); err != nil {
+		return fmt.Errorf("failed to add the new cluster controller: %v", err)
+	}
```
enrichman: I would change this to be a more generic `ServiceController`. The fact that is updating just the Status is probably an internal specification. Also the log of the previous controller should be updated.

```suggestion
	logger.Info("adding service controller")

	if err := cluster.AddServiceController(ctx, mgr, maxConcurrentReconciles); err != nil {
		return fmt.Errorf("failed to add the new service controller: %v", err)
	}
```

pkg/controller/cluster/service.go
```diff
@@ -0,0 +1,177 @@
...
+type ServiceStatusReconciler struct {
+	ClustersContexts map[string]*syncer.SyncerContext
+	HostClient       ctrlruntimeclient.Client
+}
+
+// Add adds a new controller to the manager
+func AddServiceStatusController(ctx context.Context, mgr manager.Manager, maxConcurrentReconciles int) error {
+	reconciler := ServiceStatusReconciler{
+		HostClient:       mgr.GetClient(),
+		ClustersContexts: make(map[string]*syncer.SyncerContext),
```
enrichman: I'm a bit concerned about this map from a memory perspective, as it doesn't appear to be cleaned up. Since creating a Client isn't too expensive, I think we could remove the map and instantiate them directly when needed.

pkg/controller/cluster/service.go
```diff
@@ -0,0 +1,177 @@
...
+
+	for labelName := range labels {
+		// search for the clusterName Label which indicates its a synced service resource
+		if labelName == translate.ClusterNameLabel {
+			return true
+		}
+	}
+
+	return false
+}
```
enrichman: It seems like we have a similar check for labels/annotations inside the Reconcile function. Could we remove that duplicate logic and rely entirely on the filter here? That would ensure we only check once and simplify the Reconcile function's responsibility. Does that make sense?

Also, while we're looking at this function, I think we can make it more concise. Instead of looping through all the labels, we could check for the key's existence directly.

```suggestion
func (r *ServiceStatusReconciler) filterResources(object ctrlruntimeclient.Object) bool {
	_, hasClusterNameLabel := object.GetLabels()[translate.ClusterNameLabel]
	_, hasNameAnnotation := object.GetAnnotations()[translate.ResourceNameAnnotation]
	_, hasNamespaceAnnotation := object.GetAnnotations()[translate.ResourceNamespaceAnnotation]
	
	// Return true only if all were found
	return hasClusterNameLabel && hasNameAnnotation && hasNamespaceAnnotation
}
```

pkg/controller/cluster/service.go
```diff
@@ -0,0 +1,177 @@
...
+		return fmt.Errorf("empty cluster config")
+	}
+
+	token := clusterConfig.Data["token"]
+
+	virtConfig, err := client.VirtRestConfig(ctx, "", r.HostClient, clusterName, clusterNamespace, string(token))
+	if err != nil {
+		return err
+	}
+	fmt.Printf("\n\n\nvirtRestConfig %#v\n", *virtConfig)
```
enrichman: Leftover.

pkg/controller/cluster/service.go
```diff
@@ -0,0 +1,177 @@
...
+		if labelName == translate.ClusterNameLabel {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (r *ServiceStatusReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := ctrl.LoggerFrom(ctx).WithValues("name", req.Name, "namespace", req.Namespace)
```
enrichman: The logger already contains the name and namespace of the request:

```json
{
    "level": "info",
    "timestamp": "2025-09-23T14:42:18.391Z",
    "msg": "reconciling service status",
    "controller": "k3k-service-status-controller",
    "controllerGroup": "",
    "controllerKind": "Service",
    "Service": {
        "name": "lb-svc3-default-mycluster-6c622d737663332b64656661756c742-ffb57",
        "namespace": "k3k-mycluster"
    },
    "namespace": "k3k-mycluster",
    "name": "lb-svc3-default-mycluster-6c622d737663332b64656661756c742-ffb57",
    "reconcileID": "2c5954c7-a8c1-4126-bf70-f21a6bc7e942"
}
```

---

We should also cleanup this from the other controllers. I'll open a new issue about this.

```json
{
    "level": "info",
    "timestamp": "2025-09-23T14:42:18.438Z",
    "msg": "ensuring *v1.Service",
    "controller": "cluster",
    "controllerGroup": "k3k.io",
    "controllerKind": "Cluster",
    "Cluster": {
        "name": "mycluster",
        "namespace": "k3k-mycluster"
    },
    "namespace": "k3k-mycluster",
    "name": "mycluster",
    "reconcileID": "39d3cd9f-cb3f-4b32-9a10-f1ea61231d12",
    "cluster": {
        "name": "mycluster",
        "namespace": "k3k-mycluster"
    },
    "key": {
        "name": "k3k-mycluster-kube-dns",
        "namespace": "k3k-mycluster"
    }
}
```

enrichman (changes requested):
Thanks for the PR, it looks good! I've left a few comments from my first pass. Let's sync up again after you've had a chance to address the changes. Let me know if you have any questions!

k3k-kubelet/client/client.go
```diff
@@ -0,0 +1,156 @@
...
+
+func LoadTLSConfig(ctx context.Context, hostClient ctrlruntimeclient.Client, clusterName, clusterNamespace, nodeName, hostname, token, agentIP string) (*tls.Config, error) {
+	var (
+		cluster v1alpha1.Cluster
+		b       *bootstrap.ControlRuntimeBootstrap
+	)
+
+	if err := hostClient.Get(ctx, types.NamespacedName{Name: clusterName, Namespace: clusterNamespace}, &cluster); err != nil {
+		return nil, err
+	}
```
enrichman: It looks like we can drop this `Get`. To get the `endpoint` we are just using the `cluster.Name` and `cluster.Namespace` that we already have. So we can remove also the `hostClient` from the signature.

k3k-kubelet/client/client.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if err := retry.OnError(controller.Backoff, func(err error) bool {
+		return err != nil
+	}, func() error {
+		var err error
+		b, err = bootstrap.DecodedBootstrap(token, endpoint)
+		return err
+	}); err != nil {
+		return nil, errors.New("unable to decode bootstrap: " + err.Error())
+	}
```
enrichman: I am not sure that we should keep here this retry. It looks something the caller should handle, otherwise it's pretty hidden.

k3k-kubelet/client/client.go
```diff
@@ -0,0 +1,156 @@
...
+	// POD IP
+	podIP := net.ParseIP(os.Getenv("POD_IP"))
+	ip := net.ParseIP(agentIP)
```
enrichman: This `POD_IP` env variable looks a bit magical. Since this func is creating a TLSConfig it's probably better to accept a `*certutil.AltNames` instead, and instead of passing the `agentIP`, or fetching the `POD_IP` from the env, we can provide the expected AltNames, so it's also clearer we are using this as alternative names of the TLS Config.

pkg/controller/cluster/service.go
```diff
@@ -0,0 +1,162 @@
...
+
+	_, hasNamespaceAnnotation := object.GetAnnotations()[translate.ResourceNamespaceAnnotation]
+
+	// Return true only if all were found
+	return hasClusterNameLabel && hasNameAnnotation && hasNamespaceAnnotation
+}
+
+func (r *ServiceReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+	ctx = ctrl.LoggerInto(ctx, log)
```
enrichman:
```suggestion
```

pkg/controller/cluster/service.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	if err := r.clusterInfo(ctx, &hostService); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to get cluster info: %v", err)
+	}
+
+	// ignoring the service if annotation is empty or resoruce Name/Namespace is not assigned
+	if hostService.Annotations == nil {
+		return reconcile.Result{}, nil
+	}
```
enrichman: We are already filtering the resources in the predicate

```suggestion
```

pkg/controller/cluster/service.go
```diff
@@ -0,0 +1,162 @@
...
+
+	virtServiceName = hostService.Annotations[translate.ResourceNameAnnotation]
+	virtServiceNamespace = hostService.Annotations[translate.ResourceNamespaceAnnotation]
+
+	if virtServiceName == "" || virtServiceNamespace == "" {
+		return reconcile.Result{}, nil
+	}
```
enrichman: Same, we are already checking the resources in the predicate

```suggestion
```

pkg/controller/cluster/service.go
```diff
@@ -0,0 +1,162 @@
...
+	if err := v1.AddToScheme(baseScheme); err != nil {
+		return err
+	}
+
+	virtClient, err := ctrlruntimeclient.New(virtConfig, ctrlruntimeclient.Options{
+		Scheme: baseScheme,
+	})
+	if err != nil {
+		return err
+	}
```
enrichman: You don't have to create a new schema if you are planning to use the client with standard k8s api:

```suggestion
	virtClient, err := ctrlruntimeclient.New(virtConfig, ctrlruntimeclient.Options{})
	if err != nil {
		return err
	}
```

pkg/controller/cluster/service.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/rancher/k3k/pkg/apis/k3k.io/v1alpha1"
+	"github.com/rancher/k3k/pkg/controller"
+)
+
+const (
+	serviceController = "k3k-service-controller"
+)
+
+type ServiceReconciler struct {
+	ClusterContext *syncer.SyncerContext
```
enrichman: I don't think we need to have a `syncer.SyncerContext`, we are just using the VirtualClient.

pkg/controller/cluster/service.go
```diff
@@ -0,0 +1,162 @@
...
+		virtService.Status.LoadBalancer = hostService.Status.LoadBalancer
+		if err := r.ClusterContext.VirtualClient.Status().Update(ctx, &virtService); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (r *ServiceReconciler) clusterInfo(ctx context.Context, obj ctrlruntimeclient.Object) error {
```
enrichman: This `clusterInfo` func is not very clear from the name. You have to read it to understand what it does. I think this should just create a Client for the VirtualCluster, so it should be something like:
  
```go
func newVirtualClient(ctx, Cluster) {
    // fetch kubeconfig

    // create client

    return client, nil
}
```

This func will be probably very useful in other scenario (creating the client from just the Cluster).

pkg/controller/cluster/service.go
```diff
@@ -0,0 +1,162 @@
...
+	// get cluster information from the object
+	labels := obj.GetLabels()
+
+	if labels == nil {
+		return fmt.Errorf("object's labels are empty")
+	}
```
enrichman: This check can be skipped, if we are checking the name after (a get from a nil map doesn't panic).

k3k-kubelet/client/client.go
galal-hussein: The POD_IP is being passed to the server/agent pods as a status
```
{
						Name: "POD_IP",
						ValueFrom: &v1.EnvVarSource{
							FieldRef: &v1.ObjectFieldSelector{
								APIVersion: "v1",
								FieldPath:  "status.podIP",
							},
						},
					},
```

But anyways this code has been around, its just moved to a new package, I will revert it back to kubelet.go, but we can revisit it later if we decided to refactor the tls loading

pkg/controller/cluster/service.go
galal-hussein: we do need the translator too, I thought it would be easier to just reuse the syncerContext

Issue #453, opened 2025-09-03
Set limit/request for the controller


jp-gouin:
User should be able to define limit/request for the controller in the chart.

aganesh-suse:
Validated with k3kcli: 
```
$ k3kcli -v
k3kcli version v0.3.5-rc1
```

Used values.yaml: 
```
controller:
     resources:
        requests:
            cpu: "200m"
            memory: "100Mi"
        limits:
            cpu: "1500m"
            memory: "10Gi"
```
install k3k with: 
```
helm repo add k3k https://rancher.github.io/k3k &&     helm repo update &&     export KUBECONFIG=/etc/rancher/rke2/rke2.yaml &&     helm install --namespace k3k-system --create-namespace k3k k3k/k3k --devel -f /home/ubuntu/values.yaml
```

The k3k controller pod got set with: 
```
$ alias kctlrke2='/var/lib/rancher/rke2/bin/kubectl --kubeconfig /etc/rancher/rke2/rke2.yaml'
$ kctlrke2 describe pod/k3k-79fbb5f895-qv7jn -n k3k-system
.
.
    Limits:
      cpu:     1500m
      memory:  10Gi
    Requests:
      cpu:     200m
      memory:  100Mi
```

Issue #416, opened 2025-07-17
ImagePullSecrets not used


jp-gouin:
The field `ImagePullSecrets` in the chart is not used. 
This should be used to authenticate against registry to deploy
- K3k controller
- K3k virtual-kubelet
- K3s server/agent

aganesh-suse:
```
$ kctlrke2 get secrets -A | grep docker-secret
default       docker-secret                                                     kubernetes.io/dockerconfigjson     1      16h
k3k-system    docker-secret                                                     kubernetes.io/dockerconfigjson     1      58m
kube-system   docker-secret                                                     kubernetes.io/dockerconfigjson     1      16h
```
deployed another cluster - and did not see the log line. LGTM. we can close this issue as well if everything looks good to you as well. @galal-hussein

PR #455, opened 2025-09-03
Add imagepullsecrets to controller, server, and agents

galal-hussein:
- Refactor the chart values file, to have 4 sections: global section, controller section under option `controller`, server section under option `server` and agent section under option `agent`
- Add image pull secrets option to controller, server, and agent sections
- Add resource limits/requests to controller section

Issues:
- #416 
- #453


charts/k3k/values.yaml
FraPazGal, line 1: Regarding image-related values, I propose the following:

- Keeping the global `imagePullSecrets` under `global.imagePullSecrets`. This way, a user that uses a single pull secret for all the containers can easily configure it.
- Supporting an additional `xxxxx.image.registry` parameter for each container and it's global alternative `global.imageRegistry` users could easily use a non-default registry for the chart as a whole or a single container.

```
global:
  # -- Global override for container image registry
  imageRegistry: ""
  # -- Global override for container image registry pull secrets
  imagePullSecrets: []
```

charts/k3k/templates/deployment.yaml
```diff
@@ -15,25 +15,54 @@ spec:
       labels:
         {{- include "k3k.selectorLabels" . | nindent 8 }}
     spec:
+      imagePullSecrets:
+      {{ range .Values.controller.imagePullSecrets }}
+      - name: {{ .name }}
+      {{ end }}
```
FraPazGal: As a best practice, we try to simplify the structure of `imagePullSecrets` to account for the user defining a list of strings or a list of maps `{name: pullSecretName}`. Helm supports both implementations so we standardize this by using:

```suggestion
      {{- with (coalesce .Values.global.imagePullSecrets .Values.controller.imagePullSecrets)}}
      imagePullSecrets:
        {{- $imagePullSecrets := list }}
            {{- range . }}
                {{- if kindIs "string" . }}
                    {{- $imagePullSecrets = append $imagePullSecrets (dict "name" .) }}
                {{- else }}
                    {{- $imagePullSecrets = append $imagePullSecrets . }}
            {{- end }}
        {{- end }}
        {{- toYaml $imagePullSecrets | nindent 8 }}
      {{- end }}
```

Additionally, this snippet should be reusable when definning `--agent-image-pull-secret` and `--server-image-pull-secret` below.

galal-hussein: I have added the snippet to the controller secrets only, since the `agent-image-pull-secret` and `server-image-pull-secret` are both a cli flags so I dont think we should even support `{name: pullSecretName}`

charts/k3k/templates/deployment.yaml
```diff
@@ -15,25 +21,68 @@ spec:
...
+                {{- else }}
+                    {{- $imagePullSecrets = append $imagePullSecrets . }}
+            {{- end }}
+        {{- end }}
+        {{- toYaml $imagePullSecrets | nindent 8 }}
+        {{- end }}
       containers:
-        - image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
-          imagePullPolicy: {{ .Values.image.pullPolicy }}
+        - image: "{{- $controllerRegistry }}/{{ .Values.controller.image.repository }}:{{ default .Chart.AppVersion .Values.controller.image.tag }}"
```
enrichman: I know that it will be a bit long, but probably it's better to have it inline for clarity, wdyt?

```suggestion
        - image: "{{- default .Values.imageRegistry .Values.controller.image.registry }}/{{ .Values.controller.image.repository }}:{{ default .Chart.AppVersion .Values.controller.image.tag }}"
```

pkg/controller/cluster/cluster.go
```diff
@@ -66,15 +66,22 @@ type ClusterReconciler struct {
...
+	K3SServerImagePullPolicy    string
+	K3SServerImageRegistry      string
+	PortAllocator               *agent.PortAllocator
+	ServerImagePullSecrets      []string
+	AgentImagePullSecrets       []string
 }
 
 // Add adds a new controller to the manager
-func Add(ctx context.Context, mgr manager.Manager, sharedAgentImage, sharedAgentImagePullPolicy, k3SImage string, k3SImagePullPolicy string, maxConcurrentReconciles int, portAllocator *agent.PortAllocator, eventRecorder record.EventRecorder) error {
+func Add(ctx context.Context, mgr manager.Manager, sharedAgentImage, sharedAgentImagePullPolicy, sharedAgentImageRegistry, virtualAgentImage, virtualAgentImagePullPolicy, virtualAgentImageRegistry, k3SServerImage, k3SServerImagePullPolicy, k3SServerImageRegistry string, maxConcurrentReconciles int, portAllocator *agent.PortAllocator, eventRecorder record.EventRecorder, serverImagePullSecrets, agentImagePullSecrets []string) error {
```
enrichman: This func is getting way too long (16 arguments). It's time to refactor it. :smile:

charts/k3k/templates/deployment.yaml
```diff
@@ -15,25 +21,68 @@ spec:
...
+        {{- $imagePullSecrets := list }}
+            {{- range . }}
+                {{- if kindIs "string" . }}
+                    {{- $imagePullSecrets = append $imagePullSecrets (dict "name" .) }}
+                {{- else }}
+                    {{- $imagePullSecrets = append $imagePullSecrets . }}
+            {{- end }}
+        {{- end }}
+        {{- toYaml $imagePullSecrets | nindent 8 }}
+        {{- end }}
```
enrichman: Not sure if this can be embedded in an helper function?

FraPazGal: Yes, it could. Just a FYI in case it is helpful, this is a snippet we use in some of our charts:

```
{{/*
    Print the image pull secrets in the expected format (an array of objects with one possible field, "name").
*/}}
{{- define "image.pullSecrets" }}
    {{- $imagePullSecrets := list }}
    {{- range . }}
        {{- if kindIs "string" . }}
            {{- $imagePullSecrets = append $imagePullSecrets (dict "name" .) }}
        {{- else }}
            {{- $imagePullSecrets = append $imagePullSecrets . }}
        {{- end }}
    {{- end }}
    {{- toYaml $imagePullSecrets }}
{{- end }}
```

charts/k3k/values.yaml
```diff
@@ -1,31 +1,10 @@
...
-extraEnv: []
-
-host:
-  # clusterCIDR specifies the clusterCIDR that will be added to the default networkpolicy, if not set
-  # the controller will collect the PodCIDRs of all the nodes on the system.
-  clusterCIDR: ""
+# -- Global override for container image registry
+imageRegistry: "docker.io"
+# -- Global override for container image registry pull secrets
+imagePullSecrets: []
```
FraPazGal: I would include these two parameters affecting the chart as a whole under `global`. To give more context, [global values](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values) are useful when deploying the chart as a subchart (part of a bigger setup) or when the chart uses other subcharts (probably not the use case here).

Besides that, this structure is used by some widely used open source charts which helps UX.

charts/k3k/values.yaml
```diff
@@ -1,31 +1,10 @@
...
-#         name: my-secret
-#         key: my-key
-extraEnv: []
-
-host:
-  # clusterCIDR specifies the clusterCIDR that will be added to the default networkpolicy, if not set
-  # the controller will collect the PodCIDRs of all the nodes on the system.
-  clusterCIDR: ""
+# -- Global override for container image registry
+imageRegistry: "docker.io"
```
FraPazGal: As it is now, the global `imagePullSecrets` and `imageRegistry` values would be overridden by their specific counterparts. We favor the other approach; defining the specific `xxxxx.image.registry` parameter and leaving the global one empty so we can override the specific values in a single place. This way users know they only need to update `global.imageRegistry` for all their charts to use their own registry/pullSecrets regardless of specific per-image values.

That said, given you are leaving the `xxxxx.image.registry` empty, the current setup allows for the workflow I described. If you prefer for the global values to be overridden, the parameters description would need to be updated.

FraPazGal: Unrelated to the above, I would just leave this global `imageRegistry` empty. It doesn't add any value and users may think we are already setting a non-default registry.

charts/k3k/templates/deployment.yaml
```diff
@@ -15,25 +15,57 @@ spec:
       labels:
         {{- include "k3k.selectorLabels" . | nindent 8 }}
     spec:
+      imagePullSecrets: {{- include "image.pullSecrets" (concat .Values.controller.imagePullSecrets .Values.global.imagePullSecrets) | nindent 8 }}
       containers:
-        - image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
-          imagePullPolicy: {{ .Values.image.pullPolicy }}
+        - image: "{{- .Values.global.imageRegistry | default .Values.controller.image.registry }}/{{ .Values.controller.image.repository }}:{{ default .Chart.AppVersion .Values.controller.image.tag }}"
```
FraPazGal: Minor nitpick: for consistency's sake, I would use the same format for `default` here, be it `default value1 value2` or `value1 | default value2`

FraPazGal (approved):
Looks great @galal-hussein! I added a minor suggestion but LGTM

main.go
```diff
@@ -67,13 +63,20 @@ func main() {
...
-	rootCmd.PersistentFlags().StringVar(&sharedAgentImage, "shared-agent-image", "", "K3K Virtual Kubelet image")
-	rootCmd.PersistentFlags().StringVar(&sharedAgentImagePullPolicy, "shared-agent-pull-policy", "", "K3K Virtual Kubelet image pull policy must be one of Always, IfNotPresent or Never")
+	rootCmd.PersistentFlags().StringVar(&config.ClusterCIDR, "cluster-cidr", "", "Cluster CIDR to be added to the networkpolicy")
+	rootCmd.PersistentFlags().StringVar(&config.SharedAgentImage, "shared-agent-image", "", "K3K Virtual Kubelet image")
+	rootCmd.PersistentFlags().StringVar(&config.SharedAgentImagePullPolicy, "shared-agent-image-pull-policy", "", "K3K Virtual Kubelet image pull policy must be one of Always, IfNotPresent or Never")
+	rootCmd.PersistentFlags().StringVar(&config.SharedAgentImageRegistry, "shared-agent-image-registry", "docker.io", "K3K Virtual Kubelet image registry")
 	rootCmd.PersistentFlags().StringVar(&kubeletPortRange, "kubelet-port-range", "50000-51000", "Port Range for k3k kubelet in shared mode")
+	rootCmd.PersistentFlags().StringVar(&config.VirtualAgentImage, "virtual-agent-image", "", "K3S Virtual Agent image")
```
enrichman: Should we have a default here?

main.go
```diff
@@ -67,13 +63,20 @@ func main() {
 
 	rootCmd.PersistentFlags().BoolVar(&debug, "debug", false, "Debug level logging")
 	rootCmd.PersistentFlags().StringVar(&kubeconfig, "kubeconfig", "", "kubeconfig path")
-	rootCmd.PersistentFlags().StringVar(&clusterCIDR, "cluster-cidr", "", "Cluster CIDR to be added to the networkpolicy")
-	rootCmd.PersistentFlags().StringVar(&sharedAgentImage, "shared-agent-image", "", "K3K Virtual Kubelet image")
-	rootCmd.PersistentFlags().StringVar(&sharedAgentImagePullPolicy, "shared-agent-pull-policy", "", "K3K Virtual Kubelet image pull policy must be one of Always, IfNotPresent or Never")
+	rootCmd.PersistentFlags().StringVar(&config.ClusterCIDR, "cluster-cidr", "", "Cluster CIDR to be added to the networkpolicy")
+	rootCmd.PersistentFlags().StringVar(&config.SharedAgentImage, "shared-agent-image", "", "K3K Virtual Kubelet image")
```
enrichman: Should we have a default?

pkg/controller/cluster/agent/shared.go
```diff
@@ -202,7 +206,7 @@ func (s *SharedAgent) podSpec() v1.PodSpec {
 		Containers: []v1.Container{
 			{
 				Name:            s.Name(),
-				Image:           s.image,
+				Image:           s.imageRegistry + "/" + s.image,
```
enrichman: I think this could break if the imageRegistry is not defined.

pkg/controller/cluster/server/server.go
```diff
@@ -38,16 +38,20 @@ type Server struct {
...
+func New(cluster *v1alpha1.Cluster, client client.Client, token, k3SImage, k3SImagePullPolicy, k3SImageRegistry string, imagePullSecrets []string) *Server {
 	return &Server{
 		cluster:            cluster,
 		client:             client,
 		token:              token,
 		mode:               string(cluster.Spec.Mode),
 		k3SImage:           k3SImage,
 		k3SImagePullPolicy: k3SImagePullPolicy,
+		imagePullSecrets:   imagePullSecrets,
+		k3SImageRegistry:   k3SImageRegistry,
```
enrichman: I'm thinking maybe we can drop the `k3s` prefix.

pkg/controller/controller.go
```diff
@@ -28,16 +28,21 @@ var Backoff = wait.Backoff{
 // K3SImage returns the rancher/k3s image tagged with the specified Version.
 // If Version is empty it will use with the same k8s version of the host cluster,
 // stored in the Status object. It will return the untagged version as last fallback.
-func K3SImage(cluster *v1alpha1.Cluster, k3SImage string) string {
+func K3SImage(cluster *v1alpha1.Cluster, k3SImage, registry string) string {
```
enrichman: I think we can change the signature of this func to be more "explicit" and clear, like `Image(registry, image, version, hostVersion string)`. Not sure if we also want to fallback to `latest` if both the version are not specified.

If we change the order it will be also more clear:

```go
func K3SImage(registry, image, version, hostVersion string) string {
	var image string

	if registry != "" {
		image = registry + "/"
	}

	version := "latest"
	if cluster.Spec.Version != "" {
		version = cluster.Spec.Version
	} else if cluster.Status.HostVersion != "" {
		version = cluster.Status.HostVersion
	}

	return image + k3sImage + ":" + version
}
```

pkg/controller/controller.go
```diff
@@ -28,16 +28,21 @@ var Backoff = wait.Backoff{
...
+	var image string
 	if cluster.Spec.Version != "" {
-		return k3SImage + ":" + cluster.Spec.Version
+		image = k3SImage + ":" + cluster.Spec.Version
 	}
 
 	if cluster.Status.HostVersion != "" {
-		return k3SImage + ":" + cluster.Status.HostVersion
+		image = k3SImage + ":" + cluster.Status.HostVersion
 	}
```
enrichman: This will overwrite any version set by the user. We will always use the HostVersion with this.

enrichman (changes requested):
Couple of comments. Probably we need a couple of tests to avoid the issue I saw. We should create a Cluster with a specific version and see the deployed cluster has that.

jp-gouin (approved):
Tested `registry` at `global` and per component level
Tested with multiple `imagePullSecrets` at `global` and per component level. Secrets are correctly propagated to all K3k components (server, agent and controller)

Issue #479, opened 2025-09-25
Upgrade Path - Env and Args upgrade


galal-hussein:
The user should be able to update or add new extra args or extra envs to cluster spec and it should reflect automatically on the servers/agent pods

Validation Steps:

### ENV update

1- create a cluster with initial extra-env
```
k3kcli cluster create --namespace shared-x --server-envs "foo=bar" sharedcluster
```
2- make sure that the server pod has that env
3- update the env by editing the cluster spec to add a new env "foo1=bar1" and edit the current env from `foo=bar` to `foo2=bar2`
4- server pod should be restarted with the new envs

### Args update

1- create a cluster
2- edit the cluster object to add a new arg:
```
serverArgs:
- --disable-network-policy
```
3- the server pod should be restarted with that new arg

repeat the same tests for virtual mode as well

mdrahman-suse:
# Validated on v1.0.0-rc2

```
$ k3kcli -v
k3kcli version v1.0.0-rc2

$ rke2 -v
rke2 version v1.33.5+rke2r1 (d1092839cf08cb901b1d40461b0fa6e7ae6f8fc4)
go version go1.24.6 X:boringcrypto
```

### Shared mode 
- Created k3k cluster
```
KUBECONFIG=/etc/rancher/rke2/rke2.yaml k3kcli cluster create --namespace shared-md --server-envs "foo=bar" sharedcluster-md
```
- Host cluster
```
$ KUBECONFIG=/etc/rancher/rke2/rke2.yaml kubectl get nodes,pods -A
NAME          STATUS   ROLES                       AGE    VERSION
node/ubuntu   Ready    control-plane,etcd,master   6h6m   v1.33.5+rke2r1

NAMESPACE            NAME                                                                  READY   STATUS      RESTARTS   AGE
k3k-system           pod/k3k-564c977767-77ch2                                              1/1     Running     0          6h3m
kube-system          pod/cloud-controller-manager-ubuntu                                   1/1     Running     0          6h6m
kube-system          pod/etcd-ubuntu                                                       1/1     Running     0          6h6m
kube-system          pod/helm-install-rke2-canal-h6mc8                                     0/1     Completed   0          6h6m
kube-system          pod/helm-install-rke2-coredns-5qmkl                                   0/1     Completed   0          6h6m
kube-system          pod/helm-install-rke2-ingress-nginx-kwhsp                             0/1     Completed   0          6h6m
[...]

PR #481, opened 2025-09-25
Adding upgrade path tests

galal-hussein:
- #479

### DIFF
--- a/tests/cluster_upgrade_test.go
+++ b/tests/cluster_upgrade_test.go
@@ -27,12 +27,20 @@ var _ = When("a shared mode cluster update its envs", Label("e2e"), Label("updat
 				Name:  "TEST_SERVER_ENV_1",
 				Value: "not_upgraded",
 			},
+			{
+				Name:  "TEST_SERVER_ENV_2",
+				Value: "toBeRemoved",
+			},
 		}
 		// Add initial environment variables for agent
 		cluster.Spec.AgentEnvs = []corev1.EnvVar{
 			{
 				Name:  "TEST_AGENT_ENV_1",
 				Value: "not_upgraded",
+			},
+			{
+				Name:  "TEST_AGENT_ENV_2",
+				Value: "toBeRemoved",
 			},
 		}
 
@@ -54,6 +62,10 @@ var _ = When("a shared mode cluster update its envs", Label("e2e"), Label("updat
 		Expect(ok).To(BeTrue())
 		Expect(serverEnv1).To(Equal("not_upgraded"))
 
+		serverEnv2, ok := getEnv(&serverPod, "TEST_SERVER_ENV_2")
+		Expect(ok).To(BeTrue())
+		Expect(serverEnv2).To(Equal("toBeRemoved"))
+
 		aPods := listAgentPods(ctx, virtualCluster)
 		Expect(len(aPods)).To(Equal(1))
 
@@ -62,6 +74,10 @@ var _ = When("a shared mode cluster update its envs", Label("e2e"), Label("updat
 		agentEnv1, ok := getEnv(&agentPod, "TEST_AGENT_ENV_1")
 		Expect(ok).To(BeTrue())
 		Expect(agentEnv1).To(Equal("not_upgraded"))
+
+		agentEnv2, ok := getEnv(&agentPod, "TEST_AGENT_ENV_2")
+		Expect(ok).To(BeTrue())
+		Expect(agentEnv2).To(Equal("toBeRemoved"))
 	})
 	It("will update server and agent envs when cluster is updated", func() {
 		Eventually(func(g Gomega) {
@@ -77,7 +93,7 @@ var _ = When("a shared mode cluster update its envs", Label("e2e"), Label("updat
 					Value: "upgraded",
 				},
 				{
-					Name:  "TEST_SERVER_ENV_2",
+					Name:  "TEST_SERVER_ENV_3",
 					Value: "new",
 				},
 			}
@@ -87,7 +103,7 @@ var _ = When("a shared mode cluster update its envs", Label("e2e"), Label("updat
 					Value: "upgraded",
 				},
 				{
-					Name:  "TEST_AGENT_ENV_2",
+					Name:  "TEST_AGENT_ENV_3",
 					Value: "new",
 				},
 			}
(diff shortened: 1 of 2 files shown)


tests/cluster_upgrade_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = When("a shared mode cluster upgrade its envs", Label("e2e"), Label("upgrade"), func() {
```
enrichman: Do we need a `Label("upgrade")` for these tests? Also let's keep the `update` keyword for clarity!

```suggestion
var _ = When("a shared mode cluster update its envs", Label("e2e"), func() {
```

tests/cluster_upgrade_test.go
```diff
@@ -0,0 +1,166 @@
...
+				Name:  "TEST_AGENT_ENV_1",
+				Value: "upgraded",
+			},
+			{
+				Name:  "TEST_AGENT_ENV_2",
+				Value: "new",
+			},
+		}
+		err = k8sClient.Update(ctx, cluster)
+		Expect(err).NotTo(HaveOccurred())
```
enrichman: The `BeforeEach` should setup the tests, so this should probably be part of a test.

tests/cluster_upgrade_test.go
```diff
@@ -0,0 +1,166 @@
...
+			agentEnv2 := getEnv(&aPods[0], "TEST_AGENT_ENV_2")
+			g.Expect(agentEnv2).To(Equal("new"))
+		}).
+			WithPolling(time.Second * 2).
+			WithTimeout(time.Second * 120).
+			Should(Succeed())
+	})
+})
+
+var _ = When("a shared mode cluster upgrade its server args", Label("e2e"), Label("upgrade"), func() {
```
enrichman: Same as for the envs

```suggestion
var _ = When("a shared mode cluster update its server args", Label("e2e"), func() {
```

tests/cluster_upgrade_test.go
```diff
@@ -0,0 +1,166 @@
...
+			g.Expect(len(aPods)).To(Equal(1))
+
+			agentEnv1 := getEnv(&aPods[0], "TEST_AGENT_ENV_1")
+			g.Expect(agentEnv1).To(Equal("upgraded"))
+
+			agentEnv2 := getEnv(&aPods[0], "TEST_AGENT_ENV_2")
+			g.Expect(agentEnv2).To(Equal("new"))
+		}).
+			WithPolling(time.Second * 2).
+			WithTimeout(time.Second * 120).
```
enrichman: NIT
```suggestion
			WithTimeout(time.Minute * 2).
```

tests/cluster_upgrade_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+		err := k8sClient.Get(ctx, ctrlruntimeclient.ObjectKeyFromObject(cluster), cluster)
+		Expect(err).NotTo(HaveOccurred())
+
+		cluster.Spec.ServerArgs = []string{
+			"--node-label=test_server=upgraded",
+		}
+
+		err = k8sClient.Update(ctx, cluster)
+		Expect(err).NotTo(HaveOccurred())
```
enrichman: Same: this should probably be part of the Test itself.

tests/cluster_upgrade_test.go
```diff
@@ -0,0 +1,166 @@
...
+	It("will update server args", func() {
+		Eventually(func(g Gomega) {
+			// server pods
+			sPods := serverPods(ctx, virtualCluster)
+			g.Expect(len(sPods)).To(Equal(1))
+
+			g.Expect(isArgFound(&sPods[0], "--node-label=test_server=upgraded")).To(BeTrue())
+		}).
+			WithPolling(time.Second * 2).
+			WithTimeout(time.Second * 120).
```
enrichman: NIT
```suggestion
			WithTimeout(time.Minute * 2).
```

tests/common_test.go
```diff
@@ -332,3 +336,48 @@ func restartServerPod(ctx context.Context, virtualCluster *VirtualCluster) {
 		return serverPods.Items[0].DeletionTimestamp
 	}).WithTimeout(60 * time.Second).WithPolling(time.Second * 5).Should(BeNil())
 }
+
+func serverPods(ctx context.Context, virtualCluster *VirtualCluster) []corev1.Pod {
```
enrichman: Maybe a name like `listServerPods` or `getServerPods` is more descriptive?

tests/common_test.go
```diff
@@ -332,3 +336,48 @@ func restartServerPod(ctx context.Context, virtualCluster *VirtualCluster) {
...
+	labelSelector := "cluster=" + virtualCluster.Cluster.Name + ",role=server"
+
+	serverPods, err := k8s.CoreV1().Pods(virtualCluster.Cluster.Namespace).List(ctx, v1.ListOptions{LabelSelector: labelSelector})
+	Expect(err).To(Not(HaveOccurred()))
+
+	return serverPods.Items
+}
+
+func agentPods(ctx context.Context, virtualCluster *VirtualCluster) []corev1.Pod {
```
enrichman: Same: `listAgentPods` of `getAgentPods`?

tests/common_test.go
```diff
@@ -332,3 +336,48 @@ func restartServerPod(ctx context.Context, virtualCluster *VirtualCluster) {
...
+	labelSelector := "cluster=" + virtualCluster.Cluster.Name + ",type=agent" + ",mode=shared"
+	if virtualCluster.Cluster.Spec.Mode == v1alpha1.VirtualClusterMode {
+		labelSelector = "cluster=" + virtualCluster.Cluster.Name + ",type=agent" + ",mode=virtual"
+	}
```
enrichman: I think the `%s` should work, otherwise we can use the `%v`

```suggestion
	labelSelector := fmt.Sprintf("cluster=%s,type=agent,mode=%s", virtualCluster.Cluster.Name, virtualCluster.Cluster.Spec.Mode)
```

tests/common_test.go
```diff
@@ -332,3 +336,48 @@ func restartServerPod(ctx context.Context, virtualCluster *VirtualCluster) {
...
+func getEnv(pod *corev1.Pod, envName string) string {
+	container := pod.Spec.Containers[0]
+	for _, envVar := range container.Env {
+		if envVar.Name == envName {
+			return envVar.Value
+		}
+	}
+
+	return ""
+}
```
enrichman, line 368: I would return a `(string, bool)` to check if the var is actually empty or not exists.

enrichman (changes requested):
Could you add also a test also for the removal of them?

Maybe starting with:
```
VAR_1=var_1
VAR_2=var_2
```
and then updating one of the old, removing one of the old, adding a new one
```
VAR_1=var_1_update
VAR_3=var_3
```

enrichman (approved):

PR #492, opened 2025-10-03
Use K3S host cluster for E2E tests

galal-hussein:
The PR uses k3s as a host cluster for e2e tests instead of using docker, docker now should be optional to use by setting environment variable `K3K_DOCKER_INSTALL`


.github/workflows/test.yaml
```diff
@@ -87,10 +87,16 @@ jobs:
...
-        make build
-        make package
+        curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.32.1+k3s1 INSTALL_K3S_EXEC="--write-kubeconfig-mode=777" sh -s - 
+
+    - name: Build and package and push dev images
+      run: |
+        uuidgen > uuidgen.txt
+        REPO="ttl.sh/$(cat uuidgen.txt)" VERSION=1h make build
+        REPO="ttl.sh/$(cat uuidgen.txt)" VERSION=1h make package
+        REPO="ttl.sh/$(cat uuidgen.txt)" VERSION=1h make push
```
enrichman: You can export the variable using the `GITHUB_ENV`, and they will be available across the steps:
```suggestion

    - name: Setup env
      run: |
        echo "REPO=ttl.sh/$(uuidgen)"       >> $GITHUB_ENV
        echo "VERSION=1h"                   >> $GITHUB_ENV

    - name: Build and package and push dev images
      env:
        REPO: ${{ env.REPO }}
        VERSION: ${{ env.VERSION }}
      run: |
        make build
        make package
        make push
```

.github/workflows/test.yaml
```diff
@@ -99,7 +105,7 @@ jobs:
       run: k3kcli -v
 
     - name: Run e2e tests
-      run: make test-e2e
+      run: E2E_KUBECONFIG=/etc/rancher/k3s/k3s.yaml REPO="ttl.sh/$(cat uuidgen.txt)" VERSION=1h make test-e2e
```
enrichman:
```suggestion
      env:
        E2E_KUBECONFIG: /etc/rancher/k3s/k3s.yaml
        REPO: ${{ env.REPO }}
        VERSION: ${{ env.VERSION }}
      run:  make test-e2e
```

tests/common_test.go
```diff
@@ -87,7 +88,7 @@ func NewVirtualClusters(n int) []*VirtualCluster {
 func NewNamespace() *corev1.Namespace {
 	GinkgoHelper()
 
-	namespace := &corev1.Namespace{ObjectMeta: v1.ObjectMeta{GenerateName: "ns-"}}
+	namespace := &corev1.Namespace{ObjectMeta: v1.ObjectMeta{GenerateName: "e2e-ns-", Labels: map[string]string{"e2e": "true"}}}
```
enrichman: I'm ok with renaming the `ns`, just asking if it was for any particular reason, or just to see that they are used for the `e2e`

tests/tests_suite_test.go
```diff
@@ -38,7 +39,7 @@ import (
 )
 
 const (
-	k3kNamespace = "k3k-system"
+	k3kNamespace = "k3k-e2e-system"
```
enrichman: Let's keep this as `k3k-system`. I guess we have it hardcoded somewhere, and something could break if we install K3k in a different one.

```suggestion
	k3kNamespace = "k3k-system"
```

tests/tests_suite_test.go
```diff
@@ -48,47 +49,72 @@ func TestTests(t *testing.T) {
...
-	kubeconfigPath string
+	k3sContainer       *k3s.K3sContainer
+	hostIP             string
+	restcfg            *rest.Config
+	k8s                *kubernetes.Clientset
+	k8sClient          client.Client
+	kubeconfigPath     string
+	repo               string
+	helmActionConfig   *action.Configuration
+	dockerInstallation bool
```
enrichman: I think we can avoid this `dockerInstallation` var and use the `k3sContainer` instead. If that is `nil` that means we are using an external cluster.

tests/tests_suite_test.go
```diff
@@ -432,3 +459,57 @@ func caCertSecret(name, namespace string, crt, key []byte) *corev1.Secret {
 		},
 	}
 }
+
+func cleanupE2EResources(ctx context.Context) error {
```
enrichman: As we said if we are using an external installation we can just leave it. For the containerized tests we can do as before, and destroy the container.

tests/common_test.go
galal-hussein: They conflict with some of k8s conformance tests, but I am okay if you want to revert it back

enrichman: Ah, I see. It's because using `ns-` or `pod-` is probably the idiomatic way. I would keep it as before for the same reason, but maybe it makes sense to have a label to filter them. Maybe `k3k-e2e`, to be more explicit. :slightly_smiling_face:

.github/workflows/test.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
         version: v2.3.0
 
   validate:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest>> $GIT
```
enrichman: Typo. It's also breaking the CI 😄 
```suggestion
    runs-on: ubuntu-latest
```

tests/cluster_upgrade_test.go
enrichman, line 1: I think this file was not supposed to be committed.

tests/tests_suite_test.go
```diff
@@ -48,52 +48,64 @@ func TestTests(t *testing.T) {
...
-	k3sContainer, err = k3s.Run(ctx, "rancher/k3s:v1.32.1-k3s1")
-	Expect(err).To(Not(HaveOccurred()))
+	repo = os.Getenv("REPO")
+	if repo == "" {
+		repo = "rancher"
+	}
 
-	hostIP, err = k3sContainer.ContainerIP(ctx)
-	Expect(err).To(Not(HaveOccurred()))
+	dockerInstallationEnv := os.Getenv("K3K_DOCKER_INSTALL")
```
enrichman: NIT: the `os.LookupEnv` returns also a bool


```os==go
	_, dockerInstallEnabled := os.LookupEnv("K3K_DOCKER_INSTALL")
```

tests/tests_suite_test.go
```diff
@@ -48,52 +48,64 @@ func TestTests(t *testing.T) {
...
-	Expect(os.Setenv("KUBECONFIG", kubeconfigPath)).To(Succeed())
-
-	DeferCleanup(os.Remove, kubeconfigPath)
+	hostIP, err = getServerIP(ctx, restcfg)
+	Expect(err).To(Not(HaveOccurred()))
 
-	initKubernetesClient(kubeconfig)
-	installK3kChart(ctx, kubeconfig)
+	initKubernetesClient()
+	installK3kChart(kubeconfig)
```
enrichman: This part looks a bit weird, it can be probably be simplified, or we can make it more linear. For example it seems the `initKubernetesClient()` is trying to fetch again the `KUBECONFIG`, and we are  also checking it here. I would try to do these steps in separate functions, and call them accordingly if they a re needed or not.

I.e.:
```go
if dockerInstallEnabled {
	installK3SDocker(ctx)
	initKubernetesClient()
	installK3kChart(kubeconfig)
} else {
	initKubernetesClient()
}
```

We can probably also change some of the signature to avoid setting globals everywhere, but not sure how.

tests/tests_suite_test.go
```diff
@@ -121,24 +124,66 @@ func buildScheme() *runtime.Scheme {
...
 
-func installK3kChart(ctx context.Context, kubeconfig []byte) {
+func installK3SDocker(ctx context.Context) {
+	var (
+		err        error
+		kubeconfig []byte
+	)
+
+	k3sHostVersion, ok := os.LookupEnv("K3S_HOST_VERSION")
+	Expect(ok).To(BeTrue())
```
enrichman: If empty let's fallback to a known version, the same we have set in the Github Action: `v1.32.1+k3s1`. This is because if you run the test locally maybe you don't have the var set.

```suggestion
	k3sHostVersion := os.Getenv("K3S_HOST_VERSION")
	if k3sHostVersion == "" {
		k3sHostVersion = "v1.32.1+k3s1"
	}
```

Issue #465, opened 2025-09-16
Upgrade path - K8s upgrade


jp-gouin:
A virtual cluster user should be able to upgrade the K3s version used by the virtual cluster in 

- `shared` mode
- `virtual` mode

It should include a upgrade command group in the k3kcli
it should include an e2e test

mdrahman-suse:
@jp-gouin I tried to create a k3k cluster in virtual mode with 3 servers and 1 agent, but I see only 2 servers and 1 agent when I ran the kubectl commands on that cluster

```
$ k3kcli -v
k3kcli version v1.0.0-rc2
```

```
KUBECONFIG=/etc/rancher/rke2/rke2.yaml k3kcli cluster create --namespace virt-md --servers 3 --agents 1 --version v1.32.9+k3s1 --mode virtual mdvirtcluster

# Host cluster

$ KUBECONFIG=/etc/rancher/rke2/rke2.yaml kubectl get no,po -A
NAME          STATUS   ROLES                       AGE    VERSION
node/ubuntu   Ready    control-plane,etcd,master   4d8h   v1.33.5+rke2r1

NAMESPACE            NAME                                                       READY   STATUS      RESTARTS        AGE
k3k-system           pod/k3k-564c977767-77ch2                                   1/1     Running     1 (25h ago)     4d7h
kube-system          pod/cloud-controller-manager-ubuntu                        1/1     Running     9 (25h ago)     4d8h
kube-system          pod/etcd-ubuntu                                            1/1     Running     1               4d8h
kube-system          pod/helm-install-rke2-canal-h6mc8                          0/1     Completed   0               4d8h
kube-system          pod/helm-install-rke2-coredns-5qmkl                        0/1     Completed   0               4d8h
kube-system          pod/helm-install-rke2-ingress-nginx-kwhsp                  0/1     Completed   0               4d8h
kube-system          pod/helm-install-rke2-metrics-server-rzw7z                 0/1     Completed   0               4d8h
kube-system          pod/helm-install-rke2-runtimeclasses-sgwzc                 0/1     Completed   0               4d8h
kube-system          pod/helm-install-rke2-snapshot-controller-crd-55tgk        0/1     Completed   0               4d8h
kube-system          pod/helm-install-rke2-snapshot-controller-fdw8r            0/1     Completed   1               4d8h
kube-system          pod/kube-apiserver-ubuntu                                  1/1     Running     3               4d8h
[...]

Issue #477, opened 2025-09-19
[Virtual mode] Issue while deleting a server pod


jp-gouin:
**Environmental Info:**
Host Cluster Version: 
```
k get nodes -o wide
NAME                                       STATUS   ROLES                              AGE    VERSION           INTERNAL-IP      EXTERNAL-IP   OS-IMAGE                              KERNEL-VERSION            CONTAINER-RUNTIME
chaos-virtual-clusters-pool1-wjp2n-s2tk6   Ready    control-plane,etcd,master,worker   6d     v1.31.11+rke2r1   10.144.100.103   <none>        SUSE Linux Enterprise Server 15 SP6   6.4.0-150600.21-default   containerd://2.0.5-k3s2
chaos-virtual-clusters-pool1-wjp2n-sw7fz   Ready    control-plane,etcd,master,worker   6d1h   v1.31.11+rke2r1   10.144.100.98    <none>        SUSE Linux Enterprise Server 15 SP6   6.4.0-150600.21-default   containerd://2.0.5-k3s2
chaos-virtual-clusters-pool1-wjp2n-w26w4   Ready    control-plane,etcd,master,worker   6d     v1.31.11+rke2r1   10.144.100.101   <none>        SUSE Linux Enterprise Server 15 SP6   6.4.0-150600.21-default   containerd://2.0.5-k3s2
```

K3K Cluster Configuration:
```
apiVersion: k3k.io/v1alpha1
kind: Cluster
metadata:
  name: testchaos
  namespace: test-virtual
spec:
  mode: virtual
  version: v1.32.8-k3s1
  servers: 3
  agents: 3
  persistence:
    type: dynamic
    storageRequestSize: 2Gi
    storageClassName: harvester
```

**Describe the bug:**
When a server pod is deleted, it fail to come back online
```
time="2025-09-18T07:37:35Z" level=fatal msg="Failed to start networking: unable to initialize network policy controller: error getting node subnet: failed to find interface with specified node ip" 
```

**Steps To Reproduce:**
- Create the virtual cluster
- wait for all servers to be running
- delete one of the server pods

**Expected behavior:**
The server pod should come up running after a deletion

jp-gouin:
Probably related to https://github.com/k3s-io/k3s/issues/12844

jp-gouin:
Temporary solution will be to have `--server-args="--disable-network-policy"` by default for `virtual` mode.

PR #503, opened 2025-10-10
Add k8s version upgrade test

galal-hussein:
The PR adds 2 new tests related to cluster version upgrade, the test consists of the following steps:

- Create a cluster with dynamic persistent storage in shared/virtual mode with version "v1.31.13-k3s1"
- Create a nginx pod on the cluster
- Update the cluster spec to upgrade the version to "v1.32.8-k3s1"
- Check if the server/agents pods are updated with the right version
- Check if the k8s cluster version matches the new version
- Check if nginx pod still intact

- #465

### DIFF
--- a/tests/cluster_update_test.go
+++ b/tests/cluster_update_test.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/kubernetes/pkg/api/v1/pod"
 	"k8s.io/utils/ptr"
 
 	v1 "k8s.io/api/core/v1"
@@ -406,8 +407,8 @@ var _ = When("a shared mode cluster update its version", Label("e2e"), func() {
 		virtualCluster *VirtualCluster
 		nginxPod       *v1.Pod
 	)
-	ctx := context.Background()
 	BeforeEach(func() {
+		ctx := context.Background()
 		namespace := NewNamespace()
 
 		cluster := NewCluster(namespace.Name)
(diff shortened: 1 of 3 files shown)


tests/cluster_update_test.go
```diff
@@ -398,3 +400,177 @@ var _ = When("a virtual mode cluster update its server args", Label("e2e"), func
 			Should(Succeed())
 	})
 })
+
+var _ = When("a shared mode cluster update its version", Label("e2e"), func() {
+	var (
+		virtualCluster *VirtualCluster
+		nginxPod       *v1.Pod
+	)
+	ctx := context.Background()
```
enrichman: I would try to be consistent and have the same kind of declaration/assignments. I.e I'm ok declaring the var on top if they are reused, but I would keep the assignment in the inner functions.

i.e.:
```go
var (
	virtualCluster *VirtualCluster
	nginxPod       *v1.Pod
	ctx				context.Context
)

func do() {
	ctx = context.Background()
}
```

or

```go
var (
	virtualCluster *VirtualCluster
	nginxPod       *v1.Pod
)

func do() {
	ctx := context.Background()
}
```

I'm more inclined towards the **second** approach, because I think the context should be scoped on where is used, maybe setting timeouts, cancellation, and such.

tests/cluster_update_test.go
```diff
@@ -398,3 +400,177 @@ var _ = When("a virtual mode cluster update its server args", Label("e2e"), func
...
+			var cluster v1alpha1.Cluster
+
+			err := k8sClient.Get(ctx, ctrlruntimeclient.ObjectKeyFromObject(virtualCluster.Cluster), &cluster)
+			g.Expect(err).NotTo(HaveOccurred())
+
+			// update cluster version
+			cluster.Spec.Version = "v1.32.8-k3s1"
+
+			err = k8sClient.Update(ctx, &cluster)
+			g.Expect(err).NotTo(HaveOccurred())
```
enrichman: This update can be moved outside the `Eventually`, we don't need to run it multiple times.

tests/cluster_update_test.go
```diff
@@ -398,3 +400,177 @@ var _ = When("a virtual mode cluster update its server args", Label("e2e"), func
...
+
+			// server pods
+			serverPods := listServerPods(ctx, virtualCluster)
+			g.Expect(len(serverPods)).To(Equal(1))
+
+			serverPod := serverPods[0]
+			cond := findPodCondition(serverPod.Status.Conditions, v1.PodReady)
```
enrichman: There is a Kubernetes native func to check this: [meta.IsStatusConditionPresentAndEqual](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#IsStatusConditionPresentAndEqual), I've used it here:

https://github.com/rancher/k3k/blob/a0fd4728412c8e1a3fb89cb07737ac1eaf005f80/pkg/controller/cluster/status.go#L82-L94

It looks like there is also a simpler [meta.IsStatusConditionTrue](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#IsStatusConditionTrue), maybe it's fine in this scenario.

tests/common_test.go
```diff
@@ -140,6 +140,9 @@ func NewCluster(namespace string) *v1alpha1.Cluster {
 			Persistence: v1alpha1.PersistenceConfig{
 				Type: v1alpha1.EphemeralPersistenceMode,
 			},
+			ServerArgs: []string{
+				"--disable-network-policy",
+			},
```
enrichman, line 145: Which is the reason about this? I'm a bit concerned because it impacts all the tests.

tests/common_test.go
```diff
@@ -261,32 +264,35 @@ func (c *VirtualCluster) NewNginxPod(namespace string) (*corev1.Pod, string) {
 
 	var podIP string
 
-	// check that the nginx Pod is up and running in the host cluster
-	Eventually(func() bool {
-		podList, err := k8s.CoreV1().Pods(c.Cluster.Namespace).List(ctx, v1.ListOptions{})
-		Expect(err).To(Not(HaveOccurred()))
+	// only check the pod on the host cluster if the mode is shared mode
+	if c.Cluster.Spec.Mode == v1alpha1.SharedClusterMode {
```
enrichman: To avoid adding an extra level of nesting I would do an early return instead:


```go
if c.Cluster.Spec.Mode != v1alpha1.SharedClusterMode {
	return
}
```

tests/common_test.go
galal-hussein: its a known issue unfortuantely in k3s https://github.com/rancher/k3k/issues/477, it has been fixed but not yet included or backported in k3s versions, basically when a server pod restarts and gets a new IP the network policy controller fails to initialize, so we have to disable it

enrichman: Ah, I thought this was about our NetworkPolicies, not the one of K3s!

Issue #464, opened 2025-09-16
Upgrade path - scale up and down


jp-gouin:
A virtual cluster user should be able to scale up and down a virtual cluster in 

- `shared` mode
- `virtual` mode

It should include a `upgrade` command group in the `k3kcli`
it should include an e2e test

PR #508, opened 2025-10-14
Scale up/down tests for virtual and shared mode

galal-hussein:
- #464 
The PR adds 4 tests for scaling up/down in both modes:

#### Scale Up

1. Start a cluster with 1 server
2. Make sure that endpointslices for kubernetes points to one address (since there is no way to confirm the number of nodes)
3. Start a nginx pod
4. scale up the cluster to 3 servers
5. Recheck nginx pod and the endpoint slices and make sure it points to 3 addresses

#### Scale Down

1. Start a cluster with 3 servers
2. Make sure that endpointslices for kubernetes points to 3 addresses
3. Start a nginx pod
4. scale down the cluster to 1  servers
5. Recheck nginx pod and the endpoint slices and make sure it points to 1 addresses

#### Scale Up

1. Start a cluster with 1 server
2. List the nodes in the virtual cluster and make sure there is only 1 server node
3. Start a nginx pod
4. scale up the cluster to 3 servers
5. Recheck nginx pod and list the nodes and make sure that there are now 3 server nodes

#### Scale Down

1. Start a cluster with 3 servers
2. List the nodes in the virtual cluster and make sure there are 3 server nodes
3. Start a nginx pod
4. scale down the cluster to 1  servers
5. Recheck nginx pod and the endpoint slices and make sure it points to 1 addresses (we are checking endpointslices here because nodes will still be there but in a not ready state)

### DIFF
--- a/tests/cluster_update_test.go
+++ b/tests/cluster_update_test.go
@@ -356,7 +356,7 @@ var _ = When("a virtual mode cluster update its envs", Label("e2e"), Label(updat
 	})
 })
 
-var _ = When("a virtual mode cluster update its server args", Label(updateTestsLabel), Label(slowTestsLabel), func() {
+var _ = When("a virtual mode cluster update its server args", Label("e2e"), Label(updateTestsLabel), Label(slowTestsLabel), func() {
 	var virtualCluster *VirtualCluster
 	ctx := context.Background()
 	BeforeEach(func() {
@@ -414,7 +414,7 @@ var _ = When("a virtual mode cluster update its server args", Label(updateTestsL
 	})
 })
 
-var _ = When("a shared mode cluster update its version", Label(updateTestsLabel), Label(slowTestsLabel), func() {
+var _ = When("a shared mode cluster update its version", Label("e2e"), Label(updateTestsLabel), Label(slowTestsLabel), func() {
 	var (
 		virtualCluster *VirtualCluster
 		nginxPod       *v1.Pod
@@ -500,7 +500,7 @@ var _ = When("a shared mode cluster update its version", Label(updateTestsLabel)
 	})
 })
 
-var _ = When("a virtual mode cluster update its version", Label(updateTestsLabel), Label(slowTestsLabel), func() {
+var _ = When("a virtual mode cluster update its version", Label("e2e"), Label(updateTestsLabel), Label(slowTestsLabel), func() {
 	var (
 		virtualCluster *VirtualCluster
 		nginxPod       *v1.Pod
@@ -601,7 +601,7 @@ var _ = When("a virtual mode cluster update its version", Label(updateTestsLabel
 	})
 })
 
-var _ = When("a shared mode cluster scales up servers", Label(updateTestsLabel), Label(slowTestsLabel), func() {
+var _ = When("a shared mode cluster scales up servers", Label("e2e"), Label(updateTestsLabel), Label(slowTestsLabel), func() {
 	var (
 		virtualCluster *VirtualCluster
 		nginxPod       *v1.Pod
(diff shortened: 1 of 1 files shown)


tests/cluster_update_test.go
```diff
@@ -580,3 +580,340 @@ var _ = When("a virtual mode cluster update its version", Label("e2e"), func() {
...
 })
+
+var _ = When("a shared mode cluster scales up servers", Label("e2e"), Label("SLOW"), func() {
+	var (
+		virtualCluster *VirtualCluster
+		nginxPod       *v1.Pod
+	)
+	BeforeEach(func() {
+		ctx := context.Background()
+		namespace := NewNamespace()
```
enrichman, line 611: At the bottom of the `BeforeEach` we need to remember to cleanup:

```go
DeferCleanup(func() {
	DeleteNamespaces(namespace.Name)
})
```

We should probably centralize this logic, but for now it's fine to delete them in the test itself.

tests/cluster_update_test.go
```diff
@@ -580,3 +580,340 @@ var _ = When("a virtual mode cluster update its version", Label("e2e"), func() {
...
+			RestConfig: restConfig,
+			Client:     client,
+		}
+		sPods := listServerPods(ctx, virtualCluster)
+		Expect(len(sPods)).To(Equal(1))
+
+		Eventually(func(g Gomega) {
+			// since there is no way to check nodes in shared mode
+			// we can check if the endpoints are registered to N nodes
+			k8sEndpointSlices, err := virtualCluster.Client.DiscoveryV1().EndpointSlices("default").Get(ctx, "kubernetes", metav1.GetOptions{})
```
enrichman, line 639: I'm not sure what we are checking here.

tests/cluster_update_test.go
```diff
@@ -580,3 +580,340 @@ var _ = When("a virtual mode cluster update its version", Label("e2e"), func() {
...
+	})
+	It("will scale up server pods", func() {
+		var cluster v1beta1.Cluster
+		ctx := context.Background()
+
+		err := k8sClient.Get(ctx, ctrlruntimeclient.ObjectKeyFromObject(virtualCluster.Cluster), &cluster)
+		Expect(err).NotTo(HaveOccurred())
+
+		// scale cluster servers to 3 nodes
+		cluster.Spec.Servers = ptr.To(int32(3))
```
enrichman: I think this should work:

```suggestion
		cluster.Spec.Servers = ptr.To[int32](3)
```

tests/cluster_update_test.go
```diff
@@ -580,3 +580,340 @@ var _ = When("a virtual mode cluster update its version", Label("e2e"), func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		Eventually(func(g Gomega) {
+			// server pods
+			serverPods := listServerPods(ctx, virtualCluster)
+			g.Expect(len(serverPods)).To(Equal(3))
+
+			for _, serverPod := range serverPods {
+				condIndex, cond := pod.GetPodCondition(&serverPod.Status, v1.PodReady)
+				g.Expect(condIndex).NotTo(Equal(-1))
```
enrichman: I think it's fine to check just one of the two return:

```suggestion
				_, cond := pod.GetPodCondition(&serverPod.Status, v1.PodReady)
```

tests/cluster_update_test.go
```diff
@@ -580,3 +580,340 @@ var _ = When("a virtual mode cluster update its version", Label("e2e"), func() {
...
+			}
+
+			k8sEndpointSlices, err := virtualCluster.Client.DiscoveryV1().EndpointSlices("default").Get(ctx, "kubernetes", metav1.GetOptions{})
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(len(k8sEndpointSlices.Endpoints)).To(Equal(3))
+
+			nginxPod, err = virtualCluster.Client.CoreV1().Pods(nginxPod.Namespace).Get(ctx, nginxPod.Name, metav1.GetOptions{})
+			g.Expect(err).To(BeNil())
+			condIndex, cond := pod.GetPodCondition(&nginxPod.Status, v1.PodReady)
+			g.Expect(condIndex).NotTo(Equal(-1))
```
enrichman: Same
```suggestion
			_, cond := pod.GetPodCondition(&nginxPod.Status, v1.PodReady)
```

enrichman (changes requested):
I left some comments only on the first test, but they can be made also on the others.

Issue #478, opened 2025-09-25
Issue with fake PV


jp-gouin:
**Environmental Info:**
Host Cluster Version: 

```
NAME            STATUS   ROLES                              AGE   VERSION          INTERNAL-IP   EXTERNAL-IP      OS-IMAGE                              KERNEL-VERSION                 CONTAINER-RUNTIME
ip-10-0-0-143   Ready    control-plane,etcd,master,worker   17h   v1.31.5+rke2r1   10.0.0.143    x.x.x.x.      SUSE Linux Enterprise Server 15 SP3   5.3.18-150300.59.198-default   containerd://1.7.23-k3s2
```

K3K Cluster Configuration:
`shared` mode, single server

**Describe the bug:**
When an app with a `pvc` is created, a fake `pv` is created in the virtual cluster.
Once the app is deleted, the fake `pv` remain in released status which will block the redeployment of the app. 

```
Events:
  Type     Reason            Age   From               Message
  ----     ------            ----  ----               -------
  Warning  FailedScheduling  12s   default-scheduler  0/1 nodes are available: pod has unbound immediate PersistentVolumeClaims. preemption: 0/1 nodes are available: 1 Preemption is not helpful for scheduling.
❯ k get pvc
NAME         STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   VOLUMEATTRIBUTESCLASS   AGE
sample-pvc   Pending                                      local-path     <unset>                 21s
❯ k get pv
NAME         CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS     CLAIM                STORAGECLASS   VOLUMEATTRIBUTESCLASS   REASON   AGE
sample-pvc   1Gi        RWO            Delete           Released   default/sample-pvc   local-path     <unset>                          114s
```

**Expected behavior:**
Redeploying the same app should work, `pvc` should claim a volume and another fake `pv` should be created. 

**Actual behavior:**
I have to delete the app, delete the fake 'pv' and then redeploy it.

PR #511, opened 2025-10-16
Fix pseudo PV deletion

galal-hussein:
- #478 
This PR instructs the controller to delete the pseudo PV if the pod claiming it is deleted, allowing for recreation of the pod and reusing of the pvc

### DIFF
--- a/k3k-kubelet/controller/syncer/pod.go
+++ b/k3k-kubelet/controller/syncer/pod.go
@@ -211,5 +211,5 @@ func (r *PodReconciler) handlePodDeletion(ctx context.Context, pv *v1.Persistent
 		return err
 	}
 
-	return ctrlruntimeclient.IgnoreNotFound(r.VirtualClient.Delete(ctx, pv))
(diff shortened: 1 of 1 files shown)


k3k-kubelet/controller/syncer/pod.go
```diff
@@ -188,3 +194,22 @@ func (r *PodReconciler) pseudoPV(obj *v1.PersistentVolumeClaim) *v1.PersistentVo
...
+	pvPatch.Spec.ClaimRef = nil
+	pvPatch.Status.Phase = v1.VolumeReleased
+
+	controllerutil.RemoveFinalizer(pvPatch, "kubernetes.io/pv-protection")
+
+	if err := r.VirtualClient.Status().Update(ctx, pvPatch); err != nil {
+		return err
+	}
+
+	return ctrlruntimeclient.IgnoreNotFound(r.VirtualClient.Delete(ctx, pv))
```
enrichman: I am not sure if we should delete the `currentPV` instead, because it's the latest version we have:

```suggestion
	return ctrlruntimeclient.IgnoreNotFound(r.VirtualClient.Delete(ctx, currentPV))
```

PR #523, opened 2025-10-21
Fix for flaky test

enrichman:
When creating the Nginx Pod used in our tests we were not waiting for it to be in the Ready state.

This was causing an issue during the tests in virtual mode, because it's status was not being correctly checked.
This PR check the Pod status, and split the check to be more resilient.

It also has many changes because of a couple of import changes:
```
"k8s.io/api/core/v1":                      corev1 -> v1
"k8s.io/apimachinery/pkg/apis/meta/v1":    v1 -> metav1
```


jp-gouin (approved):
LGTM

PR #527, opened 2025-10-23
Cleanup of `customCAs` spec

enrichman:
Moving `customCAs` field to be a pointer. This will avoid creating an empty struct by default, when not needed:

```yaml
apiVersion: k3k.io/v1alpha1
kind: Cluster
metadata:
  name: foo
  namespace: k3k-foo
spec:
  customCAs:
    sources:
      clientCA: {}
      etcdPeerCA: {}
      etcdServerCA: {}
      requestHeaderCA: {}
      serverCA: {}
      serviceAccountToken: {}
```

Also moving the `enabled` to true as default, and the `secretName` as a required field.

This will result in not having the customCAs if not provided, and force the user providing the secretName.

This way the `validateCustomCACerts` check _should_ not be needed, but...

Result:

```yaml
apiVersion: k3k.io/v1alpha1
kind: Cluster
metadata:
  name: foo
  namespace: k3k-foo
spec:
  customCAs:
    enabled: true
    sources:
      clientCA:
        secretName: foo
      etcdPeerCA:
        secretName: foo
      etcdServerCA:
        secretName: foo
      requestHeaderCA:
        secretName: foo
      serverCA:
        secretName: foo
      serviceAccountToken:
        secretName: foo
```

### DIFF
--- a/charts/k3k/crds/k3k.io_clusters.yaml
+++ b/charts/k3k/crds/k3k.io_clusters.yaml
@@ -228,6 +228,7 @@ spec:
                   certificates.
                 properties:
                   enabled:
+                    default: true
                     description: Enabled toggles this feature on or off.
                     type: boolean
                   sources:
@@ -244,6 +245,8 @@ spec:
                               - For TLS pairs (e.g., ServerCA): 'tls.crt' and 'tls.key'.
                               - For ServiceAccountTokenKey: 'tls.key'.
                             type: string
+                        required:
+                        - secretName
                         type: object
                       etcdPeerCA:
                         description: ETCDPeerCA specifies the etcd-peer-ca cert/key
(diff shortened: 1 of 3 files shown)


pkg/controller/cluster/server/server.go
```diff
@@ -434,6 +434,10 @@ func (s *Server) setupStartCommand() (string, error) {
 }
 
 func (s *Server) loadCACertBundle(ctx context.Context) ([]v1.Volume, []v1.VolumeMount, error) {
+	if s.cluster.Spec.CustomCAs == nil {
+		return nil, nil, fmt.Errorf("customCAs not found")
+	}
```
pmatseykanets, line 439: Seems to be unnecessary given the conditional on L333.

enrichman: Yes, that's true. I just wanted to be extra safe in case someone calls this method and forgets to check that. :)

pkg/controller/cluster/server/server.go
pmatseykanets, line 454: This can be
```suggestion
		sortedCertIDs = slices.Sorted(maps.Keys(caCertMap))
```

pmatseykanets: On a different note is there any reason why it should be sorted? If yes is there any reason not to pre-sort it as it's hardcoded?

PR #534, opened 2025-10-28
Add Conformance Tests for virtual mode

### LINKED ISSUE #399: Conformance test for Virtual mode
The `virtual` mode cannot run on Github Actions within their runners, because they will not allow nested virtualization.

For the same reason these tests cannot run with dockerized environments.

To add these tests we will need to use self-hosted runners, or a different approach.

enrichman:
Fix #399

This PR adds the Conformance Tests CI for the virtual mode.

See https://github.com/cncf/k8s-conformance

A test result can be seen here: https://github.com/enrichman/k3k/actions/runs/18871535458

### DIFF
--- a/.github/workflows/test-conformance-virtual.yaml
+++ b/.github/workflows/test-conformance-virtual.yaml
@@ -1,0 +1,125 @@
+name: Conformance Tests - Virtual Mode
+
+on:
+  schedule:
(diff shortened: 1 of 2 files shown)


raulcabello (approved):
LGTM

PR #539, opened 2025-10-30
Improve tests resiliency

enrichman:
Now that is possible to run the tests against an external cluster we have found some issues.

I.e. we were making the assumption to have a single node clusters, or that we have always a "clean" installation.

This PR fixes the issues, cleaning up also some leftover namespaces, and improved the download of the coverage data.


bigkevmcd (approved):
Fine with me.

PR #540, opened 2025-10-31
Cleanup orphaned resources after Cluster deletion

### LINKED ISSUE #52: Cluster deletion
Deleting the virtual cluster through the UI will left several resources behind
- `PersistentVolumeClaims` if the virtual cluster is deployed using persistence
- `Secrets`, several secrets are created for the configuration of the virtual cluster and they remain present

Ideally, the cluster deletion in the UI should also delete the `clusters.k3k.io` in the downstream cluster and delegate this cleanup to the controller.

enrichman:
Fix https://github.com/rancher/virtual-clusters-ui/issues/52

During the deletion of a Virtual Cluster some resources were left orphaned in the namespace.
This was because during the sync we missed a couple of OwnerReference.

Checking also other resources we were not deleting the coordination Lease of the API server, even though it was not harmful.

This PR adds the controller reference for the garbage collection, and the deletion of the Lease, with a couple of tests.

### DIFF
--- a/pkg/controller/cluster/cluster_finalize.go
+++ b/pkg/controller/cluster/cluster_finalize.go
@@ -12,7 +12,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	coordinationv1 "k8s.io/api/coordination/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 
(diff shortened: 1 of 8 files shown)


pkg/controller/cluster/cluster_finalize.go
```diff
@@ -51,6 +53,21 @@ func (c *ClusterReconciler) finalizeCluster(ctx context.Context, cluster *v1beta
 		}
 	}
 
+	// delete API server lease
+	lease := &coordinationv1.Lease{
```
raulcabello, line 57: I'm missing context here. What is the purpose of this lease? Would it be possible to add an owner reference when creating it?

enrichman: Yes, it's something I didn't know, because it's something done automatically by Kubernetes.
We don't create that object, so we cannot add the reference: https://kubernetes.io/docs/concepts/architecture/leases/#api-server-identity

Basically is like a flag for the API servers, even from the doc it doesn't seems much of a use:

```
While not particularly useful on its own, this provides a mechanism for clients to discover
how many instances of kube-apiserver are operating the Kubernetes control plane.
Existence of kube-apiserver leases enables future capabilities that may require
coordination between each kube-apiserver.
```

I found this resource with this command:

```
-> % kubectl api-resources --verbs=list --namespaced -o name  | xargs -n 1 kubectl get --show-kind --ignore-not-found -n k3k-leftover
NAME                         DATA   AGE
configmap/kube-root-ca.crt   1      10m
NAME                                                           STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   VOLUMEATTRIBUTESCLASS   AGE
persistentvolumeclaim/varlibrancherk3s-k3k-leftover-server-0   Bound    pvc-272112f5-0961-45f7-9223-d320bbb8a28b   5Gi        RWO            local-path     <unset>                 8m54s
NAME                        READY   STATUS      RESTARTS   AGE
pod/import-leftover-54p84   0/1     Completed   0          8m54s
NAME                     SECRETS   AGE
serviceaccount/default   0         10m
NAME                                 HOLDER                                                            AGE
lease.coordination.k8s.io/leftover   k3k-leftover-kubelet-bsst8_49549c74-3e0f-4981-895a-074721953992   7m48s
```

so I just opted for deleting it. :grin:

raulcabello: ah ok! It makes sense now! Thanks for the info

raulcabello (approved):

pkg/controller/cluster/cluster_finalize.go
bigkevmcd: I think it will be automatically deleted?

https://kubernetes.io/docs/concepts/architecture/leases/

> Expired leases from kube-apiservers that no longer exist are garbage collected by new kube-apiservers after 1 hour.

bigkevmcd (approved):
I think you can use OwnerReferences and not ControllerReferences which would maybe simplify some code, but having an owner at all means they'll be garbage collected.

PR #552, opened 2025-11-10
Moved CRDs to Helm templates folder

enrichman:
Helm doesn't handle CRDs updates automatically. To handle this different strategies are possible. The simplest is to treat CRDs like normal resources, putting them in the `templates` folder, and adding the `helm.sh/resource-policy: keep` annotation, to prevent their deletion.

This is the same strategy adopted by `cert-manager` or `kubewarden`.

If someone is **upgrading** from a previous release the CRDs will not have the annotations used by Helm to track the managed resource, and the upgrade will fail. To prevent this the user should upgrade specifying the `--take-ownership` flag, or adding manually the needed labels and annotations:

```
# assuming the release name is "k3k" and the namespace "k3k-system"

kubectl label    crd clusters.k3k.io --overwrite "app.kubernetes.io/managed-by=Helm"
kubectl annotate crd clusters.k3k.io --overwrite "meta.helm.sh/release-name=k3k"
kubectl annotate crd clusters.k3k.io --overwrite "meta.helm.sh/release-namespace=k3k-system"

kubectl label    crd virtualclusterpolicies.k3k.io --overwrite "app.kubernetes.io/managed-by=Helm"
kubectl annotate crd virtualclusterpolicies.k3k.io --overwrite "meta.helm.sh/release-name=k3k"
kubectl annotate crd virtualclusterpolicies.k3k.io --overwrite "meta.helm.sh/release-namespace=k3k-system"
```

The PR also updates the generate script to output CRDs to the correct directory and include the keep resource policy annotation.

### DIFF
--- a/k3k-kubelet/controller/syncer/syncer_suite_test.go
+++ b/k3k-kubelet/controller/syncer/syncer_suite_test.go
@@ -92,7 +92,7 @@ func NewTestEnv() *TestEnv {
 	By("bootstrapping test environment")
 
 	testEnv := &envtest.Environment{
-		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "..", "charts", "k3k", "crds")},
+		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "..", "charts", "k3k", "templates", "crds")},
 		ErrorIfCRDPathMissing: true,
 		BinaryAssetsDirectory: tempDir,
 		Scheme:                buildScheme(),
--- a/pkg/controller/cluster/cluster_suite_test.go
+++ b/pkg/controller/cluster/cluster_suite_test.go
@@ -41,7 +41,7 @@ var (
 var _ = BeforeSuite(func() {
 	By("bootstrapping test environment")
 	testEnv = &envtest.Environment{
-		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "..", "charts", "k3k", "crds")},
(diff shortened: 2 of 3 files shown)


charts/k3k/templates/crds/k3k.io_clusters.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
+    helm.sh/resource-policy: keep
```
raulcabello, line 6: the CRD will never be deleted if a user uninstall the k3k chart. We shouldn't leave any resource if k3k is deleted

enrichman: Hm, I was using the same "safer" current approach. Right now since Helm is not touching the CRDs they will be there. `cert-manager` does the same. I'm not sure actually, without the controller it's not possible to update the Virtual Clusters, but they will keep running. If we remove the annotations they will be completely destroyed. This is a little bit safer, but I see the possible implication.

Maybe we should add some value to let the user install them, or remove the annotation?
Something like `--set crds.keep=false` to remove the keep annotation (even though I am not sure how to add it since they are generated).

raulcabello: I had a look, and this is also how it works for the crd folder in helm. So it should be fine

PR #565, opened 2025-11-14
Added labels and annotations flags to cluster and policy create

### LINKED ISSUE #355: add cluster labels and annotations from k3kcli
A handy feature would be to have an option to be able to add "--labels" during cluster creation from k3kcl

enrichman:
Fix #355 

There is sometimes the need to label and annotate the k3k resources.

This PR adds the `--labels` and `--annotations` flags to the `create cluster` and `create policy` commands.

### DIFF
--- a/cli/cmds/cluster_create.go
+++ b/cli/cmds/cluster_create.go
@@ -186,15 +186,12 @@ func createAction(appCtx *AppContext, config *CreateConfig) func(cmd *cobra.Comm
 }
 
 func newCluster(name, namespace string, config *CreateConfig) *v1beta1.Cluster {
-	labels := parseKeyValuePairs(config.labels, "label")
-	annotations := parseKeyValuePairs(config.annotations, "annotation")
-
 	cluster := &v1beta1.Cluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        name,
 			Namespace:   namespace,
-			Labels:      labels,
-			Annotations: annotations,
+			Labels:      parseKeyValuePairs(config.labels, "label"),
+			Annotations: parseKeyValuePairs(config.annotations, "annotation"),
 		},
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Cluster",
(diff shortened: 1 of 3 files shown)


cli/cmds/cluster_create.go
```diff
@@ -366,6 +370,27 @@ func caCertSecret(certName, clusterName, clusterNamespace string, cert, key []by
 	}
 }
 
+func parseKeyValuePairs(pairs []string, pairType string) map[string]string {
+	resultMap := make(map[string]string)
+
```
bigkevmcd, line 375: This could maybe be a bit more robust as it'll be receiving user data from the CLI?

Maybe if len(keyValue) != 2 then it's a problem, and don't limit the split?

enrichman: @bigkevmcd for the labels that would work, but annotations permits the value to contains `=` (or other chars).
Should I split the parsing logic in two? Not sure if there is a Kubernetes func that can be used to validate that.

Also an empty label, or empty annotation is still valid, so len could be 1.

bigkevmcd: @enrichman Good questions...I'm not sure...it feels like work for something that can be fixed with `kubectl label` guess :-)

I just tested this: 

```
$ kubectl label secret/my-test-secret example.com/test=`
secret/my-test-secret labeled
```

Then...

```
$ kubectl get secret/my-test-secret -o json | jq ".metadata.labels"
{
   "example.com/test": ""
}
```

So...I guess that behaviour is desirable 😄

bigkevmcd (approved):

cli/cmds/cluster_create.go
enrichman: Yep, I've just checked the [doc](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) and it seems to be documented as well:

> Valid label value:
> - must be 63 characters or less (**can be empty**),
> - unless empty, must begin and end with an alphanumeric character ([a-z0-9A-Z]),
> - could contain dashes (-), underscores (_), dots (.), and alphanumerics between.

Thank you!

PR #570, opened 2025-11-21
SecretMounts feature and private registries

galal-hussein:
The PR Adds `secretMounts` field to the Cluster spec, allowing users to mount secrets into server and agent pods, it can be used with:

  - Private registry configuration (registries.yaml)
  -  Addon manifests
  - Any other files needed by k3s
  
```go
  type SecretMount struct {
      SecretName  string            `json:"secretName,omitempty"`
      MountPath   string            `json:"mountPath,omitempty"`
      SubPath     string            `json:"subPath,omitempty"`
      KeysToPaths []v1.KeyToPath    `json:"keysToPaths,omitempty"`
  }
```
### Features

  - Mount secrets as directories: Mount entire secrets to a directory path
  - Mount secrets as files with `subPath`: Mount individual files while keeping parent directories writable (required for paths like `/var/lib/rancher/k3s/server/manifests/` where k3s needs write access)
  - `KeysToPaths` support: Map specific secret keys to custom file paths
  - Applied to both server and agent pods: Secrets are mounted consistently across all cluster pods

 
#### Private Registry:
  ```yaml
spec:
    secretMounts:
      - secretName: my-registry-config
        mountPath: /etc/rancher/k3s/registries.yaml
        subPath: registries.yaml
      - secretName: my-registry-ca
        mountPath: /etc/rancher/k3s/tls/ca.crt
        subPath: ca.crt
```

#### Addon Manifests:
```yaml
spec:
    secretMounts:
      - secretName: my-addons
        mountPath: /var/lib/rancher/k3s/server/manifests/nginx.yaml
        subPath: nginx.yaml
```
  Other Changes

  - Fix a longstanding bug with Addons spec
  - Added e2e tests for private registry configuration
  - Added e2e tests for addon deployment via secretMounts
  - Added e2e tests for addon deployment via addon spec

### DIFF
--- a/charts/k3k/templates/crds/k3k.io_clusters.yaml
+++ b/charts/k3k/templates/crds/k3k.io_clusters.yaml
@@ -241,10 +241,9 @@ spec:
                         properties:
                           secretName:
                             description: |-
-                              SecretName specifies the name of an existing secret to use.
-                              The custom certificate controller expects specific keys inside based on the credential type:
-                              - For TLS pairs (e.g., ServerCA): 'tls.crt' and 'tls.key'.
-                              - For ServiceAccountTokenKey: 'tls.key'.
+                              The secret must contain specific keys based on the credential type:
+                              - For TLS certificate pairs (e.g., ServerCA): `tls.crt` and `tls.key`.
+                              - For the ServiceAccountToken signing key: `tls.key`.
                             type: string
                         required:
                         - secretName
@@ -255,10 +254,9 @@ spec:
                         properties:
                           secretName:
                             description: |-
-                              SecretName specifies the name of an existing secret to use.
-                              The custom certificate controller expects specific keys inside based on the credential type:
-                              - For TLS pairs (e.g., ServerCA): 'tls.crt' and 'tls.key'.
-                              - For ServiceAccountTokenKey: 'tls.key'.
+                              The secret must contain specific keys based on the credential type:
+                              - For TLS certificate pairs (e.g., ServerCA): `tls.crt` and `tls.key`.
+                              - For the ServiceAccountToken signing key: `tls.key`.
                             type: string
                         required:
                         - secretName
@@ -269,10 +267,9 @@ spec:
                         properties:
                           secretName:
                             description: |-
-                              SecretName specifies the name of an existing secret to use.
-                              The custom certificate controller expects specific keys inside based on the credential type:
-                              - For TLS pairs (e.g., ServerCA): 'tls.crt' and 'tls.key'.
-                              - For ServiceAccountTokenKey: 'tls.key'.
+                              The secret must contain specific keys based on the credential type:
+                              - For TLS certificate pairs (e.g., ServerCA): `tls.crt` and `tls.key`.
+                              - For the ServiceAccountToken signing key: `tls.key`.
                             type: string
                         required:
                         - secretName
@@ -283,10 +280,9 @@ spec:
                         properties:
                           secretName:
                             description: |-
-                              SecretName specifies the name of an existing secret to use.
-                              The custom certificate controller expects specific keys inside based on the credential type:
-                              - For TLS pairs (e.g., ServerCA): 'tls.crt' and 'tls.key'.
-                              - For ServiceAccountTokenKey: 'tls.key'.
+                              The secret must contain specific keys based on the credential type:
+                              - For TLS certificate pairs (e.g., ServerCA): `tls.crt` and `tls.key`.
+                              - For the ServiceAccountToken signing key: `tls.key`.
                             type: string
                         required:
                         - secretName
@@ -296,10 +292,9 @@ spec:
                         properties:
                           secretName:
                             description: |-
-                              SecretName specifies the name of an existing secret to use.
-                              The custom certificate controller expects specific keys inside based on the credential type:
-                              - For TLS pairs (e.g., ServerCA): 'tls.crt' and 'tls.key'.
-                              - For ServiceAccountTokenKey: 'tls.key'.
+                              The secret must contain specific keys based on the credential type:
+                              - For TLS certificate pairs (e.g., ServerCA): `tls.crt` and `tls.key`.
+                              - For the ServiceAccountToken signing key: `tls.key`.
                             type: string
                         required:
                         - secretName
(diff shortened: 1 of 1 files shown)


pkg/controller/cluster/agent/virtual.go
```diff
@@ -98,6 +99,15 @@ func (v *VirtualAgent) deployment(ctx context.Context) error {
 			"mode":    "virtual",
 		},
 	}
+	podSpec := v.podSpec(image, name, v.cluster.Spec.AgentArgs, &selector)
```
enrichman: I know that this unrelated but since this is a method of `v` we don't need to pass the `v.cluster.Spec.AgentArgs`.

pkg/controller/cluster/mounts/mounts.go
```diff
@@ -0,0 +1,49 @@
...
+		volMounts           []v1.VolumeMount
+		projectedVolSources []v1.VolumeProjection
+	)
+
+	for _, secretMount := range secretMounts {
+		if secretMount.SecretName == "" || secretMount.MountPath == "" {
+			continue
+		}
+
+		projectedVolSources = []v1.VolumeProjection{
```
enrichman: NIT

```suggestion
	)

	for _, secretMount := range secretMounts {
		if secretMount.SecretName == "" || secretMount.MountPath == "" {
			continue
		}

		projectedVolSources := []v1.VolumeProjection{
```

pkg/controller/cluster/mounts/mounts_test.go
```diff
@@ -0,0 +1,337 @@
...
+				if a.Name < b.Name {
+					return -1
+				}
+
+				if a.Name > b.Name {
+					return 1
+				}
+
+				return 0
+			})
```
enrichman: I think this can/should removed. We should check that the order is respected.

pkg/controller/cluster/server/server.go
```diff
@@ -545,3 +577,15 @@ func sortedKeys(keyMap map[string]string) []string {
...
+func sortedAddons(addons []v1beta1.Addon) []v1beta1.Addon {
+	sorted := make([]v1beta1.Addon, len(addons))
+	copy(sorted, addons)
+
+	slices.SortFunc(sorted, func(a, b v1beta1.Addon) int {
+		return strings.Compare(a.SecretRef, b.SecretRef)
+	})
+
+	return sorted
+}
```
enrichman: I am not sure we should sort them.

tests/cluster_addons_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+	It("will load the addon manifest in server pod and deploys the pod", func() {
+		ctx := context.Background()
+
+		labelSelector := "cluster=" + virtualCluster.Cluster.Name + ",role=server"
+		serverPods, err := k8s.CoreV1().Pods(virtualCluster.Cluster.Namespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+		Expect(err).To(Not(HaveOccurred()))
+
+		Expect(len(serverPods.Items)).To(Equal(1))
+		serverPod := serverPods.Items[0]
```
enrichman: This fetch is repeated many times, and we possibly need it for other reasons. It should be part of some util func along within the check of the server/agent status.

pkg/controller/cluster/mounts/mounts_test.go
galal-hussein: it wasnt as a matter of fact, I think I will re-add the sort for the main logic

pkg/controller/cluster/server/server.go
```diff
@@ -533,6 +498,71 @@ func (s *Server) mountCACert(volumeName, certName, secretName string, subPathMou
...
+					Namespace: s.cluster.Namespace,
+				},
+				Data: addons.Data,
+			}
+
+			if err := s.client.Create(ctx, &clusterAddons); err != nil {
+				if !apierrors.IsAlreadyExists(err) {
+					return nil, nil, err
+				}
+			}
```
enrichman, line 540: I have not tested this, but I think we are not cleaning up the secret addons. I would also an OwnerRef, wdyt?

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -185,6 +185,35 @@ type ClusterSpec struct {
...
+	// +optional
+	MountPath string `json:"mountPath,omitempty"`
+	// SubPath is an optional path within the secret to mount instead of the root.
+	// When specified, only the specified key from the secret will be mounted as a file
+	// at MountPath, keeping the parent directory writable.
+	//
+	// +optional
+	SubPath string `json:"subPath,omitempty"`
+	// Role is the type of the k3k pod that will be used to mount the secret.
+	// This can be sever, or agent, or both.
```
```suggestion
	// This can be 'server', 'agent', or 'all' (for both).
```

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -185,6 +185,35 @@ type ClusterSpec struct {
...
 	Sync *SyncConfig `json:"sync,omitempty"`
+
+	// SecretMounts specifies a list of secrets to mount into server and agent pods.
+	// Each entry defines a secret and its mount path within the pods.
+	//
+	// +optional
+	SecretMounts []SecretMount `json:"secretMounts,omitempty"`
+}
+
+// SecretMount represent any extra mount that the user will specify
```
```suggestion
// SecretMount defines a secret to be mounted into server or agent pods, allowing for custom configurations, certificates, or other sensitive data.
```

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -471,7 +500,7 @@ type CredentialSources struct {
 // It can represent either a TLS key pair or a single private key.
 type CredentialSource struct {
 	// SecretName specifies the name of an existing secret to use.
-	// The controller expects specific keys inside based on the credential type:
+	// The custom certificate controller expects specific keys inside based on the credential type:
 	// - For TLS pairs (e.g., ServerCA): 'tls.crt' and 'tls.key'.
 	// - For ServiceAccountTokenKey: 'tls.key'.
```
```suggestion
	// The secret must contain specific keys based on the credential type:
	// - For TLS certificate pairs (e.g., ServerCA): `tls.crt` and `tls.key`.
	// - For the ServiceAccountToken signing key: `tls.key`.
```

PR #571, opened 2025-11-24
Disable network-policy-controller for older versions of k3s

galal-hussein:
The PR adds `--disable-network-policy` by default for clusters that has k3s versions older than the following patch releases:

- v1.31.14+k3s1
- v1.32.10+k3s1
- v1.33.6+k3s1
- v1.34.2+k3s1

- #477

### DIFF
--- a/pkg/controller/cluster/server/config.go
+++ b/pkg/controller/cluster/server/config.go
@@ -18,7 +18,7 @@ import (
 var k3sNetpolVersions = []string{"v1.31.14", "v1.32.10", "v1.33.6", "v1.34.2"}
 
 func (s *Server) Config(init bool, serviceIP string) (*v1.Secret, error) {
-	name := configSecretName(s.cluster.Name, init)
+	name := ConfigSecretName(s.cluster.Name, init)
 
 	sans := sets.NewString(s.cluster.Spec.TLSSANs...)
 	sans.Insert(
@@ -97,7 +97,7 @@ func serverOptions(cluster *v1beta1.Cluster, token string) string {
 	return opts
 }
 
-func configSecretName(clusterName string, init bool) string {
+func ConfigSecretName(clusterName string, init bool) string {
 	if !init {
 		return controller.SafeConcatNameWithPrefix(clusterName, configName)
 	}
--- a/pkg/controller/cluster/server/server.go
+++ b/pkg/controller/cluster/server/server.go
@@ -63,7 +63,7 @@ func (s *Server) podSpec(image, name string, persistent bool, startupCmd string)
 				Name: "initconfig",
 				VolumeSource: v1.VolumeSource{
 					Secret: &v1.SecretVolumeSource{
-						SecretName: configSecretName(s.cluster.Name, true),
+						SecretName: ConfigSecretName(s.cluster.Name, true),
 						Items: []v1.KeyToPath{
 							{
 								Key:  "config.yaml",
@@ -77,7 +77,7 @@ func (s *Server) podSpec(image, name string, persistent bool, startupCmd string)
 				Name: "config",
 				VolumeSource: v1.VolumeSource{
 					Secret: &v1.SecretVolumeSource{
-						SecretName: configSecretName(s.cluster.Name, false),
+						SecretName: ConfigSecretName(s.cluster.Name, false),
 						Items: []v1.KeyToPath{
 							{
 								Key:  "config.yaml",
--- a/tests/cluster_update_test.go
+++ b/tests/cluster_update_test.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kubernetes/pkg/api/v1/pod"
 	"k8s.io/utils/ptr"
 
@@ -13,6 +14,7 @@ import (
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/rancher/k3k/pkg/apis/k3k.io/v1beta1"
+	"github.com/rancher/k3k/pkg/controller/cluster/server"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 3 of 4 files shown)


pkg/controller/cluster/server/config.go
```diff
@@ -81,17 +85,47 @@ func serverOptions(cluster *v1beta1.Cluster, token string) string {
...
 }
+
+func requiresNetworkPolicyDisable(cluster *v1beta1.Cluster) bool {
+	imageVersion := "latest"
+
+	if cluster.Spec.Version != "" {
+		imageVersion = cluster.Spec.Version
+	} else if cluster.Status.HostVersion != "" {
+		imageVersion = cluster.Status.HostVersion
+	}
```
enrichman, line 115: I would try to avoid copy-pasting the same check: if we need to change it, or fix something we could miss one of them.

I would create a `K3sVersion` along with the `K3sImage`, so we can call directly:

```go
func K3SImage(cluster *v1beta1.Cluster, k3SImage string) string {
	return k3SImage + ":" + K3SVersion(cluster)
}

func K3SVersion(cluster *v1beta1.Cluster) string {
	imageVersion := "latest"

	if cluster.Spec.Version != "" {
		imageVersion = cluster.Spec.Version
	} else if cluster.Status.HostVersion != "" {
		imageVersion = cluster.Status.HostVersion
	}

	return imageVersion
}
```

pkg/controller/cluster/server/config.go
```diff
@@ -13,8 +15,10 @@ import (
 	"github.com/rancher/k3k/pkg/controller/cluster/agent"
 )
 
+var k3sNetpolVersions = []string{"v1.31.14", "v1.32.10", "v1.33.6", "v1.34.2"}
```
enrichman, line 18: I would keep this declaration inside the func, so it's easier to see which are the versions that have the fix.

pkg/controller/cluster/server/config.go
```diff
@@ -81,17 +85,47 @@ func serverOptions(cluster *v1beta1.Cluster, token string) string {
...
+		imageVersion = cluster.Spec.Version
+	} else if cluster.Status.HostVersion != "" {
+		imageVersion = cluster.Status.HostVersion
+	}
+
+	if imageVersion == "latest" {
+		return false
+	}
+
+	for _, fixedVersion := range k3sNetpolVersions {
```
enrichman: We are missing the check for **newer** versions (`>= v1.35`).

pkg/controller/cluster/server/config.go
```diff
@@ -81,17 +85,47 @@ func serverOptions(cluster *v1beta1.Cluster, token string) string {
...
+		if semver.MajorMinor(imageVersion) == semver.MajorMinor(fixedVersion) {
+			// if the major and minor match then check if the patch version is less than fixed version
+			version, _ := strings.CutSuffix(imageVersion, "-k3s1")
```
enrichman, line 124: I am not familiar with the k3s release naming convention but it looks it happens to have also the `k3s2` metadata ([`v1.29.6+k3s2`](https://github.com/k3s-io/k3s/releases/tag/v1.29.6%2Bk3s2)). In those cases this is not going to work. I would relax the check with a simpler `Cut`:

```suggestion
			version, _, _ := strings.Cut(imageVersion, "-")
```

The `semver` package has also a `Canonical` func, but unfortunately it's not working only with the build metadata. The `-` is used to mark a prerelease instead.
```go
semver.Canonical("v1.33.3-k3s1") // v1.33.3-k3s1
semver.Canonical("v1.33.3+k3s1") // v1.33.3
```

pkg/controller/cluster/server/config.go
```diff
@@ -81,17 +85,47 @@ func serverOptions(cluster *v1beta1.Cluster, token string) string {
...
-func configSecretName(clusterName string, init bool) string {
+func ConfigSecretName(clusterName string, init bool) string {
 	if !init {
 		return controller.SafeConcatNameWithPrefix(clusterName, configName)
 	}
 
 	return controller.SafeConcatNameWithPrefix(clusterName, initConfigName)
 }
+
+func requiresNetworkPolicyDisable(cluster *v1beta1.Cluster) bool {
```
enrichman, line 108: Since the flag is called `disable-network-policy` I would just use the same name:

```suggestion
func disableNetworkPolicy(cluster *v1beta1.Cluster) bool {
```

pkg/controller/cluster/server/config.go
```diff
@@ -81,17 +85,47 @@ func serverOptions(cluster *v1beta1.Cluster, token string) string {
...
+	for _, fixedVersion := range k3sNetpolVersions {
+		if semver.MajorMinor(imageVersion) == semver.MajorMinor(fixedVersion) {
+			// if the major and minor match then check if the patch version is less than fixed version
+			version, _ := strings.CutSuffix(imageVersion, "-k3s1")
+
+			return semver.Compare(version, fixedVersion) == -1
+		}
+	}
+
+	return false
```
enrichman, line 130: What it the k3s version is lower (i.e. `1.30`)? I think we are missing this check.

enrichman (changes requested):
I have not tested but I was thinking we can use a map for a more direct approach, instead of the loop, wdyt?

```go
majorMinor := semver.MajorMinor(imageVersion)

// for latest releases, or >1.35 we don't need to disable the network policy
if imageVersion == "latest" || semver.Compare(majorMinor, "v1.35") >= 0 {
	return false
}

// for 1.31-1.34 we need to check the patch release
version, _, _ := strings.Cut(imageVersion, "-")

fixedVersion := map[string]string{
	"v1.31": "v1.31.14",
	"v1.32": "v1.32.10",
	"v1.33":  "v1.33.6",
	"v1.34":  "v1.34.2",
}

// if the version is lower we return true to disable the network policy
// Note: if the majorMinor is not found the check will return true
return semver.Compare(version, fixedVersion[majorMinor]) < 0
```

Please add some **unit** tests for the `disableNetworkPolicy`, they are easy/cheap, and perfect for these kind of func. I would also change the signature of the func to use just the version, so it's even easier:

```go
k3sVersion := controller.K3sVersion(cluster)
if disableNetworkPolicy(k3sVersion) {
	opts = opts + "disable-network-policy: true\n"
}
```

PR #573, opened 2025-11-25
Fixed CreatePod and UpdatePod in Virtual Kubelet for Downward API

### LINKED ISSUE #495: Permission issue with ingress installation
**Environmental Info:**
Host Cluster Version: 
v1.33.5+k3s1

openSUSE Linux, x86

Host Cluster Configuration:
3 servers

K3K Cluster Configuration:
Everything default, shared cluster

**Describe the bug:**
The logs for `ingress-nginx-admission-patch-97zvq` show:
```
W1007 16:42:21.407074       1 client_config.go:667] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
{"err":"secrets \"ingress-nginx-admission\" is forbidden: User \"system:serviceaccount:ingress-nginx:ingress-nginx-admission\" cannot get resource \"secrets\" in API group \"\" in the namespace \"k3k-myvcluster\"","level":"fatal","msg":"error getting secret","source":"k8s/k8s.go:232","time":"2025-10-07T16:42:21Z"}
```

**Steps To Reproduce:**
Ran: `kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.13.3/deploy/static/provider/cloud/deploy.yaml`

**Expected behavior:**
For the nginx ingress controller to install properly.

**Actual behavior:**
The admission patch pods enter a crashloop.

enrichman:
Fix #495.

Between `v0.3.4` and `v.0.3.5` we introduced a regression. In shared mode when a Pod is created on the Host cluster the `fieldRef.fieldPath` were not hardcoded anymore, keeping the reference. This means that, for example, with `metadata.namespace` the Pod gets a reference on the real namespace is running on:

```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  namespace: default
spec:
  containers:
  - image: nginx
    name: nginx
    env:
    - name: POD_NAME
      valueFrom:
        fieldRef:
          fieldPath: metadata.name
    - name: POD_NAMESPACE
      valueFrom:
        fieldRef:
          fieldPath: metadata.namespace
```

The `$POD_NAMESPACE` is evaluated as `k3k-mycluster` for example, instead of the `default` where the Pod was scheduled instead (in the virtual cluster).

To avoid this issue we need to hardcode the value that are not reflecting the virtual cluster values, such as the metadata.name and the metadata.namespace. All the other values of the [downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/) should be ok.

```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  namespace: default
spec:
  containers:
  - image: nginx
    name: nginx
    env:
    - name: POD_NAME
      value: nginx
    - name: POD_NAMESPACE
      value: default
```

Some tests were added (e2e and unit) and also a test that will install the ingress-nginx controller.

This PR also adds the support for the `configMapKeyRef` and `secretkeyref`, that should match the translated name in the host.

### DIFF
--- a/tests/cluster_pod_test.go
+++ b/tests/cluster_pod_test.go
@@ -222,7 +222,7 @@ var _ = Context("In a shared cluster", Label(e2eTestLabel), Ordered, func() {
 
 			Expect(os.WriteFile("vk-kubeconfig.yaml", virtualCluster.Kubeconfig, 0o644)).To(Succeed())
 
-			ingressNginx := "https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.12.0/deploy/static/provider/baremetal/deploy.yaml"
+			ingressNginx := "testdata/resources/ingress-nginx-v1.14.1.yaml"
 			cmd := exec.Command("kubectl", "apply", "--kubeconfig", "vk-kubeconfig.yaml", "-f", ingressNginx)
 			output, err := cmd.CombinedOutput()
 
--- a/tests/testdata/resources/ingress-nginx-v1.14.1.yaml
+++ b/tests/testdata/resources/ingress-nginx-v1.14.1.yaml
@@ -1,0 +1,668 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    app.kubernetes.io/instance: ingress-nginx
+    app.kubernetes.io/name: ingress-nginx
+  name: ingress-nginx
+---
+apiVersion: v1
+automountServiceAccountToken: true
+kind: ServiceAccount
+metadata:
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/instance: ingress-nginx
+    app.kubernetes.io/name: ingress-nginx
+    app.kubernetes.io/part-of: ingress-nginx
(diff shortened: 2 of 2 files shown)


tests/cluster_pod_test.go
```diff
@@ -15,162 +19,263 @@ import (
...
+		})
+	})
+
+	When("installing the nginx-ingress controller", func() {
+		BeforeAll(func() {
+			By("installing the nginx-ingress controller")
+
+			Expect(os.WriteFile("vk-kubeconfig.yaml", virtualCluster.Kubeconfig, 0o644)).To(Succeed())
+
+			ingressNginx := "https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.12.0/deploy/static/provider/baremetal/deploy.yaml"
```
galal-hussein: Do we need to pull the file from Github?, I think we can store a local copy of the yaml in our testdata to be consistent with other tests

enrichman: Hm, that's a good tip. I will store it locally. :v: 
It's also better to avoid any issues with the network.

galal-hussein (approved):

Issue #574, opened 2025-11-25
BUG Issue with PVC in `shared` mode


jp-gouin:
**Environmental Info:**
Host Cluster Version: 
K3S v1.33.5+k3s1

Host Cluster Configuration:
Single node cluster

K3K Cluster Configuration:
1 server `shared` mode virtual cluster
K3k v1.0.0

**Describe the bug:**
When deploying an application with PVC, the application is not scheduled and the PVC shows 
```
Bound claim has lost reference to PersistentVolume. Data on the volume is lost!
``` 

**Steps To Reproduce:**
On the virtual cluster, apply the following app
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: sample-pvc
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: rancher-env-app
spec:
  replicas: 2
  selector:
    matchLabels:
      app: rancher-env-app
  template:
    metadata:
      labels:
        app: rancher-env-app
    spec:
      containers:
      - name: rancher-env-app
        image: jpgouin/suse-demo
        imagePullPolicy: Always
        ports:
        - containerPort: 8080
        env:
        - name: NODE_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
        - name: POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
        volumeMounts:
        - name: data-volume
          mountPath: /data
      volumes:
      - name: data-volume
        persistentVolumeClaim:
          claimName: sample-pvc
```

**Expected behavior:**
The application is scheduled, the PVC in the virtual cluster shows the bound status. 

**Actual behavior:**
The application failed to scheduled with 
```
0/1 nodes are available: persistentvolumeclaim "sample-pvc" bound to
        non-existent persistentvolume "". preemption: 0/1 nodes are available: 1
        Preemption is not helpful for scheduling.
```

aganesh-suse:
### Validated with K3KCLI_VERSION="v1.0.1"
  
## Environment Details

  
*Infrastructure*
- [x] Cloud
- [ ] Hosted
  
*Node(s) CPU architecture, OS, and Version:*

```
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 24.04 LTS"

$ uname -m
x86_64
```
  
*Cluster Configuration:*

```
1 server
```

[...]

PR #577, opened 2025-11-25
Attach creation of Pseudo PV to the PVC instead of the pods

galal-hussein:
This PR will change how we create pseudo (fake) PVs to allow the scheduling of the pods that claims a PVC, instead of the previous behavior of creating a fake PV only when a consumer Pod is created, we now will create/delete the fake PV whenever the PVC is created/delete on the virtual cluster, this will eliminate errors/bugs like scaling down deployments and random deletion of the pods in virtual cluster.

- #574

### DIFF
--- a/k3k-kubelet/controller/syncer/persistentvolumeclaims.go
+++ b/k3k-kubelet/controller/syncer/persistentvolumeclaims.go
@@ -158,8 +158,8 @@ func (r *PVCReconciler) createVirtualPersistentVolume(ctx context.Context, pvc v
 
 	pv := newPersistentVolume(&pvc)
 
-	if err := r.VirtualClient.Create(ctx, pv); err != nil {
-		return ctrlruntimeclient.IgnoreAlreadyExists(err)
+	if err := r.VirtualClient.Create(ctx, pv); err != nil && !apierrors.IsAlreadyExists(err) {
+		return err
 	}
 
 	orig := pv.DeepCopy()


k3k-kubelet/controller/syncer/persistentvolumeclaims.go
```diff
@@ -103,8 +107,14 @@ func (r *PVCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	if !virtPVC.DeletionTimestamp.IsZero() {
 		// deleting the synced pvc if exists
 		if err := r.HostClient.Delete(ctx, syncedPVC); err != nil && !apierrors.IsNotFound(err) {
+			return reconcile.Result{}, ctrlruntimeclient.IgnoreNotFound(err)
```
enrichman: Since we are already checking the `IsNotFound` error this should not be needed:

```suggestion
			return reconcile.Result{}, err
```

k3k-kubelet/controller/syncer/persistentvolumeclaims.go
```diff
@@ -127,7 +137,13 @@ func (r *PVCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 
 	// note that we dont need to update the PVC on the host cluster, only syncing the PVC to allow being
 	// handled by the host cluster.
-	return reconcile.Result{}, ctrlruntimeclient.IgnoreAlreadyExists(r.HostClient.Create(ctx, syncedPVC))
+	if err := ctrlruntimeclient.IgnoreAlreadyExists(r.HostClient.Create(ctx, syncedPVC)); err != nil {
+		return reconcile.Result{}, err
+	}
```
enrichman: NIT: just to follow the same pattern as before we can check for the error type:

```suggestion
	if err := r.HostClient.Create(ctx, syncedPVC); err != nil && !apierrors.IsAlreadyExists(err) {
		return reconcile.Result{}, err
	}
```

k3k-kubelet/controller/syncer/persistentvolumeclaims.go
```diff
@@ -136,3 +152,81 @@ func (r *PVCReconciler) pvc(obj *v1.PersistentVolumeClaim) *v1.PersistentVolumeC
...
+
+	pvcPatch.Annotations[volume.AnnBoundByController] = "yes"
+	pvcPatch.Annotations[volume.AnnBindCompleted] = "yes"
+	pvcPatch.Status.Phase = v1.ClaimBound
+	pvcPatch.Status.AccessModes = pvcPatch.Spec.AccessModes
+
+	return r.VirtualClient.Status().Update(ctx, pvcPatch)
+}
+
+func (r *PVCReconciler) pv(obj *v1.PersistentVolumeClaim) *v1.PersistentVolume {
```
enrichman: I don't think this needs to be a method, but it can be just a func. Also I would use a more descriptive name for it (`newPersistentVolume`, or `persistentVolume`), so we could eventually use `pv` for variables.


```go
// newPersistentVolume creates a PersistentVolume with a ClaimRef to the provided PersistentVolumeClaim
func newPersistentVolume(pvc *v1.PersistentVolumeClaim) *v1.PersistentVolume {
    // [...]
}

// pv := newPersistentVolume(pvc)
```

k3k-kubelet/controller/syncer/persistentvolumeclaims.go
```diff
@@ -136,3 +152,81 @@ func (r *PVCReconciler) pvc(obj *v1.PersistentVolumeClaim) *v1.PersistentVolumeC
 
 	return hostPVC
 }
+
+func (r *PVCReconciler) pseudoPV(ctx context.Context, pvc v1.PersistentVolumeClaim, log logr.Logger) error {
+	log.Info("Creating pseudo Persistent Volume")
```
enrichman: We already have the logger in the context, so I would try to avoid passing it again. Also to avoid too many logs I would try to stick with the `Info` only for the final outcome of the reconciliation (`PersistentVolumeClaim created/updated in the host/virtual cluster`) and keep everything else at a Debug level `V(1)`.

i.e.:

```go
func (r *PVCReconciler) pseudoPV(ctx context.Context, pvc v1.PersistentVolumeClaim) error {
	log := ctrl.LoggerFrom(ctx)
	log.V(1).Info("Creating pseudo Persistent Volume")

	// [...]
}
```

k3k-kubelet/controller/syncer/persistentvolumeclaims.go
```diff
@@ -136,3 +152,81 @@ func (r *PVCReconciler) pvc(obj *v1.PersistentVolumeClaim) *v1.PersistentVolumeC
 
 	return hostPVC
 }
+
+func (r *PVCReconciler) pseudoPV(ctx context.Context, pvc v1.PersistentVolumeClaim, log logr.Logger) error {
```
enrichman: I think we should rename this method to something more explicit, like `createVirtualPersistentVolume`, and add a nice and deep comment on why this is needed.

k3k-kubelet/controller/syncer/persistentvolumeclaims.go
```diff
@@ -136,3 +152,81 @@ func (r *PVCReconciler) pvc(obj *v1.PersistentVolumeClaim) *v1.PersistentVolumeC
...
+	orig := pv.DeepCopy()
+	pv.Status = v1.PersistentVolumeStatus{
+		Phase: v1.VolumeBound,
+	}
+
+	if err := r.VirtualClient.Status().Patch(ctx, pv, ctrlruntimeclient.MergeFrom(orig)); err != nil {
+		return err
+	}
+
+	log.Info("Patch the status of PersistentVolumeClaim to Bound")
```
enrichman: Same as before (`debug`).

enrichman (approved):
LGTM, thanks!

Issue #544, opened 2025-11-04
Add update option to the cli


jp-gouin:
The cli should have a `cluster update` option.
It should provide scaling up and down server and agent
it should provide k3s version update with a validation to block downgrades.

jp-gouin:
Great thanks @mdrahman-suse

PR #595, opened 2025-12-18
CLI cluster update

galal-hussein:
The PR adds cli `cluster update` command to k3kcli, the cluster cli can update the following:

```
Spec.Servers
Spec.Agents
Spec.Version
Labels
Annotations
```

- #544

### DIFF
--- a/docs/cli/k3kcli_cluster_update.md
+++ b/docs/cli/k3kcli_cluster_update.md
@@ -15,16 +15,14 @@ k3kcli cluster update [command options] NAME
 ### Options
 
 ```
-      --agents int32               number of agents
-      --annotations stringArray    Annotations to add to the cluster object (e.g. key=value)
-  -h, --help                       help for update
-      --kubeconfig-server string   Override the kubeconfig server host
-      --labels stringArray         Labels to add to the cluster object (e.g. key=value)
-  -n, --namespace string           namespace of the k3k cluster
-  -y, --no-confirm                 Skip interactive approval before applying update
-      --servers int32              number of servers
-      --timeout duration           The timeout for waiting for the cluster to become ready (e.g., 10s, 5m, 1h). (default 3m0s)
-      --version string             k3s version
+      --agents int32              number of agents
+      --annotations stringArray   Annotations to add to the cluster object (e.g. key=value)
+  -h, --help                      help for update
+      --labels stringArray        Labels to add to the cluster object (e.g. key=value)
+  -n, --namespace string          namespace of the k3k cluster
+  -y, --no-confirm                Skip interactive approval before applying update
+      --servers int32             number of servers (default 1)
+      --version string            k3s version
 ```
 
 ### Options inherited from parent commands
--- a/docs/cli/k3kcli.adoc
+++ b/docs/cli/k3kcli.adoc
@@ -150,16 +150,14 @@ k3kcli cluster update [command options] NAME
 === Options
 
 ----
-      --agents int32               number of agents
-      --annotations stringArray    Annotations to add to the cluster object (e.g. key=value)
-  -h, --help                       help for update
-      --kubeconfig-server string   Override the kubeconfig server host
-      --labels stringArray         Labels to add to the cluster object (e.g. key=value)
-  -n, --namespace string           namespace of the k3k cluster
-  -y, --no-confirm                 Skip interactive approval before applying update
-      --servers int32              number of servers
-      --timeout duration           The timeout for waiting for the cluster to become ready (e.g., 10s, 5m, 1h). (default 3m0s)
-      --version string             k3s version
+      --agents int32              number of agents
+      --annotations stringArray   Annotations to add to the cluster object (e.g. key=value)
+  -h, --help                      help for update
+      --labels stringArray        Labels to add to the cluster object (e.g. key=value)
+  -n, --namespace string          namespace of the k3k cluster
+  -y, --no-confirm                Skip interactive approval before applying update
+      --servers int32             number of servers (default 1)
+      --version string            k3s version
 ----
 
 === Options inherited from parent commands


cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,166 @@
...
+
+		logrus.Infof("Updating cluster '%s' in namespace '%s'", name, namespace)
+
+		virtualCluster.Spec.Servers = ptr.To(int32(config.servers))
+		virtualCluster.Spec.Agents = ptr.To(int32(config.agents))
+		virtualCluster.Spec.ServerArgs = config.serverArgs
+		virtualCluster.Spec.AgentArgs = config.agentArgs
+		virtualCluster.Spec.Version = config.version
+		virtualCluster.Labels = parseKeyValuePairs(config.labels, "label")
+		virtualCluster.Annotations = parseKeyValuePairs(config.annotations, "annotation")
```
enrichman: Hm, I think this needs some clarification from @jp-gouin.

With this approach we are going to overwrite everything. What if the user provided some values? They will be cleared. Even if we updated a different field, or am I wrong?

Imo if a user haven't provided any `--labels` we should keep the existing, and if the user provided an already existing field we should prevent the **overwrite** of it, unless an `--overwrite` flag is used. Also we should probably warn what we are going to remove. Maybe we should have a `--merge` flag, to instead only **add** the new configurations?

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,166 @@
...
+		var kubeconfig *clientcmdapi.Config
+
+		if err := retry.OnError(availableBackoff, apierrors.IsNotFound, func() error {
+			kubeconfig, err = cfg.Generate(ctx, client, &virtualCluster, host[0], 0)
+			return err
+		}); err != nil {
+			return err
+		}
+
+		return writeKubeconfigFile(&virtualCluster, kubeconfig, "")
```
enrichman: As already discussed I don't think we should wait for the cluster to be ready. Some of the fields are not going to actually change something that needs a restart (i.e. labels/annotations), and I think it's better to have a fire&forget approach, and eventually add a `--w/wait` flag to wait for the update.

(We can think about an `--immediate` flag on the `cluster create` eventually).

tests/cli_test.go
```diff
@@ -201,6 +202,219 @@ var _ = When("using the k3kcli", Label("cli"), func() {
...
+
+			// Create the cluster with initial version
+			_, stderr, err = K3kcli("cluster", "create", "--version", "v1.31.13-k3s1", "--namespace", clusterNamespace, clusterName)
+			Expect(err).To(Not(HaveOccurred()), string(stderr))
+			Expect(stderr).To(ContainSubstring("You can start using the cluster"))
+
+			// Update the cluster version
+			_, stderr, err = K3kcli("cluster", "update", "--version", "v1.32.8-k3s1", "--namespace", clusterNamespace, clusterName)
+			Expect(err).To(Not(HaveOccurred()), string(stderr))
+			Expect(stderr).To(ContainSubstring("Updating cluster"))
```
enrichman, line 262: What about a downgrade? We should block this, not on the CLI, but on the server side. Ideally we should have a webhook, but we can probably have a Failed status with a proper description to warn the user.

We should add a test nevertheless.

go.mod
```diff
@@ -60,7 +60,7 @@ require (
 	github.com/antlr4-go/antlr/v4 v4.13.0 // indirect
 	github.com/asaskevich/govalidator v0.0.0-20230301143203-a9d515a09cc2 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
-	github.com/blang/semver/v4 v4.0.0 // indirect
+	github.com/blang/semver/v4 v4.0.0
```
enrichman: NIT: just for tidiness I usually move the direct deps in their block. :)

enrichman (changes requested):
Left some comments!

tests/cli_test.go
galal-hussein: I think it should be blocked on both, Its good practice to return an error for the user on the cli, instead of letting the user to figure out why the cluster isnt downgraded

galal-hussein: I will postpone adding this to the controller as well, because I think this PR should only include starter code for cli update

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+
+		namespace := appCtx.Namespace(name)
+
+		var virtualCluster v1beta1.Cluster
+
+		clusterKey := types.NamespacedName{Name: name, Namespace: appCtx.namespace}
+		if err := appCtx.Client.Get(ctx, clusterKey, &virtualCluster); err != nil {
+			return fmt.Errorf("failed to fetch existing cluster: %w", err)
+		}
```
enrichman, line 80: I would check if the error is a NotFound, to actually prompt the user to check for name/namespace (in case of a typo, maybe with a WARN) or error for any other problem.

cli/cmds/cluster_update_flags.go
```diff
@@ -0,0 +1,18 @@
...
+
+func updateFlags(cmd *cobra.Command, cfg *UpdateConfig) {
+	cmd.Flags().IntVar(&cfg.servers, "servers", 1, "number of servers")
+	cmd.Flags().IntVar(&cfg.agents, "agents", 0, "number of agents")
+	cmd.Flags().StringArrayVar(&cfg.labels, "labels", []string{}, "Labels to add to the cluster object (e.g. key=value)")
+	cmd.Flags().StringArrayVar(&cfg.annotations, "annotations", []string{}, "Annotations to add to the cluster object (e.g. key=value)")
+	cmd.Flags().StringVar(&cfg.version, "version", "", "k3s version")
+	cmd.Flags().DurationVar(&cfg.timeout, "timeout", 3*time.Minute, "The timeout for waiting for the cluster to become ready (e.g., 10s, 5m, 1h).")
+	cmd.Flags().StringVar(&cfg.kubeconfigServerHost, "kubeconfig-server", "", "Override the kubeconfig server host")
+	cmd.Flags().BoolVar(&cfg.skipInteractive, "y", false, "Skip interactive approval before applying update")
```
enrichman: I would add also the long flag `--yes` and change the help to something like "Don't ask for user interaction"

zypper install
```
-y, --no-confirm            Don't require user interaction. Alias for the --non-interactive global
                            option. Default: false
```

docs/cli/k3kcli_cluster_update.md
```diff
@@ -0,0 +1,40 @@
...
+      --agents int                 number of agents
+      --annotations stringArray    Annotations to add to the cluster object (e.g. key=value)
+  -h, --help                       help for update
+      --kubeconfig-server string   Override the kubeconfig server host
+      --labels stringArray         Labels to add to the cluster object (e.g. key=value)
+  -n, --namespace string           namespace of the k3k cluster
+      --servers int                number of servers (default 1)
+      --timeout duration           The timeout for waiting for the cluster to become ready (e.g., 10s, 5m, 1h). (default 3m0s)
+      --version string             k3s version
+      --y                          Skip interactive approval before applying update
```
enrichman: Hm, that is a bit weird, I think it's because we are missing the long flag.

enrichman (commented):
I am not sure we are not overriding existing fields. I.e. if we have first created a 2 agents virtual clusters, and then update the version.

Can we add a test for this, just creating one of the clusters with a different numbers of agents, for example?

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+	cmd := &cobra.Command{
+		Use:     "update",
+		Short:   "Update existing cluster",
+		Example: "k3kcli cluster update [command options] NAME",
+		RunE:    updateAction(appCtx, updateConfig),
+		Args:    cobra.ExactArgs(1),
+	}
+
+	CobraFlagNamespace(appCtx, cmd.Flags())
+	updateFlags(cmd, updateConfig)
```
enrichman, line 44: Can we keep the flags in the same file? They are just a few, and this file is not that long.

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+
+		namespace := appCtx.Namespace(name)
+
+		var virtualCluster v1beta1.Cluster
+
+		clusterKey := types.NamespacedName{Name: name, Namespace: appCtx.namespace}
+		if err := appCtx.Client.Get(ctx, clusterKey, &virtualCluster); err != nil {
+			if apierrors.IsNotFound(err) {
+				return fmt.Errorf("cluster %s not found in namespace %s. Please verify the cluster name and namespace are correct", name, appCtx.namespace)
```
enrichman: NIT

```suggestion
				return fmt.Errorf("cluster %s not found in namespace %s", name, appCtx.namespace)
```

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+
+			return fmt.Errorf("failed to fetch existing cluster: %w", err)
```
enrichman: NIT (the failure could happen even for a non-existing cluster, i.e. a networking issue)

```suggestion
			return fmt.Errorf("failed to fetch cluster: %w", err)
```

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+			virtualCluster.Spec.Version = config.version
+		}
+
+		if cmd.Flags().Changed("servers") {
+			virtualCluster.Spec.Servers = ptr.To(int32(config.servers))
+		}
+
+		if cmd.Flags().Changed("agents") {
+			virtualCluster.Spec.Agents = ptr.To(int32(config.agents))
+		}
```
enrichman, line 130: As discussed I would check if the value changed, and print the diff, to show the user what is going to change:


```go
			// if old.Version != config.version -> print "Updating Version: X.Y.Z -> X.Y.Q"
			virtualCluster.Spec.Version = config.version
		}

		if cmd.Flags().Changed("servers") {
			// if old.Servers != config.servers -> print "Updating Servers: N -> M"
			virtualCluster.Spec.Servers = ptr.To(int32(config.servers))
		}

		if cmd.Flags().Changed("agents") {
			virtualCluster.Spec.Agents = ptr.To(int32(config.agents))
		}
```

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+
+		clusterDetails, err := getClusterDetails(&virtualCluster)
+		if err != nil {
+			return fmt.Errorf("failed to get cluster details: %w", err)
+		}
+
+		if oldClusterDetails == clusterDetails {
+			logrus.Info("No changes detected, skipping update")
+			return nil
+		}
```
enrichman: I would rather check the semantic equal of the Specs and Metadata instead of doing an eq of the "print details". Or add a bool var to check if one of the field was updated.

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+
+		if !config.noConfirm {
```
enrichman, line 159: The negation of a `noConfirm` sounds a little bit confusing.

Maybe we can switch the flag to be a `confirm`, and have a `true` by default. This is just an internal detail, on the outside we will still have the `-y/--no-confirm` flags.

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil
+	}
+}
+
+func confirmClusterUpdate(oldClusterDetails, clusterDetails string) bool {
+	r := bufio.NewReader(os.Stdin)
+
+	logrus.Infof("Current %v\n\nNew %v\n\n", oldClusterDetails, clusterDetails)
+
+	fmt.Printf("Do you want to update the cluster? [y/N]: ")
```
enrichman: I would keep the print outside of this func, keeping it for just handling the confirmation. And maybe we can pass the config flag here:

i.e.


```suggestion
func confirmClusterUpdate(askConfirm bool) bool {
	if !askConfirm {
		return true
	}
	
	r := bufio.NewReader(os.Stdin)

	fmt.Printf("Do you want to update the cluster? [y/N]: ")
```

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+		if oldClusterDetails == clusterDetails {
+			logrus.Info("No changes detected, skipping update")
+			return nil
+		}
+
+		if !config.noConfirm {
+			if !confirmClusterUpdate(oldClusterDetails, clusterDetails) {
+				return nil
+			}
+		}
```
enrichman, line 163: Wrap up (even though I would NOT print the old cluster, as already discussed):


```suggestion
		logrus.Infof("Current %v\n\nNew %v\n\n", oldClusterDetails, clusterDetails)
		
		if !confirmClusterUpdate(config.confirm) {
			return nil
		}
```

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+		if cmd.Flags().Changed("servers") {
+			virtualCluster.Spec.Servers = ptr.To(int32(config.servers))
+		}
+
+		if cmd.Flags().Changed("agents") {
+			virtualCluster.Spec.Agents = ptr.To(int32(config.agents))
+		}
+
+		virtualCluster.Labels = mapCopyWithDefault(virtualCluster.Labels, parseKeyValuePairs(config.labels, "label"))
+		virtualCluster.Annotations = mapCopyWithDefault(virtualCluster.Annotations, parseKeyValuePairs(config.annotations, "annotation"))
```
enrichman: We can use the labels.Merge Kubernetes func for this (from the `"k8s.io/apimachinery/pkg/labels"` package):

```suggestion
		newLabels := parseKeyValuePairs(config.labels, "label")
		virtualCluster.Labels = labels.Merge(virtualCluster.Labels, parseKeyValuePairs(config.labels, "label"))
		
		newAnnotations := parseKeyValuePairs(config.annotations, "annotation")
		virtualCluster.Annotations =  labels.Merge(virtualCluster.Labels, newAnnotations)
```

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+		logrus.Fatalf("unable to read string: %v", err)
+	}
+
+	fmt.Printf("\n")
+
+	if len(res) < 2 {
+		return false
+	}
+
+	return strings.ToLower(strings.TrimSpace(res))[0] == 'y'
```
enrichman: You can use a Scanner to read from STDIO. It should be a little bit easier and shorter:

```go
scanner := bufio.NewScanner(os.Stdin)
	
if !scanner.Scan() {
	if err := scanner.Err(); err != nil {
		logrus.Errorf("Error reading input: %v", err)
	}
	return false
}

fmt.Printf("\n")

return strings.ToLower(strings.TrimSpace(scanner.Text())) == "y"
```

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	if dst == nil {
+		dst = make(map[string]string)
+	}
+
+	maps.Copy(dst, src)
+
+	return dst
+}
```
enrichman: This can be dropped, if using the labels package.

tests/cli_test.go
```diff
@@ -201,6 +202,173 @@ var _ = When("using the k3kcli", Label("cli"), func() {
...
+				stderr string
+				err    error
+			)
+
+			clusterName := "cluster-" + rand.String(5)
+			clusterNamespace := "k3k-" + clusterName
+
+			DeferCleanup(func() {
+				DeleteNamespaces(clusterNamespace)
+			})
```
enrichman: To avoid discrepancies from the other tests I have updated the CLI tests with the `NewNamespace()` func:

```suggestion
			namespace := NewNamespace()
			clusterNamespace := namespace.Name

			DeferCleanup(func() {
				DeleteNamespaces(clusterNamespace)
			})
```

This will create the namespace with the same (`ns-`) prefix.

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,176 @@
...
+
+	CobraFlagNamespace(appCtx, cmd.Flags())
+	updateFlags(cmd, updateConfig)
+
+	return cmd
+}
+
+func updateAction(appCtx *AppContext, config *UpdateConfig) func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		ctx := context.Background()
```
bigkevmcd: Any reason not to use `cmd.Context()`?

cli/cmds/cluster_update_flags.go
```diff
@@ -0,0 +1,18 @@
+package cmds
+
+import (
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+func updateFlags(cmd *cobra.Command, cfg *UpdateConfig) {
+	cmd.Flags().IntVar(&cfg.servers, "servers", 0, "number of servers")
```
bigkevmcd: Why not make the UpdateConfig `servers` and `agents` fields `int32` and you can then use `Int32Var` here and avoid the type conversion?

cli/cmds/cluster_update.go
galal-hussein: we recently moved to cobra and I think this part was just missed, I will change it in the update command but it should also be changed in other cli commands

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,203 @@
...
+			return fmt.Errorf("failed to fetch cluster: %w", err)
+		}
+
+		var changes []change
+
+		if cmd.Flags().Changed("version") && config.version != virtualCluster.Spec.Version {
+			currentVersion := virtualCluster.Spec.Version
+			if currentVersion == "" {
+				currentVersion = virtualCluster.Status.HostVersion
+			}
```
enrichman, line 88: This is ok, just a reminder that we should probably have a `virtualCluster.Status.Version`, without any further calculation.

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,203 @@
...
+func updateFlags(cmd *cobra.Command, cfg *UpdateConfig) {
+	cmd.Flags().Int32Var(&cfg.servers, "servers", 0, "number of servers")
+	cmd.Flags().Int32Var(&cfg.agents, "agents", 0, "number of agents")
+	cmd.Flags().StringArrayVar(&cfg.labels, "labels", []string{}, "Labels to add to the cluster object (e.g. key=value)")
+	cmd.Flags().StringArrayVar(&cfg.annotations, "annotations", []string{}, "Annotations to add to the cluster object (e.g. key=value)")
+	cmd.Flags().StringVar(&cfg.version, "version", "", "k3s version")
+	cmd.Flags().DurationVar(&cfg.timeout, "timeout", 3*time.Minute, "The timeout for waiting for the cluster to become ready (e.g., 10s, 5m, 1h).")
+	cmd.Flags().StringVar(&cfg.kubeconfigServerHost, "kubeconfig-server", "", "Override the kubeconfig server host")
+	cmd.Flags().BoolVarP(&cfg.noConfirm, "no-confirm", "y", false, "Skip interactive approval before applying update")
+}
```
enrichman, line 56: I would keep the same defaults we have in the cluster create, so `servers` should be 1.
Also the duration and kubeconfig-server are not needed, since we are not using it anymore.

```suggestion
func updateFlags(cmd *cobra.Command, cfg *UpdateConfig) {
	cmd.Flags().Int32Var(&cfg.servers, "servers", 1, "number of servers")
	cmd.Flags().Int32Var(&cfg.agents, "agents", 0, "number of agents")
	cmd.Flags().StringArrayVar(&cfg.labels, "labels", []string{}, "Labels to add to the cluster object (e.g. key=value)")
	cmd.Flags().StringArrayVar(&cfg.annotations, "annotations", []string{}, "Annotations to add to the cluster object (e.g. key=value)")
	cmd.Flags().StringVar(&cfg.version, "version", "", "k3s version")
	cmd.Flags().BoolVarP(&cfg.noConfirm, "no-confirm", "y", false, "Skip interactive approval before applying update")
}
```

cli/cmds/cluster_update.go
```diff
@@ -0,0 +1,203 @@
...
+)
+
+type UpdateConfig struct {
+	servers              int32
+	agents               int32
+	labels               []string
+	annotations          []string
+	version              string
+	kubeconfigServerHost string
+	timeout              time.Duration
```
enrichman:
```suggestion
```

enrichman (changes requested):
Last two small fixes, then LGTM!

PR #610, opened 2026-01-15
Fix allocatable resources in shared mode and Pod scheduling

### LINKED ISSUE #608: Incorrect nodes resources in shared mode
Nodes resources in shared mode are not correctly computed.

Creating a virtual cluster on a 3 nodes cluster will result in a 3x resources. Every virtual kubelet (one per node) will report the sum of all the nodes.

We need to reflect the correct resources.

enrichman:
The Virtual Kubelet component is deployed as a DaemonSet. This means there will be one pod per node.

Each Virtual Kubelet was reporting as allocatable resources the **sum** of all the nodes. This means that in a 3 nodes cluster the result will be multipled by 3. See:

Image

To fix this now we are getting the resources only from the node where the kubelet is running on, identified by the `AGENT_HOSTNAME` var.

**Important:** this PR also will schedule the pods on the same node where the virtual cluster scheduler opted for. This now works because of the 1-1 relationship with the host nodes.

```go
// Schedule the host pod in the same host node of the virtual kubelet
hostPod.Spec.NodeName = p.agentHostname
```

### DIFF
--- a/k3k-kubelet/kubelet.go
+++ b/k3k-kubelet/kubelet.go
@@ -265,7 +265,7 @@ func (k *kubelet) start(ctx context.Context) {
 
 func (k *kubelet) newProviderFunc(cfg config) nodeutil.NewProviderFunc {
 	return func(pc nodeutil.ProviderConfig) (nodeutil.Provider, node.NodeProvider, error) {
-		utilProvider, err := provider.New(*k.hostConfig, k.hostMgr, k.virtualMgr, k.logger, cfg.ClusterNamespace, cfg.ClusterName, cfg.ServerIP, k.dnsIP)
+		utilProvider, err := provider.New(*k.hostConfig, k.hostMgr, k.virtualMgr, k.logger, cfg.ClusterNamespace, cfg.ClusterName, cfg.ServerIP, k.dnsIP, cfg.AgentHostname)
 		if err != nil {
 			return nil, nil, errors.New("unable to make nodeutil provider: " + err.Error())
 		}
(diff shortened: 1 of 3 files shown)


k3k-kubelet/kubelet.go
```diff
@@ -265,7 +265,7 @@ func (k *kubelet) start(ctx context.Context) {
 
 func (k *kubelet) newProviderFunc(cfg config) nodeutil.NewProviderFunc {
 	return func(pc nodeutil.ProviderConfig) (nodeutil.Provider, node.NodeProvider, error) {
-		utilProvider, err := provider.New(*k.hostConfig, k.hostMgr, k.virtualMgr, k.logger, cfg.ClusterNamespace, cfg.ClusterName, cfg.ServerIP, k.dnsIP)
+		utilProvider, err := provider.New(*k.hostConfig, k.hostMgr, k.virtualMgr, k.logger, cfg.ClusterNamespace, cfg.ClusterName, cfg.ServerIP, k.dnsIP, cfg.AgentHostname)
```
galal-hussein, line 268: I think its maybe time to just send the kubelet instance and a config instance 
```suggestion
		utilProvider, err := provider.New(k, cfg)
```

galal-hussein (approved):

k3k-kubelet/kubelet.go
enrichman: As discussed this unfortunaly need a bigger refactor. It's better to do it in a different PR.

PR #623, opened 2026-01-21
Use resource.Quantity instead of a string for storageRequestSize

bigkevmcd:
Previously the resource.Quantity was stored as string which allowed invalid values to be created.

Tested this with an upgraded cluster and it looks good.

Trying to edit a cluster with `kubectl edit` yields appropriate errors.

```
# clusters.k3k.io "mycluster" was not valid:
# * spec.persistence.storageRequestSize: Invalid value: "2ng": spec.persistence.storageRequestSize in body should match '^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$'
```

golangci-lint also picked up some non-compliant files which I had to fix 😄

### DIFF
--- a/cli/cmds/cluster_create.go
+++ b/cli/cmds/cluster_create.go
@@ -190,14 +190,14 @@ func createAction(appCtx *AppContext, config *CreateConfig) func(cmd *cobra.Comm
 }
 
 func newCluster(name, namespace string, config *CreateConfig) (*v1beta1.Cluster, error) {
-	var storageRequestSize resource.Quantity
+	var storageRequestSize *resource.Quantity
 	if config.storageRequestSize != "" {
 		parsed, err := resource.ParseQuantity(config.storageRequestSize)
 		if err != nil {
 			return nil, err
 		}
 
-		storageRequestSize = parsed
+		storageRequestSize = ptr.To(parsed)
 	}
 
 	cluster := &v1beta1.Cluster{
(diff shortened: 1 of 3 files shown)


cli/cmds/cluster_create.go
```diff
@@ -185,7 +189,17 @@ func createAction(appCtx *AppContext, config *CreateConfig) func(cmd *cobra.Comm
...
+func newCluster(name, namespace string, config *CreateConfig) (*v1beta1.Cluster, error) {
+	var storageRequestSize resource.Quantity
+	if config.storageRequestSize != "" {
+		parsed, err := resource.ParseQuantity(config.storageRequestSize)
+		if err != nil {
+			return nil, err
+		}
+
+		storageRequestSize = parsed
+	}
```
enrichman, line 201: Should we add this check also in the `validateCreateConfig` we have in the PreRunE?

enrichman (approved):
LGTM! Thanks!

cli/cmds/cluster_create.go
bigkevmcd: It's already there albeit in a slightly different form? https://github.com/rancher/k3k/blob/main/cli/cmds/cluster_create_flags.go#L51-L53

enrichman: :man_facepalming: ups, I was looking to a different validation!

PR #638, opened 2026-01-26
Show correct allocatable resources when a Policy is applied

enrichman:
This PR fixes/adds the node capacity for the virtual cluster nodes in _shared_ mode.

By default the virtual nodes will show the node capacity the virtual kubelet are running on. If a quota is enforced (with a VirtualClusterPolicy, or with other ResourceQuotas enforced in the namespace), then it will be shared among the available virtual nodes.

For example if a quota of 10 GPUs is enabled in a 3 nodes cluster then the virtual nodes will show a capacity of 4-3-3 GPUs.

### DIFF
--- a/k3k-kubelet/provider/configure_capacity.go
+++ b/k3k-kubelet/provider/configure_capacity.go
@@ -128,26 +128,32 @@ func mergeResourceLists(resourceLists ...corev1.ResourceList) corev1.ResourceLis
 // distributeQuotas divides the total resource quotas evenly among all active virtual nodes.
 // This ensures that each virtual node reports a fair share of the available resources,
 // preventing the scheduler from overloading a single node.
+//
+// The algorithm iterates over each resource, divides it as evenly as possible among the
+// sorted virtual nodes, and distributes any remainder to the first few nodes to ensure
+// all resources are allocated. Sorting the nodes by name guarantees a deterministic
+// distribution.
 func distributeQuotas(ctx context.Context, logger logr.Logger, virtualClient client.Client, quotas corev1.ResourceList) (map[string]corev1.ResourceList, error) {
-	// List all virtual nodes to distribute the quota stably
+	// List all virtual nodes to distribute the quota stably.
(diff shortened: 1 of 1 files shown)


k3k-kubelet/provider/configure_capacity.go
```diff
@@ -0,0 +1,185 @@
...
+	return merged
+}
+
+// distributeQuotas divides the total resource quotas evenly among all active virtual nodes.
+// This ensures that each virtual node reports a fair share of the available resources,
+// preventing the scheduler from overloading a single node.
+func distributeQuotas(ctx context.Context, logger logr.Logger, virtualClient client.Client, quotas corev1.ResourceList) (map[string]corev1.ResourceList, error) {
+	// List all virtual nodes to distribute the quota stably
+	var virtualNodeList corev1.NodeList
+	if err := virtualClient.List(ctx, &virtualNodeList); err != nil {
```
galal-hussein, line 139: I think we should also check the state of the node, to make sure its in ready state, also for the deletion timestamp, or are we relying on subsequent calls to updateNodeCapacity

k3k-kubelet/provider/configure_capacity.go
```diff
@@ -0,0 +1,185 @@
...
+		} else {
+			totalValue = totalQuantity.Value()
+		}
+
+		quantityPerNode := totalValue / numNodes
+		remainder := totalValue % numNodes
+
+		for _, virtualNode := range virtualNodeList.Items {
+			nodeQuantity := quantityPerNode
+			if remainder > 0 {
```
galal-hussein, line 186: NIT: can we add a comment explaining the logic here

k3k-kubelet/provider/configure_capacity.go
enrichman: Yes, but these nodes are the one from the virtual cluster, so the virtual kubelet registered. So they have always a fake ready status (or Pending?).

PR #647, opened 2026-01-27
Bump kubernetes dependencies (v1.33)

enrichman:
This PR bumps the remaning dependencies to the Kubernetes v1.33.

Removing also the now unneeded replace, so we have a clean go.mod. :sparkles:


tests/common_test.go
```diff
@@ -475,3 +474,19 @@ func getServerIP(ctx context.Context, cfg *rest.Config) (string, error) {
 	// If Host includes a port, u.Hostname() extracts just the hostname part
 	return u.Hostname(), nil
 }
+
+// GetPodCondition extracts the provided condition from the given status and returns that.
```
pmatseykanets: Q: What is the reason for no longer using `pod.GetPodCondition()`?

enrichman: I had another look and it was because of an issue with the `k8s.io/controller-manager` package:

```
-> % go mod tidy
go: downloading k8s.io/controller-manager v0.0.0
go: github.com/rancher/k3k/tests tested by
        github.com/rancher/k3k/tests.test imports
        k8s.io/kubernetes/pkg/api/v1/pod imports
        k8s.io/kubernetes/pkg/features imports
        k8s.io/controller-manager/pkg/features: reading k8s.io/controller-manager/go.mod at revision v0.0.0: unknown revision v0.0.0
```

But I have now tried to pin the version of it and it builds fine!

```
# go.mod
k8s.io/controller-manager v0.33.7 // indirect
```

What do you think, should I revert this part? :eyes:

pmatseykanets: > should I revert this part?

I think so, there is no point in maintaining our own copy of `GetPodCondition`.

enrichman: Done with 90b5aa1!

pmatseykanets (approved):

Issue #348, opened 2025-05-07
Projected serviceaccounttokens volumes are not being handled in shared mode


galal-hussein:
**Environmental Info:**
Host Cluster Version: 

v1.32.3+k3s1
Node(s) CPU architecture, OS, and Version: 

1 node ubuntu 24.04

Host Cluster Configuration:

1 server

K3K Cluster Configuration:

1 server shared mode

**Describe the bug:**

In shared mode we transform volumes from virtual cluster to host cluster that includes:
- secrets
- configmaps
- projected configmaps
- projected secrets

However projected serviceaccounttokens are not being handled correctly, this causing the failure of the following conformance test as well

```
 [FAIL] [sig-auth] ServiceAccounts [It] ServiceAccountIssuerDiscovery should support OIDC discovery of service account issuer [Conformance] [sig-auth, Conformance]
  k8s.io/kubernetes/test/e2e/auth/service_accounts.go:666
```

galal-hussein:
## Validation Template

Context: https://github.com/rancher/k3k/pull/660

## Root Cause

Projected service account tokens were not translated in shared mode, this issue caused a known failure in conformance tests

## What was fixed, or what change have occurred

k3k-kubelet now handles translation of projected service account tokens, when creating a pod with projected service account volume, then it should be picked up and translated successfully:

## Areas or cases that should be tested

### Case 1

1. Create a shared k3k cluster
2. Create a pod that uses projected service account token:

```
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
spec:
  containers:
  - image: nginx
    name: test-pod
    volumeMounts:
    - mountPath: /var/run/secrets/tokens
      name: sa-token
      readOnly: true
  volumes:
  - name: sa-token
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          audience: test-audiance
          expirationSeconds: 600
          path: sa-token
```

3. pod shoud be created correctly and the projected token has succefully translated on the host cluster

```
kubectl get secrets -n shared-x k3k-default-test-audiance-600-sa-token-default-shar-e9925 -o json 
{
    "apiVersion": "v1",
    "data": {
        "sa-token": "ZXlKa...."
    },
    "kind": "Secret",
    "metadata": {
        "annotations": {
            "k3k.io/name": "k3k-default-oidc-discovery-test-600-sa-token",
            "k3k.io/namespace": "default"
        },
        "creationTimestamp": "2026-04-16T19:26:51Z",
[...]

PR #660, opened 2026-02-04
Add support for projected service account tokens in shared mode

galal-hussein:
- #348

### DIFF
--- a/k3k-kubelet/provider/token_test.go
+++ b/k3k-kubelet/provider/token_test.go
@@ -12,7 +12,7 @@ import (
 	k3kcontroller "github.com/rancher/k3k/pkg/controller"
 )
 
-func Test_hasKubeAccessVolumeFound(t *testing.T) {
+func Test_hasKubeAccessVolume(t *testing.T) {
 	tests := []struct {
 		name string
 		pod  *corev1.Pod
@@ -73,7 +73,7 @@ func Test_hasKubeAccessVolumeFound(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			assert.Equal(t, tt.want, hasKubeAccessVolumeFound(tt.pod))
+			assert.Equal(t, tt.want, hasKubeAccessVolume(tt.pod))
 		})
 	}
 }


k3k-kubelet/provider/token.go
```diff
@@ -55,23 +71,95 @@ func (p *Provider) transformTokens(ctx context.Context, pod, tPod *corev1.Pod) e
...
+						return err
+					}
 
+					hostSecret, err := p.translateAndCreateHostTokenSecret(ctx, projectedSecret)
+					if err != nil {
+						return err
+					}
+
+					// replace the projected token volume with a projected secret
+					// todo check the volume name
```
pmatseykanets: Should this TODO be addressed?

k3k-kubelet/provider/token.go
```diff
@@ -24,8 +26,22 @@ const (
 // to look like the serviceaccount token
 func (p *Provider) transformTokens(ctx context.Context, pod, tPod *corev1.Pod) error {
 	logger := p.logger.WithValues("namespace", pod.Namespace, "name", pod.Name, "serviceAccountNameod", pod.Spec.ServiceAccountName)
```
pmatseykanets:
```suggestion
	logger := p.logger.WithValues("namespace", pod.Namespace, "name", pod.Name, "serviceAccountName", pod.Spec.ServiceAccountName)
```

k3k-kubelet/provider/token.go
```diff
@@ -89,14 +177,20 @@ func virtualSecret(name, namespace, serviceAccountName string) *corev1.Secret {
...
-	addKubeAccessVolume(pod, hostSecretName)
+func (p *Provider) translateAndCreateHostTokenSecret(ctx context.Context, projectedToken *corev1.Secret) (*corev1.Secret, error) {
+	hostSecret := projectedToken.DeepCopy()
+	hostSecret.Type = ""
+	hostSecret.Annotations = make(map[string]string)
+
+	p.Translator.TranslateTo(hostSecret)
+
+	if err := p.HostClient.Create(ctx, hostSecret); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
```
pmatseykanets: When the secret already exists, the function returns the locally constructed `hostSecret` not the one actually that is in the cluster. If the existing secret has different data, the pod gets mounted with the locally-built secret metadata while the actual cluster secret may have different content. We should either `Get` the existing secret or do a `Update`/`Patch`.

k3k-kubelet/provider/token.go
```diff
@@ -55,23 +71,95 @@ func (p *Provider) transformTokens(ctx context.Context, pod, tPod *corev1.Pod) e
...
+
+	tokenRequest := &authv1.TokenRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      serviceAccountName,
+			Namespace: namespace,
+		},
+		Spec: authv1.TokenRequestSpec{
+			Audiences:         []string{token.Audience},
+			ExpirationSeconds: token.ExpirationSeconds,
+			BoundObjectRef: &authv1.BoundObjectReference{
```
pmatseykanets, line 147: This most likely requires also setting the UID `UID: pod.UID` as the binding is purely name-based and won't detect pod replacement.

k3k-kubelet/provider/token.go
```diff
@@ -55,23 +71,95 @@ func (p *Provider) transformTokens(ctx context.Context, pod, tPod *corev1.Pod) e
...
+	namespace := pod.Namespace
+	serviceAccountName := pod.Spec.ServiceAccountName
+
+	tokenRequest := &authv1.TokenRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      serviceAccountName,
+			Namespace: namespace,
+		},
+		Spec: authv1.TokenRequestSpec{
+			Audiences:         []string{token.Audience},
```
pmatseykanets: If `token.Audience` is empty (optional per the K8s API), this creates `[]string{""}` rather than `nil`. The TokenRequest API treats an explicit empty audience differently from no audience; sending [""] may produce a token with unintended audience claims.

k3k-kubelet/provider/token.go
```diff
@@ -55,23 +71,95 @@ func (p *Provider) transformTokens(ctx context.Context, pod, tPod *corev1.Pod) e
...
-	if err := p.HostClient.Create(ctx, hostSecret); err != nil {
-		if !apierrors.IsAlreadyExists(err) {
-			return err
+	removeKubeAccessVolume(tPod)
+	addKubeAccessVolume(tPod, hostSecret.Name)
+
+	return nil
+}
+
+func (p *Provider) transformProjectedTokens(ctx context.Context, pod, tPod *corev1.Pod) error {
```
pmatseykanets: This is fragile. It iterates over indexes of pod.Spec.Volumes but applies changes to tPod.Spec.Volumes[i]. This work because tPod was DeepCopy'ed from pod so they have the same volume order/count. Might be worth adding a comment or do a check.

galal-hussein: I changed this part to iterate over the target pod volumes instead, so that we are working on a single slice, thanks!

k3k-kubelet/provider/token.go
galal-hussein: Added Pod UID as well

k3k-kubelet/provider/token.go
galal-hussein: okay I switched to a get-or-create pattern, but I think we are safe either way since this happens only on pod creation from the provider.

k3k-kubelet/provider/provider.go
```diff
@@ -72,24 +73,30 @@ func New(hostConfig rest.Config, hostMgr, virtualMgr manager.Manager, logger log
...
-		serverIP:         serverIP,
-		dnsIP:            dnsIP,
-		agentHostname:    agentHostname,
+		HostClient:        hostMgr.GetClient(),
+		VirtualClient:     virtualMgr.GetClient(),
+		VirtualManager:    virtualMgr,
+		Translator:        translator,
+		ClientConfig:      hostConfig,
+		CoreClient:        coreClient,
+		VirtualCoreClient: virtualCoreClient,
```
enrichman: It looks like we are having some common objects in the provider. We should probably wrap them, like we did in the Syncer, creating a dedicated struct:

```go
type ClusterContext struct {
	Client     client.Client
	CoreClient cv1.CoreV1Interface
	Manager    manager.Manager
}
```

```go
p := Provider{
	Host: &ClusterContext{
		Manager:	hostMgr,
		Client:		hostMgr.GetClient(),
		CoreClient:	hostCoreClient,
	}
	...
}
```

wdyt?

k3k-kubelet/provider/token.go
```diff
@@ -89,14 +187,24 @@ func virtualSecret(name, namespace, serviceAccountName string) *corev1.Secret {
...
+
+		if err := p.HostClient.Create(ctx, hostSecret); err != nil {
+			return nil, err
+		}
+	}
+
+	return hostSecret, nil
 }
 
 func isKubeAccessVolumeFound(pod *corev1.Pod) bool {
```
enrichman: NIT: can we rename this to `hasKubeAPIAccessVolume`?

```suggestion
func hasKubeAPIAccessVolume(pod *corev1.Pod) bool {
```

k3k-kubelet/provider/token.go
```diff
@@ -55,23 +72,104 @@ func (p *Provider) transformTokens(ctx context.Context, pod, tPod *corev1.Pod) e
...
+	removeKubeAccessVolume(tPod)
+	addKubeAccessVolume(tPod, hostSecret.Name)
+
+	return nil
+}
+
+// transformProjectedTokens will iterate over the target pod projected volume sources
+// and transform projected tokens to use a requested token secret from the virtual cluster
+// instead the automatically generated secret on the host cluster.
+func (p *Provider) transformProjectedTokens(ctx context.Context, pod, tPod *corev1.Pod) error {
```
enrichman: Same as before: can we use the `virtualPod` and `hostPod` naming convention?

```suggestion
func (p *Provider) transformProjectedTokens(ctx context.Context, virtualPod, hostPod *corev1.Pod) error {
```

k3k-kubelet/provider/token.go
```diff
@@ -55,23 +72,104 @@ func (p *Provider) transformTokens(ctx context.Context, pod, tPod *corev1.Pod) e
...
+					// replace the projected token volume with a projected secret
+					tPod.Spec.Volumes[i].Projected.Sources[j].ServiceAccountToken = nil
+					tPod.Spec.Volumes[i].Projected.Sources[j].Secret = &corev1.SecretProjection{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: hostSecret.Name,
+						},
+					}
+				}
+			}
+		}
```
enrichman: To keep a nicer linear indentation I usually prefer to use guards and early returns/continue:

```suggestion
		if volume.Projected == nil {
			continue
		}
		
		for j, source := range volume.Projected.Sources {
			if source.ServiceAccountToken == nil {
				continue
			}
			
			projectedSecret, err := p.requestTokenSecret(ctx, source.ServiceAccountToken, pod)
			if err != nil {
				return err
			}

			hostSecret, err := p.translateAndCreateHostTokenSecret(ctx, projectedSecret)
			if err != nil {
				return err
			}

			// replace the projected token volume with a projected secret
			tPod.Spec.Volumes[i].Projected.Sources[j].ServiceAccountToken = nil
			tPod.Spec.Volumes[i].Projected.Sources[j].Secret = &corev1.SecretProjection{
				LocalObjectReference: corev1.LocalObjectReference{
					Name: hostSecret.Name,
				},
			}
		}
```

k3k-kubelet/provider/token.go
```diff
@@ -55,23 +72,104 @@ func (p *Provider) transformTokens(ctx context.Context, pod, tPod *corev1.Pod) e
...
+				Kind:       "Pod",
+				APIVersion: "v1",
+			},
+		},
+	}
+
+	tokenResp, err := p.VirtualCoreClient.ServiceAccounts(namespace).CreateToken(ctx, serviceAccountName, tokenRequest, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
```
enrichman: I am not sure about the workflow, but could this be called twice and fail?

k3k-kubelet/provider/token.go
```diff
@@ -89,14 +187,24 @@ func virtualSecret(name, namespace, serviceAccountName string) *corev1.Secret {
...
 
-// translateToken will remove the serviceaccount from the pod and replace the kube-api-access volume
-// with a custom token volume and mount it to all containers within the pod
-func (p *Provider) translateToken(pod *corev1.Pod, hostSecretName string) {
-	pod.Spec.ServiceAccountName = ""
-	pod.Spec.DeprecatedServiceAccount = ""
-	pod.Spec.AutomountServiceAccountToken = ptr.To(false)
-	removeKubeAccessVolume(pod)
-	addKubeAccessVolume(pod, hostSecretName)
+func (p *Provider) translateAndCreateHostTokenSecret(ctx context.Context, projectedToken *corev1.Secret) (*corev1.Secret, error) {
```
pmatseykanets, line 193: It seems like this never updates stale secrets. We should probably go for create-or-update pattern.

When the secret already exists on the host (e.g., from a previous pod with the same SA/audience/expiration), Get populates hostSecret with the existing data, and the freshly requested token from requestTokenSecret is silently discarded.

k3k-kubelet/provider/token.go
```diff
@@ -171,4 +279,24 @@ func addKubeAccessVolume(pod *corev1.Pod, hostSecretName string) {
...
+		pod.Spec.EphemeralContainers[i].VolumeMounts = append(pod.Spec.EphemeralContainers[i].VolumeMounts, corev1.VolumeMount{
+			Name:      tokenVolumeName,
+			MountPath: serviceAccountTokenMountPath,
+		})
+	}
+}
+
+func generateTokenSecretName(serviceAccountName string, tokenReq *authv1.TokenRequest) string {
+	nameStr := "k3k-" + serviceAccountName
+	if tokenReq.Spec.Audiences != nil {
```
pmatseykanets, line 295: Audiences are user-controlled strings from the virtual cluster pod spec. They're concatenated directly into a Kubernetes resource name.
Should we:
- make sure we don't max out the identifier length (k8s 253 characters)
- sanitize (audience may contain dots, slashes, uppercase)

Also two projected volumes with the same (SA, audience, expiration) but different token.Path currently produce the same secret name — the second volume silently gets the first volume's data.

pmatseykanets: Maybe include token.Path in the name generation and use SafeConcatName or a hash to ensure uniqueness?

k3k-kubelet/provider/token.go
galal-hussein: no not really the create token will just create a new token for if called again, I handle idempotency later the actual creation of the secret already makes sure that we dont create a host secret unless the secret does not exist first

k3k-kubelet/provider/token.go
galal-hussein: I added the token path to the name generation plus using SafeConcatName to ensure safe naming

k3k-kubelet/provider/token.go
galal-hussein: I am using now `CreateOrUpdate` method, instead of ignoring already exist hostSecret

k3k-kubelet/provider/token.go
```diff
@@ -171,4 +280,29 @@ func addKubeAccessVolume(pod *corev1.Pod, hostSecretName string) {
...
+func generateTokenSecretName(serviceAccountName, tokenPath string, tokenReq *authv1.TokenRequest) string {
+	nameComponents := []string{serviceAccountName}
+
+	if tokenReq.Spec.Audiences != nil {
+		nameComponents = append(nameComponents, tokenReq.Spec.Audiences...)
+	}
+
+	if tokenReq.Spec.ExpirationSeconds != nil {
+		nameComponents = append(nameComponents, strconv.Itoa(int(*tokenReq.Spec.ExpirationSeconds)))
+	}
```
enrichman: NIT: this should work

```suggestion
	if exp := tokenReq.Spec.ExpirationSeconds; exp != nil {
		nameComponents = append(nameComponents, strconv.FormatInt(*exp, 10))
	}
```

or also with `fmt` should be fine (performance wise):

```suggestion
	if exp := tokenReq.Spec.ExpirationSeconds; exp != nil {
		nameComponents = append(nameComponents, fmt.Sprint(*exp))
	}
```

k3k-kubelet/provider/token.go
```diff
@@ -89,17 +190,25 @@ func virtualSecret(name, namespace, serviceAccountName string) *corev1.Secret {
...
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	return hostSecret, nil
 }
 
-func isKubeAccessVolumeFound(pod *corev1.Pod) bool {
+func hasKubeAccessVolumeFound(pod *corev1.Pod) bool {
```
enrichman:
```suggestion
func hasKubeAccessVolume(pod *corev1.Pod) bool {
```

PR #662, opened 2026-02-06
Change the nodes type to worker.

bigkevmcd:
This modifies the configuration of the created nodes via virtual-kubelet to set the node type to be `worker` instead of `agent`.


tests/cluster_create_test.go
```diff
@@ -0,0 +1,46 @@
...
+			g.Expect(err).To(Not(HaveOccurred()))
+			if len(nodes.Items) == 0 {
+				return map[string]string{}
+			}
+
+			g.Expect(nodes.Items).To(HaveLen(1))
+
+			return nodes.Items[0].Labels
+		}).WithTimeout(time.Minute).WithPolling(time.Second).Should(
+			HaveKeyWithValue("node-role.kubernetes.io/worker", "true"))
```
enrichman: As the Copilot suggestion is probably better to assert inside the Eventually:

```suggestion
		Eventually(func(g Gomega) map[string]string {
			nodes, err := virtualCluster.Client.CoreV1().Nodes().List(GinkgoT().Context(), metav1.ListOptions{})
			g.Expect(err).To(Not(HaveOccurred()))
			
			g.Expect(nodes.Items).To(HaveLen(1))
			g.Expect(nodes.Items).To(HaveKeyWithValue("node-role.kubernetes.io/worker", "true"))
		}).
			WithTimeout(time.Minute).
			WithPolling(time.Second).
			Should(Succeed())
```

Makefile
```diff
@@ -6,7 +6,7 @@ VERSION ?= $(shell git describe --tags --always --dirty --match="v[0-9]*")
 ## Dependencies
 
 GOLANGCI_LINT_VERSION := v2.8.0
-GINKGO_VERSION ?= v2.21.0
+GINKGO_VERSION ?= v2.28.1
```
enrichman, line 9: Oh, pretty old!

enrichman (commented):
LGTM!

Let's wait for the new release before merging, and backporting.

galal-hussein (approved):

PR #663, opened 2026-02-06
Added `policy` in Cluster Status

### LINKED ISSUE #656: When a Policy is applied some Cluster fields are overwritten
A user creates a Cluster with a nodeSelector or a priorityClass. If a policy is then applied to the cluster namespace it will override the user defined behavior, and when removed the original intention will be lost.

enrichman:
This PR fix #656 adding a new `policy` field in the Cluster.Status.

This field is used to track the current enforced policy, and some of its values. The PriorityClass and NodeSelector are now checked from this new field, without changing the original Cluster Spec.

Note: there was already a `policyName` field in the ClusterStatus, used to retrieve the current policy.


pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -561,6 +567,22 @@ type ClusterStatus struct {
 	Phase ClusterPhase `json:"phase,omitempty"`
 }
 
+// AppliedPolicy defines the observed state of an applied policy.
+type AppliedPolicy struct {
+	// Name is the name of the VirtualClusterPolicy currently applied to this cluster.
+	Name string `json:"name,omitempty"`
```
bigkevmcd, line 576: This should be marked as `optional` or `required`?

pkg/controller/policy/namespace.go
```diff
@@ -90,3 +114,29 @@ func (c *VirtualClusterPolicyReconciler) cleanupNamespaces(ctx context.Context)
 
 	return nil
 }
+
+// clearPolicyFieldsForClustersInNamespace sets the policy status on Cluster objects in the given namespace to nil.
+func (c *VirtualClusterPolicyReconciler) clearPolicyFieldsForClustersInNamespace(ctx context.Context, namespace string) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	var clusters v1beta1.ClusterList
+	if err := c.Client.List(ctx, &clusters, client.InNamespace(namespace)); err != nil {
```
bigkevmcd, line 124: Can we wrap some of these errors?

bigkevmcd (changes requested):

k3k-kubelet/provider/provider.go
```diff
@@ -401,28 +401,45 @@ func (p *Provider) createPod(ctx context.Context, pod *corev1.Pod) error {
...
+	}
 
 	// setting the hostname for the pod if its not set
 	if virtualPod.Spec.Hostname == "" {
 		hostPod.Spec.Hostname = k3kcontroller.SafeConcatName(virtualPod.Name)
 	}
 
-	// if the priorityClass for the virtual cluster is set then override the provided value
+	// When a PriorityClass is set we will use the translated one in the HostCluster.
+	// If the Cluster or a Policy define a PriorityClass of the host we are going to use that one.
```
galal-hussein: nit
```suggestion
	// If the Cluster or a Policy defines a PriorityClass of the host we are going to use that one.
```

pkg/controller/policy/namespace.go
```diff
@@ -52,6 +55,27 @@ func (c *VirtualClusterPolicyReconciler) cleanupNamespaces(ctx context.Context)
 	}
 
 	for _, ns := range namespaces.Items {
+		currentPolicyName := ns.Labels[PolicyNameLabelKey]
+
+		// If the namespace is not bound to any policy, or if the policy it was bound to no longer exists,
+		// we need to clear policy-related fields on its Cluster objects.
+		if currentPolicyName == "" {
```
galal-hussein, line 69: I think we recheck for the same condition in line 86 
https://github.com/rancher/k3k/pull/663/changes#diff-563d9396f6490a03561faf0a2396786d5696e95d6a938b71e7de1fb2c200f64bR86

galal-hussein (commented):
looks very good overall, I have some nit comments, but LGTM for me

pkg/controller/policy/namespace.go
enrichman: You're right, I merged the two branches, thanks!

pkg/controller/policy/namespace.go
enrichman: Done!

pkg/apis/k3k.io/v1beta1/types.go
enrichman: It should be required. It doesn't makes sense to have a policy without a name.

I've added the needed kubebuilder annotations for the validation. Thanks!

```go
// name is the name of the VirtualClusterPolicy currently applied to this cluster.
//
// +kubebuilder:validation:MinLength:=1
// +required
Name string `json:"name,omitempty"`
```

galal-hussein (approved):

bigkevmcd (approved):

PR #672, opened 2026-02-17
minor fix to translated ingress tls secrets in ingress object

### LINKED ISSUE #671: shared cluster ingress tls secrets are not mapped correctly
**Environmental Info:**
Host Cluster Version: 

k3s v1.34.4+k3s1

Linux k3k 5.14.21-150500.55.83-default #1 SMP PREEMPT_DYNAMIC Wed Oct 2 08:09:07 UTC 2024 (0d53847) x86_64 x86_64 x86_64 GNU/Linux

Host Cluster Configuration:

Single node vanilla k3s install

K3K Cluster Configuration:

```
apiVersion: k3k.io/v1beta1
kind: Cluster
metadata:
  name: rancher-k3k-cluster
  namespace: rancher-k3k
spec:
  mode: shared
  servers: 1
  version: v1.33.1-k3s1
  sync:
    ingresses:
      enabled: true
  addons:
    - secretRef: rancher-init
      secretNamespace: rancher-k3k-init
  persistence:
    type: ephemeral
```
**Describe the bug:**

**Steps To Reproduce:**

* Setup k3k cluster to deploy rancher via addons using the following yaml manifest
```
---
apiVersion: v1
kind: Namespace
metadata:
  name: rancher-k3k
---
apiVersion: v1
kind: Namespace
metadata:
  name: rancher-k3k-init
---
apiVersion: v1
kind: Secret
metadata:
  name: rancher-init
  namespace: rancher-k3k-init
stringData:
  manifests.yaml: |-
    apiVersion: v1
    kind: Namespace
    metadata:
      name: cattle-system
    ---
    apiVersion: v1
    kind: Namespace
    metadata:
      name: cert-manager
      labels:
        certmanager.k8s.io/disable-validation: "true"
    ---
    apiVersion: helm.cattle.io/v1
    kind: HelmChart
    metadata:
      name: cert-manager
      namespace: kube-system
    spec:
      targetNamespace: cert-manager
      repo: https://charts.jetstack.io
      chart: cert-manager
      version: v1.5.1
      helmVersion: v3
      set:
        installCRDs: "true"
    ---
    apiVersion: helm.cattle.io/v1
    kind: HelmChart
    metadata:
      name: rancher
      namespace: kube-system
    spec:
      targetNamespace: cattle-system
      repo: https://releases.rancher.com/server-charts/latest
      chart: rancher
      version: v2.13.0
      set:
        priorityClassName: system-cluster-critical
        ingress.tls.source: rancher
        hostname: rancher.10.115.53.168.sslip.io
        replicas: 1
        global.cattle.psp.enabled: "false"
        bootstrapPassword: password@1234
      helmVersion: v3
---
apiVersion: k3k.io/v1beta1
kind: Cluster
metadata:
  name: rancher-k3k-cluster
  namespace: rancher-k3k
spec:
  mode: shared
  servers: 1
  version: v1.33.1-k3s1
  sync:
    ingresses:
      enabled: true
  addons:
    - secretRef: rancher-init
      secretNamespace: rancher-k3k-init
  persistence:
    type: ephemeral
```
* After rancher is installed and ingress is created, the ingress on the host cluster references an invalid tls secret
```
apiVersion: v1
items:
- apiVersion: networking.k8s.io/v1
  kind: Ingress
  metadata:
    annotations:
      cert-manager.io/issuer: rancher
      cert-manager.io/issuer-kind: Issuer
      k3k.io/name: rancher
      k3k.io/namespace: cattle-system
      meta.helm.sh/release-name: rancher
      meta.helm.sh/release-namespace: cattle-system
      nginx.ingress.kubernetes.io/proxy-connect-timeout: "30"
      nginx.ingress.kubernetes.io/proxy-read-timeout: "1800"
      nginx.ingress.kubernetes.io/proxy-send-timeout: "1800"
    creationTimestamp: "2026-02-17T00:59:25Z"
    generation: 2
    labels:
      app: rancher
      app.kubernetes.io/managed-by: Helm
      chart: rancher-2.13.0
      heritage: Helm
      k3k.io/clusterName: rancher-k3k-cluster
      release: rancher
    name: rancher-cattle-system-rancher-k3k-cluster-72616e636865722-7127c
    namespace: rancher-k3k
    ownerReferences:
    - apiVersion: k3k.io/v1beta1
      blockOwnerDeletion: true
      controller: true
      kind: Cluster
      name: rancher-k3k-cluster
      uid: 8436709e-5cb1-4fc6-a5eb-d186c3bdfc5e
    resourceVersion: "5858"
    uid: e92efc3b-055a-4dc5-ad0b-8607ef22f4d0
  spec:
    rules:
    - host: rancher.10.115.53.168.sslip.io
      http:
        paths:
        - backend:
            service:
              name: rancher-cattle-system-rancher-k3k-cluster-72616e636865722-7127c
              port:
                number: 80
          path: /
          pathType: ImplementationSpecific
    tls:
    - hosts:
      - rancher.10.115.53.168.sslip.io
      secretName: tls-rancher-ingress
```

This is because the `tls` section is not translated by the k3k-kubelet to the reference the translated secret on the host cluster.

**Expected behavior:**

Correct translated secret from host cluster should be referenced in the host ingress

**Actual behavior:**

Secret is not translated as a result incorrect secret is used to expose rancher, which breaks subsequent functionality during rancher operation.

ibrokethecloud:
PR ensures TLS secret's are translated to host ingress object.

Addresses: https://github.com/rancher/k3k/issues/671

To validate install rancher on a shared k3k cluster with ingress sync enabled.

Please refer to issue to find manifest used for deploying rancher on k3k.

Post this change the synced ingress on host cluster references the correct tls secret, and rancher ingress leverages the correct secret for.

### DIFF
--- a/k3k-kubelet/controller/syncer/ingress.go
+++ b/k3k-kubelet/controller/syncer/ingress.go
@@ -92,12 +92,20 @@ func (r *IngressReconciler) Reconcile(ctx context.Context, req reconcile.Request
 		return reconcile.Result{}, err
 	}
 
-	syncConfig := cluster.Spec.Sync.Ingresses
+	appliedSync := cluster.Spec.Sync.DeepCopy()
+
+	// If a policy is applied to the virtual cluster we need to use its SyncConfig, if available
+	if cluster.Status.Policy != nil && cluster.Status.Policy.Sync != nil {
+		appliedSync = cluster.Status.Policy.Sync
+	}
+
+	syncConfig := appliedSync.Ingresses
+
 	if err := r.VirtualClient.Get(ctx, req.NamespacedName, &virtIngress); err != nil {
 		return reconcile.Result{}, ctrlruntimeclient.IgnoreNotFound(err)
 	}
 
-	syncedIngress := r.ingress(&virtIngress, syncConfig.SyncTLSSecrets)
+	syncedIngress := r.ingress(&virtIngress, syncConfig.DisableTLSSecretTranslation)
 
 	if err := controllerutil.SetOwnerReference(&cluster, syncedIngress, r.HostClient.Scheme()); err != nil {
 		return reconcile.Result{}, err
@@ -144,7 +152,7 @@ func (r *IngressReconciler) Reconcile(ctx context.Context, req reconcile.Request
 	return reconcile.Result{}, r.HostClient.Update(ctx, syncedIngress)
 }
 
-func (s *IngressReconciler) ingress(obj *networkingv1.Ingress, syncTLSSecrets bool) *networkingv1.Ingress {
+func (s *IngressReconciler) ingress(obj *networkingv1.Ingress, disableTLSSecretTranslation bool) *networkingv1.Ingress {
 	hostIngress := obj.DeepCopy()
 	s.Translator.TranslateTo(hostIngress)
 
@@ -159,15 +167,16 @@ func (s *IngressReconciler) ingress(obj *networkingv1.Ingress, syncTLSSecrets bo
 		}
 	}
 
+	// TLS Secret translation disable, return early without translating TLS secrets in the ingress spec
+	if disableTLSSecretTranslation {
+		return hostIngress
+	}
 	// ensure tls secrets are also translated
-	if syncTLSSecrets {
-		for i := range hostIngress.Spec.TLS {
-			if hostIngress.Spec.TLS[i].SecretName != "" {
-				hostIngress.Spec.TLS[i].SecretName = s.Translator.TranslateName(obj.GetNamespace(), hostIngress.Spec.TLS[i].SecretName)
-			}
+	for i := range hostIngress.Spec.TLS {
+		if hostIngress.Spec.TLS[i].SecretName != "" {
+			hostIngress.Spec.TLS[i].SecretName = s.Translator.TranslateName(obj.GetNamespace(), hostIngress.Spec.TLS[i].SecretName)
 		}
 	}
 
-	// don't sync finalizers to the host
 	return hostIngress
 }
--- a/k3k-kubelet/controller/syncer/ingress.go
+++ b/k3k-kubelet/controller/syncer/ingress.go
@@ -92,11 +92,12 @@ func (r *IngressReconciler) Reconcile(ctx context.Context, req reconcile.Request
 		return reconcile.Result{}, err
 	}
 
+	syncConfig := cluster.Spec.Sync.Ingresses
 	if err := r.VirtualClient.Get(ctx, req.NamespacedName, &virtIngress); err != nil {
 		return reconcile.Result{}, ctrlruntimeclient.IgnoreNotFound(err)
 	}
 
-	syncedIngress := r.ingress(&virtIngress)
+	syncedIngress := r.ingress(&virtIngress, syncConfig.SyncTLSSecrets)
 
 	if err := controllerutil.SetOwnerReference(&cluster, syncedIngress, r.HostClient.Scheme()); err != nil {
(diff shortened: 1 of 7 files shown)


jp-gouin:
Thanks  @ibrokethecloud for opening the PR.
Below the comment I added on the issue, we can think about adding a new sync option within `ingresses` to also specify to translate secret name (we might want to make it explicit to cert-manager like `host-cert-manager: true`) 

@enrichman what do you think? Maybe there is no harm to always translate the secret name ? 

> The `tls` secret not being translated is actually an expected behavior. 
> The cert-manager on the host cluster would create the certificate for the virtual cluster user.
> 
> Now I can see some scenarios when this would failed. Like if cert-manager does not exist on the host or if the virtual cluster user wants to sync the ingress and handling the certificate.

ibrokethecloud:
a better option could be to extend `IngressSyncConfig` to have an extra field say `SyncTLSSecrets` which controls if tls secrets should be sync from virtual cluster to host cluster

```
// IngressSyncConfig specifies the sync options for services.
type IngressSyncConfig struct {
	// Enabled is an on/off switch for syncing resources.
	//
	// +kubebuilder:default=false
	// +required
	Enabled bool `json:"enabled"`

	// Selector specifies set of labels of the resources that will be synced, if empty
	// then all resources of the given type will be synced.
	//
	// +optional
	Selector map[string]string `json:"selector,omitempty"`

        // SyncTLSSecrets is an on/off switch for syncing tls secrets to host cluster
	//
	// +kubebuilder:default=false
	// +required
	SyncTLSSecrets bool `json:"syncTLSSecrets"`
}
```

i can update the PR to include these changes if needed

k3k-kubelet/controller/syncer/ingress.go
```diff
@@ -92,11 +92,12 @@ func (r *IngressReconciler) Reconcile(ctx context.Context, req reconcile.Request
 		return reconcile.Result{}, err
 	}
 
+	syncConfig := cluster.Spec.Sync.Ingresses
```
enrichman: We need to check if a Policy.SyncConfig is available in the Status, because when a Policy is applied it has precedence over the one defined in the Cluster itself.
 https://github.com/rancher/k3k/blob/ef2bb0339a367e6dc8336005305e3979b46b967e/pkg/controller/policy/policy.go#L475-L482)

Basically we need to do something like this (feel free to change the implementation):

```suggestion
	appliedSync := cluster.Spec.Sync.DeepCopy()
	
	// If a policy is applied to the virtual cluster we need to use its SyncConfig, if available
	if cluster.Status.Policy != nil && cluster.Status.Policy.Sync != nil {
		appliedSync = cluster.Status.Policy.Sync
	}
	
	syncConfig = appliedSync.Ingresses
```

see also:

https://github.com/rancher/k3k/blob/ef2bb0339a367e6dc8336005305e3979b46b967e/pkg/controller/cluster/cluster.go#L719-L730

enrichman (changes requested):
Just a small fix

k3k-kubelet/controller/syncer/ingress.go
ibrokethecloud: i have updated the PR to include the feedback.

enrichman (approved):
Thank you! LGTM!

@galal-hussein can you give it a look?

galal-hussein (approved):

SantoDE:
Great work! Is this fix part of any version already? I don't think so, right?

enrichman:
> Great work! Is this fix part of any version already? I don't think so, right?

It will be in the next release (`v1.1.0`), that is expected to be released very soon!

PR #681, opened 2026-02-26
Add sync of Host StorageClasses

### LINKED ISSUE #557: UX sync storageclass
In `shared` mode, a virtual cluster user should be able to access storage classes to know which kind of storage are available on the host cluster.

Sync storageClasses from the host to the virtual cluster - should be an option

enrichman:
This PR fixes #557, adding the sync of the host StorageClasses in the Virtual Clusters.

When a Cluster enables the sync of the storage classes they will be copied into it. They can be filtered using a selector, so only the StorageClasses with the matching labels will be synced.

The label `"k3k.io/sync-enabled": "false"` can be used to avoid the sync of a particular host StorageClass.

#### Technical notes

In the controller an indexed Watch is added to the StorageClasses. When a StorageClass is updated it will trigger the Cluster reconciliation of only the Clusters that have the sync enabled from their spec, or forced by a policy.


pkg/controller/cluster/cluster.go
```diff
@@ -115,15 +124,85 @@ func Add(ctx context.Context, mgr manager.Manager, config *Config, maxConcurrent
...
 		Owns(&apps.StatefulSet{}).
 		Owns(&v1.Service{}).
 		WithOptions(ctrlcontroller.Options{MaxConcurrentReconciles: maxConcurrentReconciles}).
 		Complete(&reconciler)
 }
 
+func (r *ClusterReconciler) mapStorageClassToCluster(ctx context.Context, obj client.Object) []reconcile.Request {
+	log := ctrl.LoggerFrom(ctx)
+
+	_, ok := obj.(*storagev1.StorageClass)
```
galal-hussein:
```suggestion
	if _, ok := obj.(*storagev1.StorageClass); !ok {
	  return nil
	}
```

pkg/controller/cluster/cluster.go
```diff
@@ -115,15 +124,85 @@ func Add(ctx context.Context, mgr manager.Manager, config *Config, maxConcurrent
...
+	if !ok {
+		return nil
+	}
+
+	var specClusterList v1beta1.ClusterList
+	if err := r.Client.List(ctx, &specClusterList, client.MatchingFields{storageClassEnabledIndexField: "true"}); err != nil {
+		log.Error(err, "error listing clusters with spec sync enabled for storageclass sync")
```
galal-hussein, line 179: A question:

Would that skip reconciling that clusters forever if listing the clusters fails for that storage class?, and if so then shouldn't the Reconciler tries again if the listing fails?

pkg/controller/cluster/cluster.go
```diff
@@ -350,11 +429,22 @@ func (c *ClusterReconciler) reconcile(ctx context.Context, cluster *v1beta1.Clus
...
+
 	if err := c.ensureKubeconfigSecret(ctx, cluster, serviceIP, 443); err != nil {
 		return err
 	}
 
-	return c.bindClusterRoles(ctx, cluster)
+	// Important: if you need to call the Server API of the Virtual Cluster
+	// this needs to be done AFTER he kubeconfig has been generated
+
+	if err := c.ensureStorageClasses(ctx, cluster); err != nil {
```
galal-hussein, line 440:
```suggestion
	return c.ensureStorageClasses(ctx, cluster)
```

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -341,6 +346,21 @@ type PriorityClassSyncConfig struct {
 	Selector map[string]string `json:"selector,omitempty"`
 }
 
+// StorageClassSyncConfig specifies the sync options for services.
```
galal-hussein:
```suggestion
// StorageClassSyncConfig specifies the sync options for storageClasses.
```

tests/cluster_sync_storageclass_test.go
```diff
@@ -0,0 +1,160 @@
...
+					}
+				}
+			}).
+				MustPassRepeatedly(5).
+				WithPolling(time.Second).
+				WithTimeout(time.Second * 30).
+				Should(Succeed())
+		})
+
+		It("will not sync host storage classes with the sync disabled", func() {
```
galal-hussein, line 140: I think we should add also a test case where the synclabel does not exist

galal-hussein (changes requested):
@enrichman Looks very good to me, few changes suggested, I have an overall wondering for future Sync requests, should we have a catch all option in addition of the `SyncEnabledLabelKey`, its fine for storage classes, but if in the future we wanted to do bidirectional syncs for services for example, it might be complicated to loop over all of your resources and set the label

pkg/controller/cluster/cluster.go
enrichman: Yes, but just to be sure I have asked Gemini:

> if List fails, the `mapStorageClassToCluster` function returns nil, effectively dropping the event. Because this function is a MapFunc used by the controller's event handler, it does not support returning an error to trigger a retry of the event mapping itself.

So basically in this part we cannot force a retry. But I will change a little bit the code to try to perform both of the requests, before giving up!

pkg/controller/cluster/cluster.go
enrichman: If you don't mind I would prefer to keep it separate, but maybe I should add a log then, "wrapping" the errors? 👀

tests/cluster_sync_storageclass_test.go
enrichman: Yes, it's in the previous one. When the label is missing we will sync the StorageClass:

```go
if syncEnabled, found := hostSC.Labels[cluster.SyncEnabledLabelKey]; !found || syncEnabled == "true" {
			g.Expect(err).To(Not(HaveOccurred()))
}
```

galal-hussein (approved):
LGTM, Thank you!

Issue #682, opened 2026-03-02
Bug with mirrorHostNodes


jp-gouin:
**Environmental Info:**
Host Cluster Version: 
RKE2 v1.31.5+rke2r1

K3K Cluster Configuration:
K3k in `shared` mode with `mirrorHostNodes`
**Describe the bug:**
When using `mirrorHostNodes` the `virtual-kubelet` has the following error logs
```
2026-02-26T12:46:25.088Z	DPANIC	odd number of arguments passed as key-value pairs for logging	{"ignored key": "the cache is not started, can not read objects"}
2026-02-26T12:46:25.088Z	ERROR	error getting host node for mirroring	{"error": "the cache is not started, can not read objects"}
```

The virtual nodes are "empty"
```
# kubectl get nodes downstream2 -o yaml
apiVersion: v1
kind: Node
metadata:
  annotations:
    node.alpha.kubernetes.io/ttl: "0"
    virtual-kubelet.io/last-applied-node-status: '{"daemonEndpoints":{"kubeletEndpoint":{"Port":50311}},"nodeInfo":{"machineID":"","systemUUID":"","bootID":"","kernelVersion":"","osImage":"","containerRuntimeVersion":"","kubeletVersion":"","kubeProxyVersion":"","operatingSystem":"","architecture":""}}'
    virtual-kubelet.io/last-applied-object-meta: '{"name":"downstream2","uid":"5900fb47-1e7d-4d48-abc5-e8222ad06a5e","creationTimestamp":null}'
  creationTimestamp: "2026-02-20T16:05:36Z"
  finalizers:
  - wrangler.cattle.io/node
  - wrangler.cattle.io/managed-etcd-controller
  name: downstream2
  resourceVersion: "2843"
  uid: 5900fb47-1e7d-4d48-abc5-e8222ad06a5e
spec:
  podCIDR: 10.42.2.0/24
  podCIDRs:
  - 10.42.2.0/24
  providerID: rke2://downstream2
status:
  daemonEndpoints:
    kubeletEndpoint:
      Port: 50311
  nodeInfo:
    architecture: ""
    bootID: ""
    containerRuntimeVersion: ""
    kernelVersion: ""
    kubeProxyVersion: ""
    kubeletVersion: ""
    machineID: ""
    operatingSystem: ""
    osImage: ""
    systemUUID: ""
```

**Steps To Reproduce:**
- Created a cluster with `k3k create`: 
```
apiVersion: k3k.io/v1beta1
[...]

PR #686, opened 2026-03-03
Use APIReader instead of manager client for fetching node information in mirror mode

galal-hussein:
The PR uses apiReader instead of the host manager client when fetching the node in mirror host node mode, this avoid using the cache which has not been initialized yet
- #682

### DIFF
--- a/k3k-kubelet/provider/configure.go
+++ b/k3k-kubelet/provider/configure.go
@@ -6,6 +6,7 @@ import (
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


k3k-kubelet/provider/configure.go
```diff
@@ -6,20 +6,21 @@ import (
...
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/rancher/k3k/pkg/apis/k3k.io/v1beta1"
 )
 
-func ConfigureNode(logger logr.Logger, node *corev1.Node, hostname string, servicePort int, ip string, hostClient client.Client, virtualClient client.Client, virtualCluster v1beta1.Cluster, version string, mirrorHostNodes bool) {
+func ConfigureNode(logger logr.Logger, node *corev1.Node, hostname string, servicePort int, ip string, hostMgr manager.Manager, virtualClient client.Client, virtualCluster v1beta1.Cluster, version string, mirrorHostNodes bool) {
 	ctx := context.Background()
```
bigkevmcd, line 18: Should this just pass in `hostMgr.GetAPIReader()` rather than `hostMgr` ?

Issue #683, opened 2026-03-02
Bug with GPU allocation


jp-gouin:
**Environmental Info:**

K3K Cluster Configuration:
virtual cluster in `shared` mode with 2 nodes. One host nodes has GPUs not the other. 

**Describe the bug:**
When using VCP and quota to restrict the number of GPU allocatable, the quota can be applied to the wrong virtual nodes (aka `virtual-kubelet` assigned to the wrong host node).
This leads to GPU workload not schedulable. 

**Steps To Reproduce:**
- Created a cluster with `k3k create`: 

Use the following VCP
```
apiVersion: k3k.io/v1beta1
kind: VirtualClusterPolicy
metadata:
  name: quota-policy
spec:
  quota:
    hard:
      requests.nvidia.com/gpu: 4
```
Assign the policy to the `test102-5` namespace

Deploy the following cluster
```
apiVersion: k3k.io/v1beta1
kind: Cluster
metadata:
  name: k3kcluster-test102-5
  namespace: test102-5
spec:
  mode: shared
  servers: 2
  agents: 0
```

**Expected behavior:**
The virtual-kubelet assigned to GPU nodes should have the GPU allocatable part filled with the quota. 

**Actual behavior:**
GPU assignement is random across all host nodes.

dasarinaidu:
**Observation 1:** VCP GPU quota reduction not reflected in `virtual-kubelet` node Allocatable after update

When a VirtualClusterPolicy (VCP) GPU quota is reduced and the VCP is patched, the ResourceQuota object in the cluster namespace correctly reflects the new quota value. However, the virtual-kubelet node's Allocatable capacity does not update to reflect the reduced quota — it continues to show the original value even after the kubelet DaemonSet pods are restarted.

**Steps to Reproduce**
1 — Create VCP with GPU quota = 4:

```
apiVersion: k3k.io/v1beta1
kind: VirtualClusterPolicy
metadata:
  name: gpu-quota-update
spec:
  quota:
    hard:
      requests.nvidia.com/gpu: "4"
```
2 — Create namespace, label it, create shared-mode cluster:
```
kubectl create namespace k3k-gpu-update-test
kubectl label namespace k3k-gpu-update-test policy.k3k.io/policy-name=gpu-quota-update
```
3 — Verify baseline — virtual-kubelet GPU node shows nvidia.com/gpu: 4:
```
kubectl --kubeconfig=<vcluster-kubeconfig> describe node <gpu-node> | grep nvidia
[...]

PR #688, opened 2026-03-06
Refactor distribution algorithm to account for host capacity

galal-hussein:
The PR refactors the current distribution algorithm to account for the host capacity, the distribution algorithm follows these steps:

1. For each resource, build a sorted list of eligible nodes and look up the corresponding host capacity for that node
2. Distribute the resource between evenly between nodes
3. If the host capacity of that node exceed the assigned value then use the host capacity instead and remove the node from the eligible nodes.
4. If there is unallocated quota remaining (because some nodes were capped below their even share), repeat with the remaining nodes.
5. The loop terminates when the quota is fully distributed or no eligible nodes remaining

The new algorithm adds complexity of O(RxN<sup>2</sup>) since the distribution is a multi-pass process.

- #683

### DIFF
--- a/k3k-kubelet/provider/configure_capacity.go
+++ b/k3k-kubelet/provider/configure_capacity.go
@@ -100,6 +100,7 @@ func updateNodeCapacity(ctx context.Context, logger logr.Logger, hostClient clie
 		}
 
 		hostResourceMap := make(map[string]corev1.ResourceList)
+
 		for _, hNode := range hostNodeList.Items {
 			if _, ok := virtResourceMap[hNode.Name]; ok {
 				hostResourceMap[hNode.Name] = hNode.Status.Allocatable


k3k-kubelet/provider/configure_capacity.go
```diff
@@ -125,74 +125,122 @@ func mergeResourceLists(resourceLists ...corev1.ResourceList) corev1.ResourceLis
...
+
+			hostCap[vn.Name] = resourceQuantity.Value()
+			if useMilli {
+				hostCap[vn.Name] = resourceQuantity.MilliValue()
 			}
+
+			eligibleNodes = append(eligibleNodes, vn.Name)
+		}
+
+		sort.Strings(eligibleNodes)
```
bigkevmcd, line 196: Why sort the eligibleNodes alphabetically?

Doesn't this mean that distribution is initially weighted (arbitrarily) to the alpha-sorted nodes?

Should this perhaps just shuffle to distribute more fairly?

k3k-kubelet/provider/configure_capacity.go
```diff
@@ -125,74 +125,122 @@ func mergeResourceLists(resourceLists ...corev1.ResourceList) corev1.ResourceLis
...
-		// Use MilliValue for precise division, especially for resources like CPU,
-		// which are often expressed in milli-units. Otherwise, use the standard Value().
-		var totalValue int64
-		if _, found := milliScaleResources[resourceName]; found {
-			totalValue = totalQuantity.MilliValue()
-		} else {
-			totalValue = totalQuantity.Value()
+		useMilli := false
+		if _, ok := milliScaleResources[resourceName]; ok {
+			useMilli = true
```
andreas-kupries: Simplifiable to
```suggestion
		_, useMilli := milliScaleResources[resourceName]
```

andreas-kupries: The highlighted block is one line to small, I failed to capture the closing brace `}`.

k3k-kubelet/provider/configure_capacity.go
galal-hussein: Good question, that was the original behavior that Enrico added for distribution, actually it might be an optimization to sort by host capacity, and break the cycles if the capacity is exceeded guaranteeing that all the rest will be assigned their host capacity which will bring down the complexity to O(NLogN) probably

k3k-kubelet/provider/configure_capacity.go
```diff
@@ -125,74 +135,107 @@ func mergeResourceLists(resourceLists ...corev1.ResourceList) corev1.ResourceLis
...
+// For each resource type the algorithm uses a multi-pass redistribution loop:
+//  1. Divide the remaining quota evenly among eligible nodes (sorted by name for
+//     determinism), assigning any integer remainder to the first nodes alphabetically.
+//  2. Cap each node's share at its host allocatable capacity.
+//  3. Remove nodes that have reached their host capacity.
+//  4. If there is still unallocated quota (because some nodes were capped below their
+//     even share), repeat from step 1 with the remaining quota and remaining nodes.
+//
+// The loop terminates when the quota is fully distributed or no eligible nodes remain.
+func distributeQuotas(ctx context.Context, logger logr.Logger, hostNodeList, virtualNodeList corev1.NodeList, quotas corev1.ResourceList) (map[string]corev1.ResourceList, error) {
```
enrichman: Having a look at the func and how we use the nodes we can probably simplify it even more, passing not the `corev1.NodeList` but the two `map[string]corev1.ResourceList` for the hostNodes and virtualKubeletNodes:

```go
func distributeQuotas(hostResourceMap, virtResourceMap map[string]corev1.ResourceList, quotas corev1.ResourceList) (map[string]corev1.ResourceList, error)
```

The context and logger are not needed anymore.

k3k-kubelet/provider/configure_capacity.go
```diff
@@ -125,74 +135,107 @@ func mergeResourceLists(resourceLists ...corev1.ResourceList) corev1.ResourceLis
...
+		resourceMap[vn.Name] = corev1.ResourceList{}
+	}
+
+	hostNodeMap := make(map[string]*corev1.Node, len(hostNodeList.Items))
+	for i := range hostNodeList.Items {
+		name := hostNodeList.Items[i].Name
+		if _, ok := resourceMap[name]; ok {
+			hostNodeMap[name] = &hostNodeList.Items[i]
+		}
 	}
```
enrichman: This part could be dropped, if we pass the map from the outer.

k3k-kubelet/provider/configure_capacity.go
```diff
@@ -83,7 +83,17 @@ func updateNodeCapacity(ctx context.Context, logger logr.Logger, hostClient clie
...
+
+		if err := virtualClient.List(ctx, &virtualNodeList); err != nil {
+			logger.Error(err, "error listing virtual nodes for stable capacity distribution")
+		}
+
+		if err := hostClient.List(ctx, &hostNodeList); err != nil {
+			logger.Error(err, "error listing host nodes for stable capacity distribution")
+		}
+
+		m, err := distributeQuotas(ctx, logger, hostNodeList, virtualNodeList, mergedResourceLists)
```
enrichman: We can extract the virtual and host nodes map here, so we can just pass a simpler `map[string]corev1.ResourceList`:

```suggestion
		// build virtual nodes resource map
		var virtualNodeList corev1.NodeList
		if err := virtualClient.List(ctx, &virtualNodeList); err != nil {
			logger.Error(err, "error listing virtual nodes for stable capacity distribution")
		}

		virtNodeMap := make(map[string]corev1.ResourceList)
		for _, virtNode := range virtualNodeList.Items {
			virtNodeMap[virtNode.Name] = virtNode.Status.Allocatable
		}

		// build host nodes resource map
		var hostNodeList corev1.NodeList
		if err := hostClient.List(ctx, &hostNodeList); err != nil {
			logger.Error(err, "error listing host nodes for stable capacity distribution")
		}

		hostNodeMap := make(map[string]corev1.ResourceList)
		for _, hostNode := range hostNodeList.Items {
			hostNodeMap[hostNode.Name] = hostNode.Status.Allocatable
		}

		m, err := distributeQuotas(hostNodeMap, virtNodeMap, mergedResourceLists)
```

k3k-kubelet/provider/configure_capacity_test.go
```diff
@@ -85,43 +91,135 @@ func Test_distributeQuotas(t *testing.T) {
...
+				"nvidia.com/gpu": resource.MustParse("10"),
+			},
+			// Total host capacity is 4, quota is 10. Each node gets its full capacity.
+			want: map[string]corev1.ResourceList{
+				"node-1": {"nvidia.com/gpu": resource.MustParse("2")},
+				"node-2": {"nvidia.com/gpu": resource.MustParse("1")},
+				"node-3": {"nvidia.com/gpu": resource.MustParse("1")},
+			},
+			wantErr: false,
+		},
```
enrichman, line 273: Tests will be easier to write and understand, they will be something like this:


```suggestion
		{
			name: "quota exceeds total host capacity",
			virtualNodes: map[string]corev1.ResourceList{
				"node-1": {},
				"node-2": {},
				"node-3": {},
			},
			hostNodes: map[string]corev1.ResourceList{
				"node-1": {
					"nvidia.com/gpu": resource.MustParse("2"),
				},
				"node-2": {
					"nvidia.com/gpu": resource.MustParse("1"),
				},
				"node-3": {
					"nvidia.com/gpu": resource.MustParse("1"),
				},
			},
			quotas: corev1.ResourceList{
				"nvidia.com/gpu": resource.MustParse("10"),
			},
			// Total host capacity is 4, quota is 10. Each node gets its full capacity.
			want: map[string]corev1.ResourceList{
				"node-1": {"nvidia.com/gpu": resource.MustParse("2")},
				"node-2": {"nvidia.com/gpu": resource.MustParse("1")},
				"node-3": {"nvidia.com/gpu": resource.MustParse("1")},
			},
			wantErr: false,
		},
```

k3k-kubelet/provider/configure_capacity.go
```diff
@@ -89,15 +90,23 @@ func updateNodeCapacity(ctx context.Context, logger logr.Logger, hostClient clie
...
 
-		m, err := distributeQuotas(ctx, logger, hostNodeList, virtualNodeList, mergedResourceLists)
-		if err != nil {
-			logger.Error(err, "error distributing policy quota")
+		hostResourceMap := make(map[string]corev1.ResourceList, len(hostNodeList.Items))
+		for _, hNode := range hostNodeList.Items {
+			if _, ok := virtResourceMap[hNode.Name]; ok {
+				hostResourceMap[hNode.Name] = hNode.Status.Allocatable
+			}
 		}
```
enrichman: We are initializing the map with the len of the hostNodes, but we are then adding probably less elements to it. Not a big issue, but in this case it's probably cleaner to initialize it without the capacity (?).

k3k-kubelet/provider/configure_capacity.go
```diff
@@ -89,15 +90,23 @@ func updateNodeCapacity(ctx context.Context, logger logr.Logger, hostClient clie
...
 
+		m := distributeQuotas(hostResourceMap, hostResourceMap, mergedResourceLists)
```
enrichman: Copy paste typo! ⚠️ 

```suggestion
		m := distributeQuotas(hostResourceMap, virtResourceMap, mergedResourceLists)
```

k3k-kubelet/provider/configure_capacity.go
```diff
@@ -149,48 +158,38 @@ func mergeResourceLists(resourceLists ...corev1.ResourceList) corev1.ResourceLis
...
-			if hostNode == nil {
+		for vn := range virtResourceMap {
+			hostNodeResources := hostResourceMap[vn]
+			if hostNodeResources == nil {
 				continue
 			}
 
-			resourceQuantity := hostNode.Status.Allocatable[resourceName]
+			resourceQuantity, ok := hostNodeResources[resourceName]
+			if !ok {
```
enrichman: Nit: usually `ok` is used with type assertions, while with maps `found` is used.

Issue #522, opened 2025-10-21
Affinity/Anti-affinity


jp-gouin:
User should be able to setup affinity and anti affinity at the cluster level for the agent, server pods and virtual kubelet independently

derhornspieler:
**Update:** We've implemented a sidecar-based auto-rebalancing workaround for this.

Since there's no affinity/anti-affinity support in the k3k CRD, our [ingress-watcher](https://github.com/derhornspieler/rancher-k3k/blob/main/ingress-watcher.yaml) now detects when multiple server pods land on the same node after a drain/uncordon and deletes the highest-ordinal pod so the scheduler redistributes it. The Kubernetes default `PodTopologySpread` scoring for StatefulSets handles the actual placement.

**How it works:**
1. Every 10s poll, builds a node→pod map from `pod.spec.nodeName`
2. If any node has >1 server pod AND enough schedulable (uncordoned) nodes exist → triggers rebalance
3. Deletes the highest-ordinal pod on the most-loaded node (protects `server-0` as etcd init node)
4. Debounce (5min between attempts) and cooldown (30min after failed move) prevent delete loops
5. Skips if any pod became Ready <60s ago (avoids race with active drains)

**Tested scenario:** Cordon node → server-0 moves to another node alongside server-1 → uncordon → rebalancer deletes server-1 (not server-0) → scheduler places it on the empty node → balanced, Rancher healthy.

This works but is a workaround. Native `spec.serverAffinity` / `spec.agentAffinity` in the CRD (as suggested above) would be the proper fix since it would prevent imbalance at scheduling time rather than detecting and correcting it after the fact.

---

> This comment was authored by [@derhornspieler](https://github.com/derhornspieler) with assistance from [Claude Code](https://claude.ai/claude-code) (Anthropic).

Priyashetty17:
Validated the tests for this feature with:
- Host Cluster K8s Version: 1.35.2+rke2r1
- K3KCLI Version: 1.1.0-rc3
- K3K Helm Chart Version: 1.1.0-rc3

Validated the following test scenarios (detailed test plan available in Qase) in both **Shared** and **Virtual** Mode K3K clusters:
- Server and agent affinity defined in Cluster spec works correctly (no VCP). 
- VCP applies default affinity rules when Cluster spec has no affinity. 
- Affinity is enforced via VCP when creating clusters using K3KCLI+policy. 
- VCP affinity overrides Cluster spec affinity. 
[...]

PR #696, opened 2026-03-12
Add initial affinity to podspecs

JonCrowther:
Issue https://github.com/rancher/k3k/issues/522

Add a new field to both the `VirtualClusterPolicy` and `Cluster` CRDs called `ServerAffinity` and `AgentAffinity`. These are the affinity to apply to the server and agent pods respectively.

If they are specified in the `VirtualClusterPolicy`, it's added to the status of the `Cluster` and takes priority over any affinity specified in the `Cluster.spec`

### DIFF
--- a/docs/cli/k3kcli.adoc
+++ b/docs/cli/k3kcli.adoc
@@ -1,8 +1,8 @@
-=== k3kcli
+== k3kcli
 
 CLI for K3K.
 
-==== Options
+=== Options
 
 ----
       --debug               Turn on debug logs
@@ -10,24 +10,24 @@ CLI for K3K.
       --kubeconfig string   kubeconfig path ($HOME/.kube/config or $KUBECONFIG if set)
 ----
 
-=== k3kcli cluster
+== k3kcli cluster
 
 K3k cluster command.
 
-==== Options
+=== Options
 
 ----
   -h, --help   help for cluster
 ----
 
-==== Options inherited from parent commands
+=== Options inherited from parent commands
 
 ----
       --debug               Turn on debug logs
       --kubeconfig string   kubeconfig path ($HOME/.kube/config or $KUBECONFIG if set)
 ----
 
-=== k3kcli cluster create
+== k3kcli cluster create
 
 Create a new cluster.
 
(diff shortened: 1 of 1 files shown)


pkg/controller/cluster/agent/shared.go
```diff
@@ -165,7 +165,14 @@ func (s *SharedAgent) podSpec() v1.PodSpec {
 		image = s.imageRegistry + "/" + s.image
 	}
 
+	// Use the agent affinity from the policy status if it exists, otherwise fall back to the spec.
+	agentAffinity := s.cluster.Spec.AgentAffinity
+	if s.cluster.Status.Policy.AgentAffinity != nil {
+		agentAffinity = s.cluster.Status.Policy.AgentAffinity
```
galal-hussein, line 175: nit: should we add a warning log in the controller stating that policy affinity has override the cluster spec

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -605,6 +615,16 @@ type AppliedPolicy struct {
 	// +optional
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 
+	// serverAffinity is the affinity rules for server pods enforced by the active VirtualClusterPolicy.
```
galal-hussein, line 620: We probably need to clarify what kind of affinity are we talking about here, is it node affinity or inter-pod affinity, or both

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -605,6 +615,16 @@ type AppliedPolicy struct {
...
+	//
+	// +optional
+	ServerAffinity *corev1.Affinity `json:"serverAffinity,omitempty"`
+
+	// agentAffinity is the affinity rules for agent pods enforced by the active VirtualClusterPolicy.
```
galal-hussein, line 626: same as server affinity note

pkg/controller/cluster/agent/shared.go
JonCrowther: Added warnings each time we override the cluster spec

PR #704, opened 2026-03-17
Change build and generate scripts for arm64 compatibility

JonCrowther:
In the build script, I added a check to see if `build_args` is empty, because otherwise it throws a `unbound variable` error.
In the generate script, I changed it from `GNU sed` (which doesn't work on macOS since the default is `BSD sed`) to use `yq`. It's also a little easier to read IMO.

If these changes don't work, I'll guard them with an `if` statement.

### DIFF
--- a/charts/k3k/templates/crds/k3k.io_clusters.yaml
+++ b/charts/k3k/templates/crds/k3k.io_clusters.yaml
@@ -15,1073 +15,1073 @@ spec:
     singular: cluster
   scope: Namespaced
   versions:
-    - additionalPrinterColumns:
-        - jsonPath: .spec.mode
-          name: Mode
-          type: string
-        - jsonPath: .status.phase
-          name: Status
-          type: string
-        - jsonPath: .status.policyName
-          name: Policy
-          type: string
-      name: v1beta1
-      schema:
-        openAPIV3Schema:
-          description: |-
-            Cluster defines a virtual Kubernetes cluster managed by k3k.
-            It specifies the desired state of a virtual cluster, including version, node configuration, and networking.
-            k3k uses this to provision and manage these virtual clusters.
-          properties:
-            apiVersion:
-              description: |-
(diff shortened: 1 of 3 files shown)


charts/k3k/templates/crds/k3k.io_clusters.yaml
JonCrowther, line 1: I think by using `yq`, it redid the spacing in the generated yaml

scripts/generate
```diff
@@ -14,7 +14,6 @@ go run sigs.k8s.io/controller-tools/cmd/controller-gen@${CONTROLLER_TOOLS_VERSIO
 
 # add the 'helm.sh/resource-policy: keep' annotation to the CRDs
 for f in ./charts/k3k/templates/crds/*.yaml; do
-    sed -i '0,/^[[:space:]]*annotations:/s/^[[:space:]]*annotations:/&\n    helm.sh\/resource-policy: keep/' "$f"
-    echo "Validating $f"
-    yq . "$f" > /dev/null
+    echo "Annotating $f"
+    yq -i '.metadata.annotations["helm.sh/resource-policy"] = "keep"' "$f"
```
enrichman: I love `yq`, it looks very clean. But the indentation changes are so noisy.
Can you try if this `perl` command works for you? It does on linux.


```suggestion
    echo "Annotating $f"
    perl -i -pe 'if (!$done && s/^(\s*annotations:)/$1\n    helm.sh\/resource-policy: keep/) { $done = 1 }' "$f"
    yq . "$f" > /dev/null
```

I know it looks kinda magic.. 🧙‍♂️

JonCrowther: This command does work on macOS.

I added the `-c` option to the yq command and it maintains the indentation. It does update some of the lines to no longer wrap, but that doesn't seem like a big difference. I personally would prefer to keep just the yq line if possible since it is a lot simpler.

Issue #589, opened 2025-12-10
Refactoring Testing framework for the repo


galal-hussein:
Currently the e2e and cli tests are combined under `tests/` dir, also the some tests are using envtest in both `k3k-kubelet/controller/syncer` and `pkg/controller/[cluster|policy]` which can considered to be integration tests that are also using the ginkgo test suite to test the integration between some of the controllers and the API, the remaining are unit tests scattered through our package.

This is not the ideal setup for our testing framework, a suggestion is to use more organized structure that can facilitate our testing for the code:

```sh
tests/
├── e2e #all of our e2e tests
├── integration # all integration tests related to controllers
├── cli # k3kcli tests
....
└── tests_suite_test.go
```
The unit tests should only be contained in the respective package, and shouldn't not contain the Ginkgo test suite there, only testing the functions in the packages.

PR #723, opened 2026-03-24
Refactor tests to their own directories

JonCrowther:
Addressing https://github.com/rancher/k3k/issues/589

Working on improving the layout of the tests. I'm also trying to keep the Makefile working so you can individual kick off any of the tests.

```
tests
-> cli
-> e2e
-> integration
---> cluster
---> policy
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,15 +64,15 @@ test-unit:	## Run the unit tests (skips the e2e and integration tests)
 	$(GINKGO) $(GINKGO_FLAGS) --skip-file=tests/*
 
 .PHONY: test-kubelet
-test-kubelet:	## Run the k3k-kubelet tests (tests/integration/k3k-kubelet)
+test-kubelet:	## Run the k3k-kubelet controller tests (tests/integration/k3k-kubelet)
 	$(GINKGO) $(GINKGO_FLAGS) tests/integration/k3k-kubelet
 
 .PHONY: test-policy
-test-policy:	## Run the policy tests (tests/integration/policy)
+test-policy:	## Run the policy controller tests (tests/integration/policy)
 	$(GINKGO) $(GINKGO_FLAGS) tests/integration/policy
 
 .PHONY: test-cluster
-test-cluster:	## Run the cluster tests (tests/integration/cluster)
+test-cluster:	## Run the cluster controller tests (tests/integration/cluster)
 	$(GINKGO) $(GINKGO_FLAGS) tests/integration/cluster
 
 .PHONY: test-integration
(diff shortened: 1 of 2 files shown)


tests/cli/common_test.go
JonCrowther, line 1: This is hopefully temporary. `cli_test.go` made use of a lot of the same shared functionality of the e2e tests. I copied it over for now while I'm splitting things up.

Makefile
```diff
@@ -63,21 +63,21 @@ test:	## Run all the tests
 test-unit:	## Run the unit tests (skips the e2e)
 	$(GINKGO) $(GINKGO_FLAGS) --skip-file=tests/*
 
-.PHONY: test-controller
-test-controller:	## Run the controller tests (pkg/controller)
-	$(GINKGO) $(GINKGO_FLAGS) pkg/controller
-
```
enrichman, line 69: Why was this removed?

Makefile
```diff
@@ -63,21 +63,21 @@ test:	## Run all the tests
...
 
 .PHONY: test-e2e
 test-e2e:	## Run the e2e tests
-	$(GINKGO) $(GINKGO_FLAGS) --label-filter="$(E2E_LABEL_FILTER)" tests
+	$(GINKGO) $(GINKGO_FLAGS) --label-filter="$(E2E_LABEL_FILTER)" tests/e2e
 
 .PHONY: test-cli
 test-cli:	## Run the cli tests
-	$(GINKGO) $(GINKGO_FLAGS) --label-filter=cli --flake-attempts=3 tests
+	$(GINKGO) $(GINKGO_FLAGS) --label-filter=cli --flake-attempts=3 tests/cli
```
enrichman: Since these are now in their dedicated folder we can skip the `cli` label for them.

```suggestion
	$(GINKGO) $(GINKGO_FLAGS) --flake-attempts=3 tests/cli
```

Makefile
JonCrowther: Now that we've moved the ginkgo tests out of that directory, I figured we'd want to do the integration tests instead. To re-create this, we'd test the following:

- `tests/integration/cluster`
- `tests/integration/policy`
- unit tests in `pkg/controller`

Not opposed to that as a set of tests, but I tried to break the make commands into doing individual directories.

What about something like `test-cluster`, `test-policy` and `test-k3k-kubelet`? That way you can specify them individually, or use `test-integration` for all the `envtest` tests to run

tests/cli/common_test.go
JonCrowther: I was able to remove most of the copied code

Makefile
enrichman: Ah sorry, I missed that part! So basically the tests involving envtest are under the integration folder? Much better, thanks!

JonCrowther: Yeah, I feel like that's an appropriate name for those tests. I've added some more granular options to the Makefile. Once this is merged I can maybe revisit the CI to see how we want to split the tests in the github actions.

PR #724, opened 2026-03-26
Schedule Pods in the same Node with a preferred affinity

### LINKED ISSUE #727: Cannot schedule Deployment with PVCs
With PR https://github.com/rancher/k3k/pull/610 we made the change on _where_ the Pod are being scheduled, to have possibly a 1:1 match with the virtual cluster.

In some cases this is not working:
> If you choose to use WaitForFirstConsumer, do not use nodeName in the Pod spec to specify node affinity. If nodeName is used in this case, the scheduler will be bypassed and PVC will remain in pending state.

Also another ref: https://github.com/openebs/openebs/issues/2915

To replicate the issue (using the local-path provisioner):

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: sample-pvc
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: sample-app
spec:
  replicas: 3
  selector:
    matchLabels:
      app: sample-app
  template:
    metadata:
      labels:
        app: sample-app
    spec:
      containers:
      - name: nginx
        image: nginx
        volumeMounts:
        - name: data-volume
          mountPath: /data
      volumes:
      - name: data-volume
        persistentVolumeClaim:
          claimName: sample-pvc
```

enrichman:
This PR fixes #727 adding a `PreferredDuringSchedulingIgnoredDuringExecution` Affinity to the Pod, instead of an hardcoded `nodeName`.

Also adding some tests, with Deployment and StatefulSet, to avoid regressions.

### DIFF
--- a/tests/e2e/cluster_pod_test.go
+++ b/tests/e2e/cluster_pod_test.go
@@ -59,7 +59,7 @@ var _ = Context("In a shared cluster", Label(e2eTestLabel), Ordered, func() {
 			Expect(err).To(Not(HaveOccurred()))
 		})
 
-		It("should have a default Affinity", func() {
+		It("should have the default Affinity", func() {
 			ctx := context.Background()
 
 			Eventually(func(g Gomega) {
@@ -121,7 +121,7 @@ var _ = Context("In a shared cluster", Label(e2eTestLabel), Ordered, func() {
 			Expect(err).To(Not(HaveOccurred()))
 		})
 
-		It("should have a default Affinity", func() {
+		It("should not have the default Affinity", func() {
 			ctx := context.Background()
 
 			Eventually(func(g Gomega) {


tests/e2e/cluster_pod_test.go
```diff
@@ -20,16 +20,132 @@ import (
...
+						},
+					},
+				},
+			}
+
+			pod, err = virtualCluster.Client.CoreV1().Pods(pod.Namespace).Create(ctx, pod, metav1.CreateOptions{})
+			Expect(err).To(Not(HaveOccurred()))
+		})
+
+		It("should have a default Affinity", func() {
```
JonCrowther: I think you meant 
```suggestion
		It("should not have the default Affinity", func() {
```

PR #741, opened 2026-04-03
Cleanup tests and small fixes

enrichman:
This PR address few fixes in our tests, and cleanup.

### Fix for wrong exported var

This was just echoing the whole line:
```
echo "K3S_HOST_VERSION=v1.32.1+k3s1 >> $GITHUB_ENV"
```

### Common `KUBERNETES_VERSION` var

With a common `KUBERNETES_VERSION` var it will be easier to upgrade the version everywhere, and possibly automate this. This is now also used to create the k3d cluster for the conformance tests with the correct version.

### Missing k3k logs dump

Cli tests were missing the k3k server logs dump, so the archive was failing, and it was not possible to check them on the CI.

### Helm installation

We are now using directly the helm releases to install Helm, instead of the `azure/setup-helm` action.

```yaml
env:
  HELM_VERSION: v4.1.3
  HELM_BIN_HASH_AMD64: 02ce9722d541238f81459938b84cf47df2fdf1187493b4bfb2346754d82a4700

...

    - name: Install helm
      run: |
        curl -sSfL -o helm.tar.gz https://get.helm.sh/helm-${{ env.HELM_VERSION }}-linux-amd64.tar.gz
        echo "${{ env.HELM_BIN_HASH_AMD64 }}  helm.tar.gz" | shasum -a 256 --check
        tar -xvzf helm.tar.gz --strip-components=1 -C /tmp/
        sudo mv /tmp/helm /usr/local/bin
        sudo chmod +x /usr/local/bin/helm
```

### DIFF
--- a/.github/workflows/test-conformance-shared.yaml
+++ b/.github/workflows/test-conformance-shared.yaml
@@ -41,7 +41,7 @@ jobs:
     - name: Install helm
       run: |
         curl https://get.helm.sh/helm-${{ env.HELM_VERSION }}-linux-amd64.tar.gz | tar xvzf - --strip-components=1 -C /tmp/
-        echo "${{ env.HELM_BIN_HASH_AMD64 }}  /tmp/helm" | sha256sum -a 256 --check
+        echo "${{ env.HELM_BIN_HASH_AMD64 }}  /tmp/helm" | shasum -a 256 --check
         sudo mv /tmp/helm /usr/local/bin
         sudo chmod +x /usr/local/bin/helm
 
--- a/.github/workflows/test-conformance-shared.yaml
+++ b/.github/workflows/test-conformance-shared.yaml
@@ -11,6 +11,10 @@ permissions:
 env:
   ARCH: amd64
   KUBERNETES_VERSION: v1.35.3
+  HELM_VERSION: v4.1.3
+  HELM_BIN_HASH_AMD64: 02ce9722d541238f81459938b84cf47df2fdf1187493b4bfb2346754d82a4700
+  K3D_VERSION: v5.8.3
+  K3D_BIN_HASH_AMD64: dbaa79a76ace7f4ca230a1ff41dc7d8a5036a8ad0309e9c54f9bf3836dbe853e
 
 jobs:
   conformance:
@@ -22,9 +26,7 @@ jobs:
         type:
         - parallel
         - serial
-    env:
-      K3D_VERSION: v5.8.3
-      K3D_BIN_HASH_AMD64: "dbaa79a76ace7f4ca230a1ff41dc7d8a5036a8ad0309e9c54f9bf3836dbe853e"
+      
     steps:
     - name: Checkout code
       uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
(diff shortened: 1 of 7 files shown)


.github/workflows/test-conformance-shared.yaml
```diff
@@ -37,21 +43,26 @@ jobs:
...
-        echo "${K3D_BIN_HASH_AMD64}  k3d" | shasum -a 256 --check
+        curl -sSfL -o k3d "https://github.com/k3d-io/k3d/releases/download/${{ env.K3D_VERSION }}/k3d-linux-${ARCH}"
+        echo "${{ env.K3D_BIN_HASH_AMD64 }}  k3d" | shasum -a 256 --check
         sudo mv k3d /usr/local/bin
         sudo chmod +x /usr/local/bin/k3d
 
     - name: Install k3d and kubectl
       run: |
-        curl -LO "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl"
+        curl -LO "https://dl.k8s.io/release/${{ env.KUBERNETES_VERSION }}/bin/linux/amd64/kubectl"
```
pmatseykanets, line 65: Should we verify the hash here

.github/workflows/test-conformance-virtual.yaml
```diff
@@ -31,15 +40,20 @@ jobs:
...
     - name: Install hydrophone
       run: go install sigs.k8s.io/hydrophone@3de3e886a2f6f09635d8b981c195490af1584d97 #v0.7.0
 
     - name: Install k3s
       env:
         KUBECONFIG: /etc/rancher/k3s/k3s.yaml
-        K3S_HOST_VERSION: v1.35.2+k3s1
+        K3S_HOST_VERSION: ${{ env.KUBERNETES_VERSION }}+k3s1
       run: |
         curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=${K3S_HOST_VERSION} INSTALL_K3S_EXEC="--write-kubeconfig-mode=777" sh -s -
```
pmatseykanets, line 58: Here and in other places the install script itself is not pinned.

.github/workflows/test-e2e.yaml
pmatseykanets, line 95: Both tests-e2e and tests-e2e-slow jobs upload artifacts with identical name `e2e-k3s-logs`.
Maybe change the name for slow job to e2e-slow-k3s-logs?

pmatseykanets, line 102: Same here.

pmatseykanets: See https://github.com/rancher/k3k/pull/741/changes#r3032975590

pmatseykanets: See https://github.com/rancher/k3k/pull/741/changes#r3032977356

tests/cli/tests_suite_test.go
```diff
@@ -216,10 +217,28 @@ var _ = AfterSuite(func() {
...
 	}
 })
 
+func getK3kLogs(ctx context.Context) io.ReadCloser {
+	var podList v1.PodList
+
+	err := k8sClient.List(ctx, &podList, &client.ListOptions{Namespace: k3kNamespace})
+	Expect(err).To(Not(HaveOccurred()))
+
+	k3kPod := podList.Items[0]
```
pmatseykanets, line 235: This will panic the `Items` is empty. Maybe guard with

```
Expect(podList.Items).NotTo(BeEmpty())
```

.github/workflows/test-conformance-shared.yaml
enrichman: Yes, thanks! I will follow the instruction from here: https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/

.github/workflows/test-conformance-virtual.yaml
enrichman: I am not sure if that's possible. How can we pin it?

.github/workflows/test-e2e.yaml
enrichman: Ah, good catch. This part was failing because we were not getting the logs from anywhere. It's a leftover from an old test implementation done with testcointainers.

I'll try to fix them.

.github/workflows/test-conformance-virtual.yaml
pmatseykanets: Get the install script at a specific sha and then run it? Not sure what k3s suggests TBH.

enrichman: I don't think the script is versioned. Even in airgapped environment they suggest you to download it: https://docs.k3s.io/installation/airgap#download-binary-and-script

But looking at the docs it seems we can use directly the binary, if we don't need to install it as a system service: https://docs.k3s.io/installation/configuration#configuration-with-binary

I will try!

enrichman: @pmatseykanets it's seems a bit complicated. We could copy the script, but it seems overkill. :/

pmatseykanets (approved):

PR #769, opened 2026-04-09
Refactoring of tests moving common logic in `tests/framework`

enrichman:
This PR extract all the duplicated test setup code into a shared `tests/framework/` package, removing a bit of of copy-pasted logic across integration, CLI, and E2E tests.

```
  tests/framework/
  ├── client/          # Kubernetes client setup, scheme building, REST config
  ├── container/       # Testcontainers and K3s setup, Helm chart installation
  ├── env/             # Envtest environment helpers for integration tests
  ├── k3k/             # K3k-specific utilities (namespace management, etc.)
  └── log/             # Log fetching and diagnostic utilities
```


tests/cli/common_test.go
JonCrowther, line 1: Should we just get rid of this file and replace any instance of `NewNamespace` with `fwk3k.CreateNamespace` and `DeleteNamespace` with `fwk3k.DeleteNamespaces`?

tests/e2e/common_test.go
```diff
@@ -91,41 +88,11 @@ func NewVirtualClusters(n int) []*VirtualCluster {
 }
 
 func NewNamespace() *v1.Namespace {
```
JonCrowther: Same here, these could probably be fully replaced.

tests/cli/tests_suite_test.go
```diff
@@ -215,40 +95,12 @@ var _ = AfterSuite(func() {
 		// dump k3s logs
 		k3sLogs, err := k3sContainer.Logs(ctx)
 		Expect(err).To(Not(HaveOccurred()))
-		writeLogs("k3s.log", k3sLogs)
+		fwlog.WriteToTemp("k3s.log", k3sLogs)
```
galal-hussein: Do we need to change its name to writeToTemp, I actually liked writeLogs, which points more to what it does

JonCrowther (commented):
Overall looks really good. Just a quick question about the need for functions that just call another common function

tests/framework/client/scheme.go
```diff
@@ -0,0 +1,58 @@
...
+	if err := v1beta1.AddToScheme(scheme); err != nil {
+		panic(err)
+	}
+
+	return scheme
+}
+
+// NewPolicyScheme creates a new Kubernetes runtime scheme with core APIs, networking APIs,
+// and k3k CRDs. This is required for policy controller tests that need apps/v1 and networking/v1.
+func NewPolicyScheme() *runtime.Scheme {
```
galal-hussein: Since we are refactoring, Can we use `NewScheme` instead as a scheme to have all these types and use it throughout the tests

galal-hussein (commented):
Looks really good, only couple of nits

tests/cli/common_test.go
enrichman: Not really, it was because the PR was going to be a little bit less bloated by the changes, but I'm happy to change that!

tests/cli/tests_suite_test.go
enrichman: Hm, it's probably because on what it does, but yeah, I don't like it a lot either.

What about changing it to just `Write`? Since it's in the `log` package it's implied that we are writing logs, and you could use it to write anything actually.

```go
fwlog.Write("k3s.log", k3sLogs)
```

tests/framework/client/scheme.go
enrichman: Absolutely, it was probably a leftover. Thanks!

JonCrowther (approved):

PR #774, opened 2026-04-10
Fix nil error handling in webhook removal

enrichman:
This PR fixes a bug regarding Webhook removal and cleanup.

Previously, the logic only checked if the error was not a `NotFound` error. This meant the code entered the error-handling branch even when the error was nil. While a restart would eventually resolve the state (as the error would no longer be nil), adding an explicit check for nil fixes the root cause immediately.


enrichman:
Thanks @joesims22 for spotting this out!

galal-hussein (approved):

JonCrowther (approved):

Issue #563, opened 2025-11-13
Deleting a policy is not removing the label to the namespace


enrichman:
After the deletion of a Policy the namespace is still referencing it.

<img width="685" height="295" alt="Image" src="https://github.com/user-attachments/assets/226a9560-8697-409d-b80b-f86afc6421cf" />

Priyashetty17:
Validated the tests for this feature with:
- Host Cluster RKE2 Version: 1.35.2+rke2r1
- K3KCLI Version: 1.1.0-rc4
- K3K Helm Chart Version: 1.1.0-rc4

Validated the following test scenarios (detailed test plan available in Qase) in **Shared Mode** and **Virtual Mode** K3K clusters:
- Delete VCP cleans references from the namespace and cluster
- Delete VCP cleans references from multiple namespaces
- Recreate same-name VCP does not reuse stale references
- Delete VCP via K3K CLI
- Upgrade Check - Delete VCP cleans references from namespaces and cluster

PR #781, opened 2026-04-13
Added Policy finalizer

### LINKED ISSUE #703: Policy Name Persistence Causes Unintended Cluster Re-Association
Upon deletion of a Virtual Cluster Policy (e.g., `Policy1`), the `policyName` identifier remains persistently associated (orphaned) with the clusters it governed. These clusters effectively operate in an unmanaged or 'no-policy' state. If a new Virtual Cluster Policy is subsequently created using the exact same name (`Policy1`), the orphaned clusters automatically and unintentionally adopt the new policy configuration, regardless of whether the settings (e.g., changing from `Shared` to `Virtual`) are functionally different from the original.

enrichman:
This PR handles the VirtualClusterPolicy deletion.

Ref:
- #563 
- #703 

When a Policy was deleted its references were not cleared up, causing issues for the Cluster/Namespaces where it was applied to. This fixes the issue cleaning up the namespaces from the association label, and the Cluster Status object.

### DIFF
--- a/pkg/controller/cluster/pod.go
+++ b/pkg/controller/cluster/pod.go
@@ -59,8 +59,16 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	}
 
 	if !pod.DeletionTimestamp.IsZero() {
-		virtName := pod.GetAnnotations()[translate.ResourceNameAnnotation]
-		virtNamespace := pod.GetAnnotations()[translate.ResourceNamespaceAnnotation]
+		annotations := pod.GetAnnotations()
+		virtName, hasName := annotations[translate.ResourceNameAnnotation]
+		virtNamespace, hasNamespace := annotations[translate.ResourceNamespaceAnnotation]
+
+		// Some pods are owned by the cluster but don't have the annotations set (i.e. system pods)
+		// They don't exist in the virtual cluster, so we can skip them
+		if !hasName || !hasNamespace {
+			log.V(1).Info("Pod does not have virtual pod annotations, skipping deletion", "pod", pod.Name, "namespace", pod.Namespace)
+			return reconcile.Result{}, nil
+		}
 
 		virtPod := corev1.Pod{
 			ObjectMeta: metav1.ObjectMeta{
--- a/pkg/controller/cluster/pod.go
+++ b/pkg/controller/cluster/pod.go
@@ -59,16 +59,8 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	}
 
 	if !pod.DeletionTimestamp.IsZero() {
-		annotations := pod.GetAnnotations()
-		virtName, hasName := annotations[translate.ResourceNameAnnotation]
-		virtNamespace, hasNamespace := annotations[translate.ResourceNamespaceAnnotation]
-
-		// Some pods are owned by the cluster but don't have the annotations set (i.e. system pods)
-		// They don't exist in the virtual cluster, so we can skip them
-		if !hasName || !hasNamespace {
-			log.V(1).Info("Pod does not have virtual pod annotations, skipping deletion", "pod", pod.Name, "namespace", pod.Namespace)
-			return reconcile.Result{}, nil
-		}
+		virtName := pod.GetAnnotations()[translate.ResourceNameAnnotation]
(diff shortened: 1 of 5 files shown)


pkg/controller/cluster/pod.go
```diff
@@ -59,8 +59,16 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
...
 	if !pod.DeletionTimestamp.IsZero() {
-		virtName := pod.GetAnnotations()[translate.ResourceNameAnnotation]
-		virtNamespace := pod.GetAnnotations()[translate.ResourceNamespaceAnnotation]
+		annotations := pod.GetAnnotations()
+		virtName, hasName := annotations[translate.ResourceNameAnnotation]
+		virtNamespace, hasNamespace := annotations[translate.ResourceNamespaceAnnotation]
+
+		// Some pods are owned by the cluster but don't have the annotations set (i.e. system pods)
+		// They don't exist in the virtual cluster, so we can skip them
+		if !hasName || !hasNamespace {
```
galal-hussein: Do we need this extra check since we have already `newClusterPredicate()`

pkg/controller/policy/policy.go
```diff
@@ -256,6 +259,22 @@ func (c *VirtualClusterPolicyReconciler) Reconcile(ctx context.Context, req reco
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
 
+	// if DeletionTimestamp is not Zero -> finalize the object
+	if !policy.DeletionTimestamp.IsZero() {
+		return c.finalizePolicy(ctx, &policy)
+	}
+
+	// add finalizer
```
galal-hussein: nit
```suggestion
```

galal-hussein: I saw that we add similar comments throughout the code in  various places as well, we should consider cleaning up unneeded comments

pkg/controller/cluster/pod.go
enrichman: It was a tentative fix for the flaky tests, because in the logs it happened. Not sure how actually.

I'll revert this though, it's probably better to raise the err here.

pkg/controller/policy/policy_finalize.go
```diff
@@ -0,0 +1,111 @@
...
+		// Remove Pod Security Admission labels only if the policy set them
+		if policy.Spec.PodSecurityAdmissionLevel != nil {
+			delete(ns.Labels, "pod-security.kubernetes.io/enforce")
+			delete(ns.Labels, "pod-security.kubernetes.io/enforce-version")
+			delete(ns.Labels, "pod-security.kubernetes.io/warn")
+			delete(ns.Labels, "pod-security.kubernetes.io/warn-version")
+		}
+
+		if !reflect.DeepEqual(orig.Labels, ns.Labels) {
+			nsLog.V(1).Info("Updating namespace to remove policy labels")
```
JonCrowther: Could you add the namespace name to this log? That way if something goes wrong we know where it happened.

pkg/controller/policy/policy.go
```diff
@@ -256,6 +259,21 @@ func (c *VirtualClusterPolicyReconciler) Reconcile(ctx context.Context, req reco
...
+	}
+
+	if controllerutil.AddFinalizer(&policy, policyFinalizerName) {
+		log.V(1).Info("Updating VirtualClusterPolicy adding finalizer")
+
+		if err := c.Client.Update(ctx, &policy); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		return reconcile.Result{Requeue: true}, nil
```
JonCrowther: Why are we returning early and reconciling here? Shouldn't we just let it continue to reconcile the resource?

enrichman: Yes, I was a bit unsure about this. When you update the Reconcile will be triggered again anyway. I tried to look for the most idiomatic approach, but probably it's better to just continue, since we should have the latest version returned from the Update.

pkg/controller/policy/policy_finalize.go
enrichman: It has already the namespace, added few lines before:

```go
// line 74
nsLog := log.WithValues("namespace", ns.Name)
```

but I don't like the `nsName`, so I will update it with the standard `log` name we have in other places.

galal-hussein (approved):
LGTM, thanks

Issue #582, opened 2025-11-28
Integrating sysbox runc with k3k virtual mode


galal-hussein:
Sysbox is a specialized container runtime that utilizes Linux user namespace along with virtualization practices to provide system containers that runs **un-privileged** in a containerization environment.

Since sysbox-runc now [supports containerd ](https://github.com/nestybox/sysbox-runc/pull/106), it can be installed on the host cluster and used to create k3k virtual clusters in virtual mode in un-privileged state.

The following criteria:
- [ ] Creating virtual mode clusters with servers/agents in unprivileged mode.
- [ ] Running conformance test should pass successfully with sysbox runtime

aganesh-suse:
@galal-hussein If possible, kindly expand on what needs to be tested for: "Creating virtual mode clusters with servers/agents in unprivileged mode."
Is installing sysbox-runc and updating the container toml template and then trying to apply virtual cluster policy or updating the k3k cluster namespace as "unprivileged" the way to go. 
there is only privileged, baseline and restricted here though: https://github.com/rancher/k3k/blob/main/docs/virtualclusterpolicy.md#5-enforcing-pod-security-admission-podsecurityadmissionlevel

or i can create the k3k virtual cluster and see if it has a label 'unprivileged'? 

If it is a little more involved than this, kindly provide the steps to test/verify this issue. Thanks.

galal-hussein:
> [@galal-hussein](https://github.com/galal-hussein) If possible, kindly expand on what needs to be tested for: "Creating virtual mode clusters with servers/agents in unprivileged mode." Is installing sysbox-runc and updating the container toml template and then trying to apply virtual cluster policy or updating the k3k cluster namespace as "unprivileged" the way to go. there is only privileged, baseline and restricted here though: https://github.com/rancher/k3k/blob/main/docs/virtualclusterpolicy.md#5-enforcing-pod-security-admission-podsecurityadmissionlevel
> 
> or i can create the k3k virtual cluster and see if it has a label 'unprivileged'?
> 
> If it is a little more involved than this, kindly provide the steps to test/verify this issue. Thanks.

I will add  validation template shortly, however this may need extra work from my side to validate, since sysbox has not yet released 0.7 yet which has the fixes that makes sysbox integrates with k3k, I will have to compile from source the latest release and publish it in my fork to be able to validate.

galal-hussein:
## Validation Template

Context: 
- https://github.com/rancher/k3k/pull/742
[...]

PR #787, opened 2026-04-14
Add hostUsers to cluster spec

galal-hussein:
Running custom runtime classes like sysbox requires enabling linux usernamespaces, so this PR adds the hostUsers option to the cluster spec

#582


pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -216,6 +216,12 @@ type ClusterSpec struct {
 	//
 	// +optional
 	RuntimeClassName *string `json:"runtimeClassName,omitempty"`
+
+	// HostUsers enables Linux User Namespace option for the
+	// agent and server pods of the cluster in virtual or shared mode.
+	//
+	// +optional
+	HostUsers *bool `json:"hostUsers,omitempty"`
```
enrichman: Maybe the doc is a little bit unclear, because it works the other way around:

> If set to true or not present, the pod will be run in the host user namespace.

https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#alpha-level

Maybe we can write something like:

```
If set to true or not present, the pod will be run in the host user namespace.
When set to false, a new userns is created for the pod.
This field is alpha-level and is only honored by servers that enable the UserNamespacesSupport feature.
```

enrichman (commented):
Just a small comment.

enrichman (approved):

PR #788, opened 2026-04-14
Replace GoReleaser action and align other tools

enrichman:
This PR cleanup the script in #752, and align the other scripts to use the same pattern.


galal-hussein (approved):
LGTM, Thanks

Issue #796, opened 2026-04-16
workerLimit is not associated to the virtual-kubelet


jp-gouin:
The virtual-kubelet do not use the value of `workerLimit` 
https://github.com/rancher/k3k/blob/6968205016abc0cdd41bd2d16c27796a61372594/pkg/controller/cluster/agent/shared.go#L196 

The expected behaviour is that the value set by the user in the cluster definition
```
apiVersion: k3k.io/v1beta1
kind: Cluster
...
spec:
  workerLimit:
    cpu: "1"
    memory: "2Gi"
  ...
```

should be used by the virtual-kubelet

galal-hussein:
## Validation Template

Context:
- https://github.com/rancher/k3k/pull/798

## Description

The `Cluster.Spec.WorkerLimit` field is intended to cap the CPU/memory resources of this agent pod, but in shared mode the field was being ignored, `Resources.Limits` was hardcoded to an empty `ResourceList{}`, so `workerLimit` values had no effect on the running pod.

## What was fixed, or what change have occurred

1. Removed the hardcoded empty `Resources.Limits` on the shared agent container.
2. When `cluster.Spec.WorkerLimit` is set, it is now propagated to the shared agent container as `Resources.Limits`.
3. Also fixed the image registry handling for virtual mode and added unit tests for pod spec generation in both shared and virtual modes.

## Areas or cases that should be tested

### 1. Normal cluster creation (regression check)

Since we are changing the pod spec generation, validate that creating a shared cluster with no `workerLimit` still works as before and the agent pod comes up successfully with no resource limits set.

1- Install k3s host cluster
2- Install k3k controller via helm chart
3- Create a shared cluster with no `workerLimit`:

```yaml
apiVersion: k3k.io/v1beta1
kind: Cluster
metadata:
  name: sharedcluster
  namespace: test-x
spec:
  mode: shared
  persistence:
    storageRequestSize: 2G
    type: dynamic
  servers: 1
```

4- Verify the server pod and the shared agent (virtual-kubelet) pod come up successfully:

5- Inspect the shared agent pod and confirm there are no `resources.limits` set on the agent container:

```sh
kubectl get pod -n test-x k3k-sharedcluster-kubelet-<hash> -o jsonpath='{.spec.containers[0].resources}'
{}
```


### 2. Shared cluster with `workerLimit` set

1- Create a shared cluster with `workerLimit` values:

```yaml
apiVersion: k3k.io/v1beta1
kind: Cluster
[...]

PR #798, opened 2026-04-17
Add WorkerLimit to shared cluster and add unit tests

galal-hussein:
- #796 

* Fix WorkerLimit to shared agents
* Add unit tests for pod spec for both virtual and shared modes
* Fix image registry for virtual mode


enrichman (approved):
I have the feeling we could improve/cleanup a bit this part, but LGTM!

Thanks @galal-hussein

PR #814, opened 2026-04-27
Initial support for kata runtimes

bcraft:
Implements some basic initial changes required for kata containers on virtual mode.

- Update the entrypoint script to configure Cgroups correctly in the kata use case. Based on the DinD cgroup implementation already supported by kata.
- Conditionally add kmsg mount.
- Strip empty dir mounts.

### DIFF
--- a/docs/howtos/using-kata-containers.md
+++ b/docs/howtos/using-kata-containers.md
@@ -102,8 +102,6 @@ The pool does not survive a reboot. Consider wrapping `pool_create` in a systemd
 
 Place the following template at `/var/lib/rancher/k3s/agent/etc/containerd/config-v3.toml.d/custom.toml`. It registers the `kata-qemu` runtime handler and configures the devmapper snapshotter.
 
-ignore_image_defined_volumes can be set to false if using a custom image as above.
-
 ```toml
 version = 3
 
@@ -113,8 +111,8 @@ version = 3
   privileged_without_host_devices = true
   pod_annotations = ["io.katacontainers.*"]
 
-  [plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.kata-qemu.options]
-    ConfigPath = "/opt/kata/share/defaults/kata-containers/runtime-rs/configuration-qemu-runtime-rs.toml"
+[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.kata-qemu.options]
+  ConfigPath = "/opt/kata/share/defaults/kata-containers/runtime-rs/configuration-qemu-runtime-rs.toml"
 
 [plugins.'io.containerd.snapshotter.v1.devmapper']
   pool_name = "k3s"
@@ -124,9 +122,23 @@ version = 3
 ```
 ## 5. Fixing emptyDir mounts
 
-The standard K3s image declares `emptyDir` volumes in its image manifest. Kata detects these as tmpfs mounts and fails to start the container. The k3k controller automatically strips `emptyDir` volumes from the pod spec at runtime when the cluster's `runtimeClassName` starts with `kata`, but the declarations baked into the image manifest still need to be removed.
+The standard K3s image declares `emptyDir` volumes in its image manifest. Kata detects these as tmpfs mounts and fails to start the container. The k3k controller automatically strips `emptyDir` volumes from the pod spec at runtime when the cluster's `runtimeClassName` starts with `kata`, but the declarations baked into the image manifest will still be in place by default.
 
-We can avoid emptyDir mounts from image manifests by either building a custom image, or configuring containerd to ignore them. Note this is a global setting and will apply to all runtimes on a particular node.
+We can avoid emptyDir mounts from image manifests by either of the following options:
+
+### 5.1 Ignore the image manifest volume declaration (recommended)
+
+Configuring containerd to ignore the image manifest volumes, this is a simpler option but is a global setting and will apply to all runtimes on a particular node.
+
+Place the below at `/var/lib/rancher/k3s/agent/etc/containerd/config-v3.toml.d/volumes.toml`
+```toml
+version = 3
+
+[plugins.'io.containerd.cri.v1.runtime']
+  ignore_image_defined_volumes = true
+```
+
+### 5.2 Build a custom K3S image
 
 Build a custom image that omits those declarations:
 
@@ -145,13 +157,6 @@ Push this image to a registry accessible from the host cluster and configure the
 
 An alternative would be an OCI hook to strip the mounts at runtime, but that falls outside the scope of using the Kata project's supplied build artifacts.
 
-Alternatively place the below at `/var/lib/rancher/k3s/agent/etc/containerd/config-v3.toml.d/volumes.toml`
-```toml
-version = 3
-
-[plugins.'io.containerd.cri.v1.runtime']
-  ignore_image_defined_volumes = true
-```
 
 ## 6. Load Kernel Modules
 
--- a/docs/howtos/using-kata-containers.md
+++ b/docs/howtos/using-kata-containers.md
@@ -1,0 +1,214 @@
+# Using Kata Containers
+
+> **Experimental:** Kata Containers support is in early development. Expect rough edges.
+
+Kata Containers runs each pod inside a lightweight QEMU VM, providing stronger isolation than runc. When used with k3k, this means each virtual cluster's server and agent pods run as hardware-virtualized guests on the host node.
+
+This guide covers the full setup: installing Kata, configuring the QEMU runtime and devmapper snapshotter, preparing a custom K3s image, and deploying a cluster.
+
+## Requirements
+
+- KVM available on the host node (`/dev/kvm` accessible)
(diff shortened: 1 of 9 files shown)


galal-hussein (approved):
It looks pretty good, thank you!

one small note, is that we might skip checking for kata runtime, and instead has an option to mount /dev/kmsg on the host instead, or even extra mount:

```
cluster.Spec.ExtraMounts
```
but overall LGTM for me

pkg/controller/cluster/server/config.go
```diff
@@ -83,6 +83,10 @@ func serverOptions(cluster *v1beta1.Cluster, token string) string {
 	if cluster.Spec.Mode != agent.VirtualNodeMode {
 		opts = opts + "disable-agent: true\negress-selector-mode: disabled\ndisable:\n- servicelb\n- traefik\n- metrics-server\n- local-storage\n"
 	}
+
+	// log to both file and console.
+	opts = opts + "log: /var/log/k3s.log\nalsologtostderr: true\n"
+
```
enrichman: I've never realized how ugly this was. Would you mind clean it up? It looks like the perfect time to do it.

I was thinking about using an actual array of strings, and join them:

```go
var opts []string

if token != "" {
	opts = append(opts, "token: " + token)
}

if cluster.Status.ClusterCIDR != "" {
	opts = append(opts, "cluster-cidr: " + cluster.Status.ClusterCIDR)
}

// [...]

opts = append(opts, "log: /var/log/k3s.log")
opts = append(opts, "alsologtostderr: true")

return strings.Join(opts, "\n")
```

pkg/controller/cluster/server/server.go
```diff
@@ -197,6 +199,26 @@ func (s *Server) podSpec(ctx context.Context, image, name string, persistent boo
...
+				},
+			},
+		}...)
+
+		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, []corev1.VolumeMount{
+			{
+				Name:      "dev-kmsg",
+				MountPath: "/dev/kmsg",
+			},
+		}...)
```
enrichman: You don't need to allocate an array, and unwind it. You can just append the single elements:

```suggestion
		podSpec.Volumes = append(podSpec.Volumes, corev1.Volume{
			Name: "dev-kmsg",
			VolumeSource: corev1.VolumeSource{
				HostPath: &corev1.HostPathVolumeSource{
					Path: "/dev/kmsg",
				},
			},
		})

		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, corev1.VolumeMount{
			Name:      "dev-kmsg",
			MountPath: "/dev/kmsg",
		})
```

pkg/controller/cluster/server/server.go
```diff
@@ -51,6 +52,7 @@ func New(cluster *v1beta1.Cluster, client client.Client, token, image, imagePull
 		image:            image,
 		imagePullPolicy:  imagePullPolicy,
 		imagePullSecrets: imagePullSecrets,
+		isKata:           cluster.Spec.RuntimeClassName != nil && strings.HasPrefix(*cluster.Spec.RuntimeClassName, "kata"),
```
enrichman: I would remove this from here. The check can be done inside the other func
```suggestion
```

pkg/controller/cluster/server/server.go
```diff
@@ -40,6 +40,7 @@ type Server struct {
 	image            string
 	imagePullPolicy  string
 	imagePullSecrets []string
+	isKata           bool
```
enrichman: Same, we can drop this.

pkg/controller/cluster/agent/virtual.go
```diff
@@ -303,5 +309,25 @@ func (v *VirtualAgent) podSpec(ctx context.Context, image, name string) corev1.P
 
 	podSpec.HostUsers = hostUsers
 
+	if v.isKata {
```
enrichman:
```suggestion
	if cluster.Spec.RuntimeClassName != nil && strings.HasPrefix(*cluster.Spec.RuntimeClassName, "kata") {
```

But I'm not sure if it's the right way to do this. For now we should be fine.

enrichman (changes requested):
Thanks for the PR! Great work 👏 

I left some small comments.

pkg/controller/cluster/server/server.go
bcraft: Ahhh this is because previously i was adding the empty dir mounts as a list, redundant now.

pkg/controller/cluster/server/config.go
bcraft: np

galal-hussein: I am working on refactoring this part for the server/agent, by using a struct for the handpicked configs we use for components:
```yaml
// serverConfig are few options from k3s server options that will
// construct the yaml config file for k3s server
type serverConfig struct {
	clusterInit        bool     `yaml:"cluster-init,omitempty"`
	disableAgent       bool     `yaml:"disable-agent,omitempty"`
	egressSelectorMode string   `yaml:"egress-selector-mode,omitempty"`
	tlsSAN             []string `yaml:"tls-san,omitempty"`
	disable            []string `yaml:"disable,omitempty"`
	server             string   `yaml:"server,omitempty"`
	token              string   `yaml:"token,omitempty"`
	clusterCIDR        string   `yaml:"cluster-cidr,omitempty"`
	serviceCIDR        string   `yaml:"service-cidr,omitempty"`
	clusterDNS         string   `yaml:"cluster-dns,omitempty"`
	dataStoreEndpoint  string   `yaml:"datastore-endpoint,omitempty"`
	dataStoreCAFile    string   `yaml:"datastore-cafile,omitempty"`
	dataStoreCertFile  string   `yaml:"datastore-certfile,omitempty"`
	dataStoreKeyFile   string   `yaml:"datastore-keyFile,omitempty"`
}
```

so If you want to leave that as is for now, and I will refactor this part after rebasing your PR when merged

pkg/controller/cluster/server/server.go
```diff
@@ -304,6 +304,49 @@ func (s *Server) podSpec(ctx context.Context, image, name string, persistent boo
...
+			},
+		})
+
+		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, corev1.VolumeMount{
+			Name:      "dev-kmsg",
+			MountPath: "/dev/kmsg",
+		})
+
+		// Remove all EmptyDir volumes and their corresponding mounts.
+		emptyDirNames := make(map[string]bool)
```
galal-hussein: since this is repeated logic, can we add it in a common function like `FilterEmptyDirVolumes`

docs/howtos/using-kata-containers.md
```diff
@@ -0,0 +1,225 @@
...
+version = 3
+
+[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.kata-qemu]
+  runtime_type = "io.containerd.kata.v2"
+  snapshotter = "devmapper"
+  privileged_without_host_devices = true
+  pod_annotations = ["io.katacontainers.*"]
+
+  [plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.kata-qemu.options]
+    ConfigPath = "/opt/kata/share/defaults/kata-containers/runtime-rs/configuration-qemu-runtime-rs.toml"
```
enrichman: Is this space intentional?

docs/howtos/using-kata-containers.md
```diff
@@ -0,0 +1,225 @@
...
+  pool_name = "k3s"
+  root_path = "/opt/devmapper"
+  base_image_size = "4GB"
+  discard_blocks = true
+```
+## 5. Fixing emptyDir mounts
+
+The standard K3s image declares `emptyDir` volumes in its image manifest. Kata detects these as tmpfs mounts and fails to start the container. The k3k controller automatically strips `emptyDir` volumes from the pod spec at runtime when the cluster's `runtimeClassName` starts with `kata`, but the declarations baked into the image manifest still need to be removed.
+
+We can avoid emptyDir mounts from image manifests by either building a custom image, or configuring containerd to ignore them. Note this is a global setting and will apply to all runtimes on a particular node.
```
enrichman: I was thinking to swap the two, suggesting the "global" approach first, and then the custom image, for a more fine grained.

I would also paragraph them:

```
## 5. Fixing emptyDir mounts

Blabla

### 5.1 Ignore the `VOLUME` in the images

Ignore with `ignore_image_defined_volumes` at CRI level, beware its global, blabla

### 5.2 Create a custom K3s image
```

WDYT?

bcraft: done!

docs/howtos/using-kata-containers.md
bcraft: It was - v2 config required it.

But v3 supports dropins without the indentation so updated it.

enrichman (approved):
That's great, thank you @bcraft! 👏

PR #816, opened 2026-04-27
Remove the node capacity from the merged lists

galal-hussein:
The current logic in the mergResourceList, was taking all quotas applied to the namespace and the **node's** resource list, and returning back a merged list with the least restrictive, that makes sense, but does not apply after the recent logic of distributing the quotas with the capacity of the host nodes, take the following example:

3 host nodes with the current resources:
```
memory: 6Gi
cpu: 6
```
and the VCP quota applied has the following hard limits:
```
memory: 20Gi
cpu: 30
```

The current logic, would distribute the most restrictive, i.e it will attempt to distribute 6Gi to the 3 nodes of the virtual cluster, so each node would take 2G of memory and 2cpu

https://github.com/rancher/k3k/issues/683


k3k-kubelet/provider/configure_capacity.go
enrichman, line 86: I think we are not able to calculate the quotas correctly in case of error. I think we should just return after the logging, or something could fail/panic later.

enrichman (approved):

PR #827, opened 2026-04-29
Refactor servers and agents configuration management

galal-hussein:
Refactor the current handling for configuration for servers and agents of both virtual mode `k3s agent` and `k3k-kubelet` modes, the PR marshals a simple struct that contain the configuration data instead of hardcoding the configuration.


pkg/controller/cluster/server/config.go
```diff
@@ -3,6 +3,7 @@ package server
 import (
 	"fmt"
 
+	"go.yaml.in/yaml/v2"
```
enrichman: Let's use the same yaml package, without adding new modules: `gopkg.in/yaml.v3`

pkg/controller/cluster/server/config.go
```diff
@@ -13,21 +14,29 @@ import (
...
+	DisableAgent       bool     `yaml:"disable-agent,omitempty"`
+	EgressSelectorMode string   `yaml:"egress-selector-mode,omitempty"`
+	TlsSAN             []string `yaml:"tls-san,omitempty"`
+	Disable            []string `yaml:"disable,omitempty"`
+	Server             string   `yaml:"server,omitempty"`
+	Token              string   `yaml:"token,omitempty"`
+	ClusterCIDR        string   `yaml:"cluster-cidr,omitempty"`
+	ServiceCIDR        string   `yaml:"service-cidr,omitempty"`
+	ClusterDNS         string   `yaml:"cluster-dns,omitempty"`
+}
```
enrichman, line 30: NIT: I would organize the flags in a different way, maybe alphabetically, or the most "importat" as first?

enrichman (commented):
A couple of small comments, but overall LGTM, thanks!

go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	go.etcd.io/etcd/client/v3 v3.6.10
 	go.uber.org/zap v1.27.1
 	gopkg.in/yaml.v3 v3.0.1
+	gotest.tools v2.2.0+incompatible
```
enrichman: Same for the `assert`. We are using `"github.com/stretchr/testify/assert"`.

enrichman (approved):

PR #832, opened 2026-04-30
Add server and worker resource request and limits

galal-hussein:
- #796 

This PR is a follow up fix to the workerLimit bug, it adds two new fields to the cluster spec:

- spec.WorkerResources
- spec.ServerResources

Which can specify both the requests and limits for the server and worker nodes, this PR deprecates the serverLimit and workerLimit fields.


enrichman (approved):
LGTM, thanks!

PR #834, opened 2026-04-30
Filter requests. infrastructure resources for node capacity

galal-hussein:
Based on recent testing https://github.com/rancher/k3k/issues/683#issuecomment-4348654767 it seems that we are not properly handling `requests.*` and `limits.*` for the virtual node capacity in shared mode.

The `updateNodeCapacity` flow previously took every key from a namespace's merged `ResourceQuota.Status.Hard` and tried to apply it directly as virtual node capacity. That conflates request budgets with capacity ceilings: keys like `requests.cpu` and `limits.cpu` were both looked up against the host node's Allocatable (which only knows the bare cpu key), so the result depended on map iteration order and at best one of them silently won. Worse, extended-resource quotas (which Kubernetes only allows in the `requests.<name>` form) never matched the host's bare resource key and were dropped entirely.

This change adds a `filterQuotaResource` step that skips quota keys that describe request budgets rather than capacity (cpu, memory, ephemeral-storage, requests.cpu, requests.memory, requests.ephemeral-storage) and strips the limits. / requests. prefix off everything else before lookup, so `limits.cpu` distributes as `cpu` capacity and `requests.nvidia.com/gpu` distributes as `nvidia.com/gpu`.

For example, given the following quota across two virtual nodes:

```yaml
  spec:
    hard:
      limits.cpu: "8"                                                                                
      limits.memory: 16Gi
      requests.cpu: "4"                                                                              
      requests.nvidia.com/gpu: "2"
```                                                                
each virtual node now reports:
```yaml                                                                                                 
  status:         
    allocatable:                                                                                     
      cpu: "4"
      memory: 8Gi                                                                                    
      nvidia.com/gpu: "1"
```

### DIFF
--- a/k3k-kubelet/provider/configure_capacity_test.go
+++ b/k3k-kubelet/provider/configure_capacity_test.go
@@ -33,7 +33,7 @@ func Test_distributeQuotas(t *testing.T) {
 			name:            "no virtual nodes",
 			virtResourceMap: map[string]corev1.ResourceList{},
 			quotas: corev1.ResourceList{
-				corev1.ResourceCPU: resource.MustParse("2"),
+				corev1.ResourceLimitsCPU: resource.MustParse("2"),
 			},
 			want: map[string]corev1.ResourceList{},
 		},
@@ -66,8 +66,8 @@ func Test_distributeQuotas(t *testing.T) {
 				"node-4": largeAllocatable,
 			},
 			quotas: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("2"),
-				corev1.ResourceMemory: resource.MustParse("4Gi"),
+				corev1.ResourceLimitsCPU:    resource.MustParse("2"),
+				corev1.ResourceLimitsMemory: resource.MustParse("4Gi"),
 			},
 			want: map[string]corev1.ResourceList{
 				"node-1": {
@@ -91,8 +91,8 @@ func Test_distributeQuotas(t *testing.T) {
 				"node-2": largeAllocatable,
 			},
 			quotas: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("2"),
-				corev1.ResourceMemory: resource.MustParse("4Gi"),
+				corev1.ResourceLimitsCPU:    resource.MustParse("2"),
+				corev1.ResourceLimitsMemory: resource.MustParse("4Gi"),
 			},
 			want: map[string]corev1.ResourceList{
 				"node-1": {
@@ -118,7 +118,7 @@ func Test_distributeQuotas(t *testing.T) {
 				"node-3": largeAllocatable,
 			},
 			quotas: corev1.ResourceList{
-				corev1.ResourceCPU: resource.MustParse("2"),
+				corev1.ResourceLimitsCPU: resource.MustParse("2"),
 			},
 			want: map[string]corev1.ResourceList{
 				"node-1": {corev1.ResourceCPU: resource.MustParse("667m")},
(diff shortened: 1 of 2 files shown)


enrichman (changes requested):
I don't think we should do different things inside the `distributeQuotas` func.
That one is used to distribute the quotas, and it's working.

I feel this should be part of a different func that we should call before, something like:

```go
mergedQuota := mergeQuotas(resourceLists...)
mergedQuota = cleanupQuotas(mergedQuota)

// get the node's quota and merge it with the current values
m := distributeQuotas(hostResourceMap, virtResourceMap, mergedQuota)
```

and this should simplify the test, so we can test the quota that goes in, and the expected result, without bothering of the result or limit or any other prefixes.

Also instead of the long `if` I think we should use a static map, like the `milliScaleResources`:

```go
// here a nice comment about what coreResources is about (???)
var coreResources = map[corev1.ResourceName]struct{}{
	corev1.ResourceCPU:                      {},
	corev1.ResourceMemory:                   {},
	corev1.ResourceStorage:                  {},
	corev1.ResourceEphemeralStorage:         {},
	corev1.ResourceRequestsCPU:              {},
	corev1.ResourceRequestsMemory:           {},
	corev1.ResourceRequestsStorage:          {},
	corev1.ResourceRequestsEphemeralStorage: {},
}
```

Also I'm a bit worried about what is going to happen when a user use both the plain and limits/requests resource. The maps don't have a guaranteed order, so I have the feeling we could get different results. This could be easily tested though.

I am not sure if we should get the limits value, when specified. That looks the upper limit to me, so maybe we should go for it.

k3k-kubelet/provider/configure_capacity_test.go
```diff
@@ -121,9 +125,9 @@ func Test_distributeQuotas(t *testing.T) {
 				corev1.ResourceCPU: resource.MustParse("2"),
 			},
 			want: map[string]corev1.ResourceList{
-				"node-1": {corev1.ResourceCPU: resource.MustParse("667m")},
-				"node-2": {corev1.ResourceCPU: resource.MustParse("667m")},
-				"node-3": {corev1.ResourceCPU: resource.MustParse("666m")},
+				"node-1": {corev1.ResourceCPU: resource.MustParse("667m"), corev1.ResourceMemory: resource.MustParse("100Gi"), corev1.ResourcePods: resource.MustParse("1000")},
+				"node-2": {corev1.ResourceCPU: resource.MustParse("667m"), corev1.ResourceMemory: resource.MustParse("100Gi"), corev1.ResourcePods: resource.MustParse("1000")},
+				"node-3": {corev1.ResourceCPU: resource.MustParse("666m"), corev1.ResourceMemory: resource.MustParse("100Gi"), corev1.ResourcePods: resource.MustParse("1000")},
```
enrichman: Nit:

```suggestion
				"node-1": {
					corev1.ResourceCPU:    resource.MustParse("667m"),
					corev1.ResourceMemory: resource.MustParse("100Gi"),
					corev1.ResourcePods:   resource.MustParse("1000"),
				},
				"node-2": {
					corev1.ResourceCPU:    resource.MustParse("667m"),
					corev1.ResourceMemory: resource.MustParse("100Gi"),
					corev1.ResourcePods:   resource.MustParse("1000"),
				},
				"node-3": {
					corev1.ResourceCPU:    resource.MustParse("666m"),
					corev1.ResourceMemory: resource.MustParse("100Gi"),
					corev1.ResourcePods:   resource.MustParse("1000"),
				},
```

enrichman (approved):
A small NIT but LGTM! Thanks.

PR #845, opened 2026-05-10
Add optional CCNP in chart

willmurnane:
Add an option to the chart which generates an appropriate Cilium clusterwide network policy to allow the k3k-kubelet to reach the kubernetes service. This is helpful in environments which use the Cilium CNI, as that blocks traffic to the host out of the box, with hubble logs like this:
```
May 10 13:30:14.412: k3k-mycluster/k3k-mycluster-kubelet-g752n:39506 (ID:48215) <> 192.168.1.170:6443 (host) policy-verdict:none TRAFFIC_DIRECTION_UNKNOWN DENIED (TCP Flags: SYN)
May 10 13:30:14.412: k3k-mycluster/k3k-mycluster-kubelet-g752n:39506 (ID:48215) <> 192.168.1.170:6443 (host) Policy denied DROPPED (TCP Flags: SYN)
```

I validated that this works for my environment, but haven't tried it in other places yet. I'll try to remember to test in EKS sometime next week, but that's the only other kubernetes/cilium environment I generally have handy.

### DIFF
--- a/charts/k3k/templates/ccnp.yaml
+++ b/charts/k3k/templates/ccnp.yaml
@@ -72,5 +72,24 @@ spec:
 {{- end -}}
 {{- /* if $k3kService */ -}}
 {{- end -}}
+---
+apiVersion: "cilium.io/v2"
+kind: CiliumClusterwideNetworkPolicy
+metadata:
+  name: "allow-k3k-to-servers"
+spec:
+  endpointSelector:
+    matchLabels:
+      app.kubernetes.io/name: k3k
+  egress:
+    -
+      toEndpoints:
+        - matchLabels:
+            role: server
+      toPorts:
+        - ports:
+            - port: "6443"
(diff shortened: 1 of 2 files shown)


willmurnane:
Okay, now it actually works out of the box for me. (I had a flux Kustomization that was writing the correct CCNP, so the invalid one I pushed initially was being overwritten with the right thing, and my initial testing was wrong.)

willmurnane:
I added a second policy to let the `k3k-mycluster-server` talk to the k3k operator (the AWS environment is more locked down, so I didn't notice this in k3s). This improves the situation in AWS, but the `kubelet` pods report that `system:serviceaccount:k3k-mycluster:k3k-mycluster-kubelet` doesn't have permission to list nodes at cluster scope. That seems like a separate issue, which I will plan to tackle separately.

Issue #735, opened 2026-03-30
Finish removal of webhook


enrichman:
In `v1.1.0` with https://github.com/rancher/k3k/pull/699 we started removing the webhook, but we kept a no-op `RemovePodMutatingWebhook`.

In `v1.2.0` we should remove also remove the remaining reference.

galal-hussein:
I do not think this issue has to be tested, since there is no clear testing route for it. closing

PR #858, opened 2026-05-14
Remove pod mutating webhook refrences

galal-hussein:
- #735


k3k-kubelet/kubelet.go
```diff
@@ -134,10 +133,6 @@ func newKubelet(ctx context.Context, c *config) (*kubelet, error) {
 
 	logger.Info("removing pod mutating webhook")
```
enrichman: We need to remove also the log line :)

enrichman (approved):

Issue #304, opened 2025-03-12
serverArgs with datastore-endpoint


Negashev:
I tried to run k3s (in k3k) on postgres)

**Environmental Info:**
Host Cluster Version: 
rke2 version v1.31.6+rke2r1

Node(s) CPU architecture, OS, and Version: 
Ubuntu 22.04.3 LTS (GNU/Linux 5.15.0-92-generic x86_64) 5.15.0-92-generic #102-Ubuntu SMP Wed Jan 10 09:33:48 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

Host Cluster Configuration:
RKE2 cilium 1 ALL + 3 worker (2cpu 4gb 20ssd)

K3K Cluster Configuration:
```
apiVersion: k3k.io/v1alpha1
kind: Cluster
metadata:
  name: single-ssssss
  namespace: default
spec:
  agents: 1
  clusterCIDR: 10.30.0.0/16
  clusterDNS: 10.31.0.10
  mode: shared
  persistence:
    type: ephemeral
  serverArgs:
    - "--write-kubeconfig-mode=777"
    - "--datastore-endpoint=postgres://postgres:passw0rd@postgres-k3s.default.svc.cluster.local:5432/k3s?sslmode=disable"
  servers: 1
  serviceCIDR: 10.31.0.0/16
  version: v1.32.2-k3s1
```

**Describe the bug:**
master say:
time="2025-03-12T19:10:54Z" level=warning msg="Received HTTP bootstrap request from 10.42.1.35:58602, but embedded etcd is not enabled."
time="2025-03-12T19:10:54Z" level=error msg="Sending HTTP/1.1 400 response to 10.42.1.35:58602: etcd disabled"
(run without restart infinity)

agent say:
{"level":"info","timestamp":"2025-03-12T19:04:57.709Z","msg":"decoded bootstrap"}
{"level":"fatal","timestamp":"2025-03-12T19:04:57.709Z","msg":"failed to create new virtual kubelet instance","error":"data does not contain a valid RSA or ECDSA private key"}
(restart infinity)

**Steps To Reproduce:**
apply cluster.yaml 

**Expected behavior:**

**Actual behavior:**

**Additional context / logs:**

[k3k-5dc6bdc8df-8r874_k3k.log](https://github.com/user-attachments/files/19216187/k3k-5dc6bdc8df-8r874_k3k.log)

[...]

PR #859, opened 2026-05-19
Add External Datastore feature

galal-hussein:
- #304 

This PR adds support for connecting a virtual cluster to an external datastore (via kine) instead of the embedded etcd.

The PR does not add a dedicated field for external datastore, instead it leverages the secretMount feature to allow adding a drop in config for external datastore.

Changes:
   - Add `pkg/request` package for a shared HTTP client for k3s API endpoint for example to query /v1-k3s/config, /v1-k3s/bootstrap, and /v1-k3s/client-kubelet.crt
   - Switch the kubelet's client cert flow to `/v1-k3s/client-kubelet.crt` instead of minting a cert locally from the bootstrap CA private key; the virtual-cluster kubeconfig is now loaded from the host-side
  - Refactor bootstrap package rename and adding a separate function for Fetch from k3s, Save to secret, and Load from secret.
  - Adding an alternative way to `Fetch` bootstrap data from k3s other than query /v1-k3s/bootstrap to handle external datastore.
  - Skip handling etcd server pods in statefulset reconciler if external datastore is in use kubeconfig secret rather than rebuilt from bootstrap data.
  - Add SecretMount.Name to v1beta1.Cluster, with mounts falling back to SecretName when unset, so multiple SecretMounts referencing the same secret don't produce duplicate volume names.

### DIFF
--- a/k3k-kubelet/node.go
+++ b/k3k-kubelet/node.go
@@ -69,7 +69,6 @@ func loadTLSConfig(cfg config, token, agentIP, podIP string) (*tls.Config, error
 	if err := retry.OnError(controller.Backoff, func(err error) bool {
 		return err == request.ErrServerNotReady
 	}, func() error {
-
 		headers := map[string]string{
 			"k3s-Node-Name":     controller.SafeConcatName(cfg.ClusterName, "server-0"),
 			"k3s-Node-Password": token,
@@ -77,6 +76,7 @@ func loadTLSConfig(cfg config, token, agentIP, podIP string) (*tls.Config, error
 		}
 
 		tlsCrtData, err = request.RequestK3sServer(serviceName, "/v1-k3s/client-kubelet.crt", "node", token, headers)
+
 		return err
 	}); err != nil {
 		return nil, errors.New("unable to decode bootstrap: " + err.Error())
--- a/k3k-kubelet/node.go
+++ b/k3k-kubelet/node.go
@@ -2,25 +2,21 @@ package main
 
 import (
 	"crypto/tls"
-	"crypto/x509"
 	"errors"
 	"fmt"
-	"net"
 	"net/http"
 
 	"github.com/virtual-kubelet/virtual-kubelet/node/nodeutil"
 	"k8s.io/client-go/util/retry"
 
-	certutil "github.com/rancher/dynamiclistener/cert"
-
 	"github.com/rancher/k3k/pkg/controller"
 	"github.com/rancher/k3k/pkg/controller/certs"
 	"github.com/rancher/k3k/pkg/controller/cluster/server"
-	"github.com/rancher/k3k/pkg/controller/cluster/server/bootstrap"
+	"github.com/rancher/k3k/pkg/request"
 )
 
 func (k *kubelet) registerNode(agentIP, podIP string, cfg config) error {
-	tlsConfig, err := loadTLSConfig(cfg, k.name, k.token, agentIP, podIP)
+	tlsConfig, err := loadTLSConfig(cfg, k.token, agentIP, podIP)
 	if err != nil {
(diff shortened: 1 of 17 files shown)


enrichman:
This is a great feature, but I'm trying to think if we should avoid to "extend" the Cluster spec for configurations that are more K3s related.

These looks like `serverArgs` to me (https://docs.k3s.io/datastore#external-datastore-configuration-param). I think we should try to cleanup how we setup and config the K3s server, via CLI arguments, or env variables.

In this case it will be nice to have a Secret to load as a config file, and/or a ConfigMap.

Currently the configuration is not very clean. We have it hardcoded, and it can be probably improved. This looks like a good moment to try out something different.

galal-hussein:
> This is a great feature, but I'm trying to think if we should avoid to "extend" the Cluster spec for configurations that are more K3s related.
> 
> These looks like `serverArgs` to me (https://docs.k3s.io/datastore#external-datastore-configuration-param). I think we should try to cleanup how we setup and config the K3s server, via CLI arguments, or env variables.
> 
> In this case it will be nice to have a Secret to load as a config file, and/or a ConfigMap.
> 
> Currently the configuration is not very clean. We have it hardcoded, and it can be probably improved. This looks like a good moment to try out something different.

Yes I agree and its indeed an extra server args, however this specific extra arg would require extra check in the code to detect if that server arg is passed or not, but I think I can do that in the code.
 
As for the cleanup of the setup, we already store the config as secret per the last PR #827, do you mean the extra args specifically ?

k3k-kubelet/node.go
```diff
@@ -59,56 +55,41 @@ func nodeOpt(mux *http.ServeMux, tlsConfig *tls.Config, port int) nodeutil.NodeO
...
-
-	clientCert, err := tls.X509KeyPair(cert, key)
-	if err != nil {
-		return nil, errors.New("unable to get key pair: " + err.Error())
-	}
+	cert, key := certs.SplitCertKeyPEM(tlsCrtData)
 
-	// create rootCA CertPool
-	certs, err := certutil.ParseCertsPEM([]byte(b.ServerCA.Content))
+	tlsCrt, err := tls.X509KeyPair(cert, key)
```
enrichman, line 87: I haven't tried but Gemini told me the `tls.X509KeyPair` is smart enough to understand what to grab from the full PEM, so we can drop the `certs.SplitCertKeyPEM` func, and use directly the `tlsCrtData`:

```go
tlsCrt, err := tls.X509KeyPair(tlsCrtData, tlsCrtData)
```

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -239,6 +239,10 @@ type ClusterSpec struct {
 // SecretMount defines a secret to be mounted into server or agent pods,
 // allowing for custom configurations, certificates, or other sensitive data.
 type SecretMount struct {
+	// Name is the name of the volume that will be mounted to the server
+	// or agent.
```
enrichman, line 243: NIT

```suggestion
	// Name is the name of the volume that will be mounted to the server or agent.
```

pkg/controller/certs/certs.go
```diff
@@ -72,3 +74,22 @@ func AddSANs(sans []string) certutil.AltNames {
...
+
+		if strings.Contains(b.Type, "PRIVATE KEY") {
+			keyPem = append(keyPem, pem.EncodeToMemory(b)...)
+		} else {
+			certPem = append(certPem, pem.EncodeToMemory(b)...)
+		}
+	}
+
+	return certPem, keyPem
+}
```
enrichman, line 95:
```suggestion
```

galal-hussein:
Per discussion with @enrichman, I will close this PR and split the changes to two PRs to limit the scopes and unburden reviewers

Issue #346, opened 2025-05-07
Lifecycle of the pod is not reported accurately in shared mode


galal-hussein:
**Environmental Info:**
Host Cluster Version: 

v1.32.3+k3s1
Node(s) CPU architecture, OS, and Version: 

1 node ubuntu 24.04

Host Cluster Configuration:

1 server

K3K Cluster Configuration:

1 server shared mode

**Describe the bug:**

In shared mode when a pod is created and started, the k3k-kubelet doesnt really report back the correct lifecycle of the pod for example, when creating a pod:

`kubectl run nginx --image=nginx -n test`

In virtual cluster
```
# kubectl get events -n test 
LAST SEEN   TYPE     REASON                  OBJECT      MESSAGE
9s          Normal   Scheduled               pod/nginx   Successfully assigned test/nginx to ubuntu2404-0
7s          Normal   ProviderCreateSuccess   pod/nginx   Create pod in provider successfully
```
in Host cluster
```
49s         Normal    Scheduled   pod/nginx-test-sharedcluster-6e67696e782b746573742b7368617265-f3992   Successfully assigned shared/nginx-test-sharedcluster-6e67696e782b746573742b7368617265-f3992 to ubuntu2404-0
49s         Normal    Pulling     pod/nginx-test-sharedcluster-6e67696e782b746573742b7368617265-f3992   Pulling image "nginx"
47s         Normal    Pulled      pod/nginx-test-sharedcluster-6e67696e782b746573742b7368617265-f3992   Successfully pulled image "nginx" in 2.444s (2.444s including waiting). Image size: 72404038 bytes.
46s         Normal    Created     pod/nginx-test-sharedcluster-6e67696e782b746573742b7368617265-f3992   Created container: nginx
46s         Normal    Started     pod/nginx-test-sharedcluster-6e67696e782b746573742b7368617265-f3992   Started container nginx
```
This is causing some conformance tests to fail since its waiting for the event `ContainerStarted` which it cant find in the virtual cluster

```
[FAIL] [sig-node] Sysctls [LinuxOnly] [NodeConformance] [It] should support sysctls [MinimumKubeletVersion:1.21] [Environment:NotInUserNS] [Conformance] [sig-node, NodeConformance, Environment:NotInUserNS, Conformance]
  k8s.io/kubernetes/test/e2e/common/node/sysctl.go:99
```

This issue maybe related to #120

joesims22:
[...]

PR #867, opened 2026-05-22
Mirror events for the cluster into the virtual cluster.

bigkevmcd:
Adds a controller to watch for virtual cluster namespace events and mirrors them into the virtual cluster.

https://github.com/rancher/k3k/issues/346

### DIFF
--- a/k3k-kubelet/controller/syncer/events.go
+++ b/k3k-kubelet/controller/syncer/events.go
@@ -1,0 +1,123 @@
+package syncer
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/rancher/k3k/k3k-kubelet/translate"
+)
+
+const (
+	eventControllerName = "event-syncer"
+)
+
+type EventSyncer struct {
+	// virtEventRecorder is a K8s EventRecorder to emit events into the
+	// virtual cluster.
+	virtEventRecorder record.EventRecorder
+
+	// SyncerContext contains all client information for host and virtual
+	// cluster.
+	*SyncerContext
+}
+
+func (s *EventSyncer) Name() string {
+	return eventControllerName
+}
+
+// AddEventSyncer adds event syncer controller to the manager of the virtual
+// cluster.
+func AddEventSyncer(ctx context.Context, virtMgr, hostMgr manager.Manager, clusterName, clusterNamespace string, virtEventRecorder record.EventRecorder) error {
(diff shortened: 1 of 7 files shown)


k3k-kubelet/controller/syncer/events.go
```diff
@@ -0,0 +1,123 @@
...
+			ClusterName:      clusterName,
+			ClusterNamespace: clusterNamespace,
+		},
+	}
+
+	name := reconciler.Translator.TranslateName(clusterNamespace, eventControllerName)
+
+	return ctrl.NewControllerManagedBy(hostMgr).
+		Named(name).
+		For(&corev1.Event{}).
```
galal-hussein, line 58: I think we should add `WithEventFilter` here to filter events that are not related to this certain cluster

k3k-kubelet/controller/syncer/events.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	if event.InvolvedObject.Kind != "Pod" {
+		return reconcile.Result{}, nil
+	}
+
+	virtualRef := s.Translator.TranslateObjectReferenceFrom(event.InvolvedObject)
+
+	// Look up the corresponding object in the virtual cluster.
+	virtObj := &unstructured.Unstructured{}
```
galal-hussein: I am kinda confused about this part, since we already sure that involvedObject is actually a pod per line 73, why are we using dynamic lookup here with Unstructured{} object?

bigkevmcd: As I said earlier, to get the UUID of the virtual object which is different from the UUID of the host object.

bigkevmcd: Ahh...why use unstructured?...lemme change that, I added the Pod filter after implementing this.

k3k-kubelet/translate/host_test.go
```diff
@@ -0,0 +1,424 @@
...
+			clusterNamespace: "host-ns",
+			obj: &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            "my-cm",
+					Namespace:       "virt-ns",
+					ResourceVersion: "123",
+					UID:             types.UID("abc-uid"),
+				},
+			},
+			verify: func(t *testing.T, obj *corev1.ConfigMap, tr *ToHostTranslator) {
```
galal-hussein:
```suggestion
			verify: func(t *testing.T, obj *corev1.ConfigMap, _ *ToHostTranslator) {
```

k3k-kubelet/controller/syncer/events.go
bigkevmcd: But we are already only watching events in the cluster namespace?

galal-hussein: Since this lives in k3k-kubelet it will be created for each shared virtual cluster, so it should be scoped only to the cluster it operates on I think, or we can make the event syncer lives on the k3k-controller instead, but that would be anti to what we have as a pattern for the syncer

bigkevmcd: https://github.com/rancher/k3k/blob/main/k3k-kubelet/kubelet.go#L108

galal-hussein: shoot, sorry missed that part

galal-hussein (approved):
LGTM, thanks!

PR #869, opened 2026-05-25
Refactor bootstrap data management

galal-hussein:
* Reorganize the bootstrap package and add helper method to Fetch boostrap from k3s and helper methods to save/load bootstrap data to and from the bootstrap secret
* Add request package to centralize requesting from k3s API
* Use `/v1-k3s/client-kubelet.crt` to request client certificate for k3k-kubelet instead of using bootstrap secret
* Add constants for k3s server directories

### DIFF
--- a/k3k-kubelet/kubelet.go
+++ b/k3k-kubelet/kubelet.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"context"
-	"errors"
+	"fmt"
 	"path"
 	"time"
 
@@ -109,13 +109,13 @@ func newKubelet(ctx context.Context, c *config) (*kubelet, error) {
 		},
 	})
 	if err != nil {
-		return nil, errors.New("unable to create controller-runtime mgr for host cluster: " + err.Error())
+		return nil, fmt.Errorf("unable to create controller-runtime mgr for host cluster: %w", err)
 	}
 
 	// virtual client will only use core types (for now), no need to add anything other than the basics
 	virtualScheme := runtime.NewScheme()
 	if err := clientgoscheme.AddToScheme(virtualScheme); err != nil {
-		return nil, errors.New("unable to add client go types to virtual cluster scheme: " + err.Error())
+		return nil, fmt.Errorf("unable to add client go types to virtual cluster scheme: %w", err)
 	}
 
 	virtualMgr, err := ctrl.NewManager(virtConfig, manager.Options{
@@ -128,7 +128,7 @@ func newKubelet(ctx context.Context, c *config) (*kubelet, error) {
 		},
 	})
 	if err != nil {
-		return nil, errors.New("unable to create controller-runtime mgr for virtual cluster: " + err.Error())
+		return nil, fmt.Errorf("unable to create controller-runtime mgr for virtual cluster: %w", err)
 	}
 
 	controllerName := c.AgentHostname
@@ -138,12 +138,12 @@ func newKubelet(ctx context.Context, c *config) (*kubelet, error) {
 	virtEventRecorder := eb.NewRecorder(virtualScheme, corev1.EventSource{Component: path.Join(controllerName, "pod-controller")})
 
 	if err := addControllers(ctx, hostMgr, virtualMgr, c, hostClient, virtEventRecorder); err != nil {
-		return nil, errors.New("failed to add controller: " + err.Error())
+		return nil, fmt.Errorf("failed to add controller: %w", err)
 	}
 
 	clusterIP, err := clusterIP(ctx, c.ServiceName, c.ClusterNamespace, hostClient)
 	if err != nil {
-		return nil, errors.New("failed to extract the clusterIP for the server service: " + err.Error())
+		return nil, fmt.Errorf("failed to extract the clusterIP for the server service: %w", err)
 	}
 
 	// get the cluster's DNS IP to be injected to pods
@@ -151,12 +151,12 @@ func newKubelet(ctx context.Context, c *config) (*kubelet, error) {
 
 	dnsName := controller.SafeConcatNameWithPrefix(c.ClusterName, "kube-dns")
 	if err := hostClient.Get(ctx, types.NamespacedName{Name: dnsName, Namespace: c.ClusterNamespace}, &dnsService); err != nil {
-		return nil, errors.New("failed to get the DNS service for the cluster: " + err.Error())
+		return nil, fmt.Errorf("failed to get the DNS service for the cluster: %w", err)
 	}
 
 	var virtualCluster v1beta1.Cluster
 	if err := hostClient.Get(ctx, types.NamespacedName{Name: c.ClusterName, Namespace: c.ClusterNamespace}, &virtualCluster); err != nil {
-		return nil, errors.New("failed to get virtualCluster spec: " + err.Error())
+		return nil, fmt.Errorf("failed to get virtualCluster spec: %w", err)
 	}
 
 	return &kubelet{
@@ -239,7 +239,7 @@ func (k *kubelet) newProviderFunc(cfg config) nodeutil.NewProviderFunc {
 	return func(pc nodeutil.ProviderConfig) (nodeutil.Provider, node.NodeProvider, error) {
 		utilProvider, err := provider.New(*k.hostConfig, k.hostMgr, k.virtualMgr, k.logger, cfg.ClusterNamespace, cfg.ClusterName, cfg.ServerIP, k.dnsIP, cfg.AgentHostname)
 		if err != nil {
-			return nil, nil, errors.New("unable to make nodeutil provider: " + err.Error())
+			return nil, nil, fmt.Errorf("unable to make nodeutil provider: %w", err)
 		}
 
 		err = provider.ConfigureNode(
@@ -276,12 +276,12 @@ func virtRestConfig(ctx context.Context, virtualConfigPath string, hostClient ct
 	}, func() error {
 		return hostClient.Get(ctx, kubeconfigSecretName, &clusterKubeConfig)
 	}); err != nil {
-		return nil, errors.New("unable to decode bootstrap: " + err.Error())
+		return nil, fmt.Errorf("unable to decode bootstrap: %w", err)
 	}
 
 	restConfig, err := clientcmd.RESTConfigFromKubeConfig(clusterKubeConfig.Data["kubeconfig.yaml"])
 	if err != nil {
-		return nil, errors.New("failed to create config from kubeconfig file: " + err.Error())
+		return nil, fmt.Errorf("failed to create config from kubeconfig file: %w", err)
 	}
 
 	return restConfig, nil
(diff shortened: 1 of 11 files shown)


pkg/request/request.go
```diff
@@ -0,0 +1,62 @@
...
+	return base64.StdEncoding.EncodeToString([]byte(auth))
+}
+
+func RequestK3sServer(serviceIP, endpoint, user, token string, extraHeaders map[string]string) ([]byte, error) {
+	url := "https://" + serviceIP + endpoint
+
+	client := http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
```
bigkevmcd:
```suggestion
				// This is insecure because the K3s CA hasn't been setup yet.
				InsecureSkipVerify: true,
```

pkg/request/request.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+
+		return nil, err
+	}
+
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+
+	return io.ReadAll(resp.Body)
```
bigkevmcd: This code doesn't verify `resp.StatusCode`?

My reading of the K3s code is that it could return an authentication error?

pkg/controller/cluster/cluster.go
```diff
@@ -449,31 +450,16 @@ func (c *ClusterReconciler) ensureBootstrapSecret(ctx context.Context, cluster *
 	log := ctrl.LoggerFrom(ctx)
 	log.V(1).Info("Ensuring bootstrap secret")
 
-	bootstrapData, err := bootstrap.GenerateBootstrapData(ctx, cluster, serviceIP, token)
-	if err != nil {
+	if err := c.Client.Status().Update(ctx, cluster); err != nil {
```
bigkevmcd: It's not clear what's being updated here?

pkg/controller/cluster/statefulset.go
```diff
@@ -180,21 +180,14 @@ func (p *StatefulSetReconciler) getETCDTLS(ctx context.Context, cluster *v1beta1
...
-	}
-
-	endpoint := server.ServiceName(cluster.Name) + "." + cluster.Namespace
-
-	var b *bootstrap.ControlRuntimeBootstrap
+	var b *bootstrap.Data
 
 	if err := retry.OnError(k3kcontroller.Backoff, func(err error) bool {
-		return true
+		return err == request.ErrServerNotReady
```
bigkevmcd: This is calling `bootstrap.LoadFromSecret` which uses a K8s client and so won't return `ErrServerNotReady` ?

This will never retry?

pkg/controller/cluster/server/bootstrap/bootstrap.go
```diff
@@ -149,20 +100,38 @@ func decodeBootstrap(bootstrap *ControlRuntimeBootstrap) error {
...
+		}
+
+		return nil
+	})
+
+	return err
 }
 
-func GetFromSecret(ctx context.Context, client client.Client, cluster *v1beta1.Cluster) (*ControlRuntimeBootstrap, error) {
+// LoadFromSecret reads the bootstrap data of a certain cluster and returun back the decoded content
```
bigkevmcd:
```suggestion
// LoadFromSecret reads the bootstrap data of a certain cluster and returns the decoded content.
```

pkg/controller/cluster/server/bootstrap/bootstrap.go
```diff
@@ -178,9 +147,29 @@ func GetFromSecret(ctx context.Context, client client.Client, cluster *v1beta1.C
...
+	if err := json.Unmarshal(resp, &bootstrap); err != nil {
+		return nil, err
+	}
+
+	// we still need to decode each certs since the bootstrap data endpoint base64 encode each cert
+	if err := decode(&bootstrap); err != nil {
+		return nil, fmt.Errorf("failed to decode bootstrap secret: %w", err)
+	}
+
+	return &bootstrap, err
```
bigkevmcd:
```suggestion
	return &bootstrap, nil
```

pkg/request/request.go
```diff
@@ -0,0 +1,62 @@
...
+		},
+		Timeout: 5 * time.Second,
+	}
+
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Add("Authorization", "Basic "+basicAuth(user, token))
```
bigkevmcd: https://pkg.go.dev/net/http#Request.SetBasicAuth

bigkevmcd (changes requested):

enrichman:
As commented internally I think part of this refactor could be moved into a dedicated `pkg/k3s` package, with a K3s HTTP Client. :slightly_smiling_face:

pkg/k3s/client.go
```diff
@@ -0,0 +1,109 @@
...
+	}()
+
+	return io.ReadAll(resp.Body)
+}
+
+func (c *Client) SetHeaders(key, value string) {
+	if c.customHeaders != nil {
+		c.customHeaders.Set(key, value)
+	}
+}
```
enrichman: This method is currently not safe from a concurrency perspective. Even if the current usage seems low-risk.

Instead of modifying the client's state after initialization, we should make the headers immutable upon creation. Passing a long list of individual strings to a New function can become error-prone and hard to maintain, so utilizing a configuration struct is a cleaner approach. We could also store the config directly inside the client.

Something like this:
```go
type ClientConfig struct {
	NodeName string
	Token    string
	ServerIP string
	AgentIP  string
	PodIP    string
}

type Client struct {
	config        ClientConfig
	httpClient    *http.Client
	serverIP      string
	staticHeaders http.Header
}

func New(config ClientConfig) *Client {
	headers := make(http.Header)

	if config.Token != "" {
		headers.Set("k3s-Node-Password", config.Token)
	}

	if config.NodeName != "" {
		headers.Set("k3s-Node-Name", config.NodeName)
	}

	var nodeIPs []string
	if config.PodIP != "" {
		nodeIPs = append(nodeIPs, config.PodIP)
	}
	if config.AgentIP != "" {
		nodeIPs = append(nodeIPs, config.AgentIP)
	}
	
	if len(nodeIPs) > 0 {
		headers.Set("k3s-Node-IP", strings.Join(nodeIPs, ","))
	}

	return &Client{
		config:        config,
		httpClient:    &http.Client{Timeout: 30 * time.Second},
		serverIP:      config.ServerIP,
		staticHeaders: headers, 
	}
}
```

k3k-kubelet/node.go
```diff
@@ -63,56 +58,44 @@ func nodeOpt(mux *http.ServeMux, tlsConfig *tls.Config, port int) nodeutil.NodeO
...
 	if err := retry.OnError(controller.Backoff, func(err error) bool {
-		return err != nil
+		return err == k3s.ErrServerNotReady
 	}, func() error {
-		var err error
+		client := k3s.New(token, serviceName)
+
+		client.SetHeaders("k3s-Node-Name", controller.SafeConcatName(cfg.ClusterName, "server-0"))
+		client.SetHeaders("k3s-Node-Password", token)
+		client.SetHeaders("k3s-Node-IP", agentIP+","+podIP)
```
enrichman: This client can be created just once, outside of the retry.

k3k-kubelet/node.go
```diff
@@ -63,56 +58,44 @@ func nodeOpt(mux *http.ServeMux, tlsConfig *tls.Config, port int) nodeutil.NodeO
...
+
+		endpoint := "/v1-k3s/serving-kubelet.crt"
 
-		b, err = bootstrap.DecodedBootstrap(token, endpoint)
+		tlsCrtData, err = k3s.GetRaw(client, endpoint, "node")
+		if err != nil {
+			return err
+		}
```
enrichman: A dedicated client can (should?) provide a nice interface to its APIs:

```go
tlsCrtData, err = k3s.GetServingKubeletCrt(client)
if err != nil {
	return err
}
```

and hiding the internals (i.e. "which user should I use?"):

```go
// GetServingKubeletCrt returns the PEM encoded serving-kubelet.crt of the K3s server
func GetServingKubeletCrt(c *Client) ([]byte, error) {
	return c.do("/v1-k3s/serving-kubelet.crt", "node", http.MethodGet)
}
```

pkg/controller/cluster/server/server.go
```diff
@@ -154,42 +166,42 @@ func (s *Server) podSpec(ctx context.Context, image, name string, persistent boo
...
 					{
 						Name:      "varlibkubelet",
-						MountPath: "/var/lib/kubelet",
+						MountPath: kubeletDir,
 						ReadOnly:  false,
 					},
 					{
 						Name:      "varlibrancherk3s",
-						MountPath: "/var/lib/rancher/k3s",
+						MountPath: dataDir,
```
enrichman: Maybe we can rename also the `name` of the Volumes and VolumeMounts? I've seen a typical pattern is to have them dashed, so easier to read. Wdyt?

`"varlibrancherk3s"` -> `"var-lib-rancher-k3s"`

pkg/controller/cluster/server/server.go
enrichman, line 550: Should we use also here a `filepath.Join`?

pkg/controller/cluster/server/bootstrap/bootstrap.go
```diff
@@ -178,9 +147,29 @@ func GetFromSecret(ctx context.Context, client client.Client, cluster *v1beta1.C
...
+
+	bootstrap, err := k3sclient.Get[Data](client, endpoint, "server")
+	if err != nil {
+		return nil, err
+	}
+
+	// we still need to decode each certs since the bootstrap data endpoint base64 encode each cert
+	if err := decode(&bootstrap); err != nil {
+		return nil, fmt.Errorf("failed to decode bootstrap secret: %w", err)
+	}
```
enrichman: Same as the other comment. We should leave the complexity to the client, so the caller will get the expected result:

```go
bootstrap, err := k3sclient.GetServerBootstrap(client)
if err != nil {
	return nil, err
}
```
and the k3s client knows how to handle the retrieved data
```go
func GetServerBootstrap(client) (*Data, error) {
	bootstrap := &Data{}

	bootstrap, err = Get(client, "/v1-k3s/server-bootstrap", "server")
	if err != nil {
		return nil, err
	}

	if err := decode(bootstrap); err != nil {
		return nil, fmt.Errorf("failed to decode bootstrap secret: %w", err)
	}

	return bootstrap, nil
}
```

enrichman (commented):
Lefta couple of comments, but I think it's a very nice refactor overall! Can't wait to see it landing in main. 😄

enrichman (commented):
Left a couple of comments, but I think it's a very nice refactor overall! Can't wait to see it landing in main. 😄

bigkevmcd (approved):

pkg/controller/cluster/server/server.go
```diff
@@ -29,9 +30,20 @@ const (
...
+	cniDir        = "/var/lib/cni"
+	dataDir       = "/var/lib/rancher/k3s"
+	etcdDataDir   = "/var/lib/rancher/k3s/server/db/etcd"
+	initConfigDir = "/opt/rancher/k3s/init"
+	kubeletDir    = "/var/lib/kubelet"
+	logDir        = "/var/log"
+	manifestDir   = "/var/lib/rancher/k3s/server/manifests"
+	runDir        = "/run"
+	tlsDir        = "/var/lib/rancher/k3s/server/tls"
+	varRunDir     = "/var/run"
```
enrichman: Question: is there any risk of some folder to overrides the more specific one?

Just to spot this I would order them based on the folder names:

```suggestion
	initConfigDir = "/opt/rancher/k3s/init"
	configDir     = "/opt/rancher/k3s/server"
	runDir        = "/run"
	cniDir        = "/var/lib/cni"
	kubeletDir    = "/var/lib/kubelet"
	dataDir       = "/var/lib/rancher/k3s"
	etcdDataDir   = "/var/lib/rancher/k3s/server/db/etcd"
	manifestDir   = "/var/lib/rancher/k3s/server/manifests"
	tlsDir        = "/var/lib/rancher/k3s/server/tls"
	logDir        = "/var/log"
	varRunDir     = "/var/run"
```

I have not changed the name, maybe we could also use the same `k3s` prefix for k3s dirs. I.e.:

```
k3sInitConfigDir
k3sServerConfigDir

k3sDataDir
k3sManifestsDir
k3sTLSDir
```

But that's a NIT. 😄

k3k-kubelet/kubelet.go
```diff
@@ -272,76 +266,32 @@ func (k *kubelet) newProviderFunc(cfg config) nodeutil.NewProviderFunc {
...
-		[]string{user.SystemPrivilegedGroup},
-		nil, []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
-		time.Hour*24*time.Duration(356),
-		b.ClientCA.Content,
-		b.ClientCAKey.Content,
-	)
+	restConfig, err := clientcmd.RESTConfigFromKubeConfig(clusterKubeConfig.Data["kubeconfig.yaml"])
 	if err != nil {
-		return nil, err
+		return nil, errors.New("failed to create config from kubeconfig file: " + err.Error())
```
enrichman: Should we use the `fmt.Errorf` to wrap the error?

```suggestion
		return nil, fmt.Errorf("failed to create config from kubeconfig file: %w", err)
```

k3k-kubelet/node.go
```diff
@@ -63,56 +58,35 @@ func nodeOpt(mux *http.ServeMux, tlsConfig *tls.Config, port int) nodeutil.NodeO
...
-		return nil, errors.New("unable to decode bootstrap: " + err.Error())
-	}
-
-	altNames := certutil.AltNames{
-		DNSNames: []string{cfg.AgentHostname},
-		IPs: []net.IP{
-			net.ParseIP(agentIP),
-			net.ParseIP(podIP),
-		},
+		return nil, errors.New("unable to request serving kubelet certificate: " + err.Error())
```
enrichman: Same as before?

```suggestion
		return nil, fmt.Errorf("unable to request serving kubelet certificate: %w", err)
```

pkg/k3s/client.go
```diff
@@ -0,0 +1,124 @@
...
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		if errors.Is(err, syscall.ECONNREFUSED) {
+			return nil, ErrServerNotReady
+		}
+
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
```
enrichman: Sometimes this is not completely true. The request could be successful but with the server returning a different `2xx` (`201` Created, `202` Accepted or `204` No Content are pretty common for example).

I would change this to a `>= 400`, to also avoid failures because of redirects (`3xx`).

```suggestion
	if resp.StatusCode >= http.StatusBadRequest {
```

pkg/k3s/client.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, fmt.Errorf("failed to request k3s server %s", resp.Status)
```
enrichman: It's probably helpful to provide some more context:

```suggestion
		return nil, fmt.Errorf("failed executing '%s' request to k3s server: status code: %s", endpoint, resp.Status)
```

enrichman (commented):
A few very small comments, but looks so much better now, thank you!

enrichman (approved):

PR #876, opened 2026-06-03
Added HCP (Hosted Control Plane) mode

### LINKED ISSUE #884: HCP as implementation mode
adding HCP mode with K3k managing agentless K3s life cycles enables provisioning without custom tooling.

Outcomes:

CR selectable modes
Secret contract parity
Reachable via Ingress/Service
GitOps-ready materials
Hub and Spoke design

enrichman:
This PR adds a third cluster mode, **`hcp`** (Hosted Control Plane), to k3k.

In HCP mode, k3k provisions and manages an agentless K3s control plane inside the host cluster, and users bring their own external worker nodes that join via the standard K3s installer (`curl -sfL https://get.k3s.io | K3S_URL=… K3S_TOKEN=… sh -`).

  ## What's in this PR

  **API / CRDs**
  - New `Cluster.spec.mode = hcp` enum value (`pkg/apis/k3k.io/v1beta1/types.go`); CRDs regenerated.

  **Controller (`pkg/controller/cluster/`)**
  - `hcp.go` — new reconciler step that:
    - Disables the in-cluster apiserver endpoint reconciler (its default `--advertise-address:6443` is the host-cluster pod IP, unreachable from external nodes).
    - Owns `default/kubernetes` `Endpoints` + `EndpointSlice` inside the virtual cluster, pointing at the externally reachable host:port so pods on external workers can reach the apiserver ClusterIP.

  **CLI (`cli/cmds/`)**
  - `--mode hcp` accepted by `k3kcli cluster create`, with a one-line warning that the mode is experimental.
  - After the cluster is ready, both `cluster create` and `kubeconfig generate` print the two-step join instructions for HCP clusters (fetch the token, run the standard K3s installer on the worker — token
  is intentionally not embedded for security).

  **CI / tests**
  - Added `.github/workflows/test-conformance-hcp.yaml` running upstream Kubernetes conformance against an HCP cluster with two QEMU-VM workers: 100% pass (https://github.com/enrichman/k3k/actions/runs/28860632297)!

Screenshot from 2026-06-15 14-50-53

image

### DIFF
--- a/.github/workflows/test-conformance-hcp.yaml
+++ b/.github/workflows/test-conformance-hcp.yaml
@@ -1,0 +1,407 @@
+name: Conformance Tests - HCP Mode
+
+on:
+  schedule:
+    - cron: "0 1 * * *"
+  workflow_dispatch:
+    inputs:
+      k3k_version:
+        description: 'K3k version to test (e.g. v1.1.0). Leave empty to build from source.'
+        required: false
+        type: string
+      k8s_version:
+        description: 'Kubernetes version to test'
+        required: false
+        type: choice
+        options:
+          - ""
+          - "v1.34.6"
+          - "v1.35.3"
+
+permissions:
+    contents: read
+
+env:
+  K8S_VERSIONS: "v1.34.6,v1.35.3"
+  HELM_VERSION: v4.1.3
+  HELM_CHECKSUM_AMD64: 02ce9722d541238f81459938b84cf47df2fdf1187493b4bfb2346754d82a4700
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      k8s_versions: ${{ steps.set-matrix.outputs.k8s_versions }}
+    steps:
+      - id: set-matrix
+        run: |
+          if [[ -z "${{ inputs.k8s_version }}" ]]; then
+            JSON_ARRAY=$(jq -nc '"${{ env.K8S_VERSIONS }}" | split(",")')
+            echo "k8s_versions=${JSON_ARRAY}" >> "$GITHUB_OUTPUT"
+          else
+            echo "k8s_versions=[\"${{ inputs.k8s_version }}\"]" >> "$GITHUB_OUTPUT"
+          fi
+
+  conformance:
+    needs: setup
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        k8s_version: ${{ fromJSON(needs.setup.outputs.k8s_versions) }}
+
+    env:
+      KUBERNETES_VERSION: ${{ matrix.k8s_version }}
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+      with:
+        fetch-depth: 0
+        fetch-tags: true
+
+    - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
+      with:
+        go-version-file: go.mod
+
+    - name: Install helm
+      env:
+        FILENAME: helm.tar.gz
+      run: |
+        curl -sSfL -o ${{ env.FILENAME }} https://get.helm.sh/helm-${{ env.HELM_VERSION }}-linux-amd64.tar.gz
+        echo "${{ env.HELM_CHECKSUM_AMD64 }}  ${{ env.FILENAME }}" | sha256sum --check
+        tar -xvzf ${{ env.FILENAME }} linux-amd64/helm
+        sudo install -m 755 linux-amd64/helm /usr/local/bin/helm
+
+        rm -fr "${{ env.FILENAME }}" linux-amd64/helm
+
+    - name: Install hydrophone
+      run: go install sigs.k8s.io/hydrophone@3de3e886a2f6f09635d8b981c195490af1584d97 #v0.7.0
+
+    - name: Install k3s
+      env:
+        KUBECONFIG: /etc/rancher/k3s/k3s.yaml
+        K3S_HOST_VERSION: ${{ env.KUBERNETES_VERSION }}+k3s1
+      run: |
+        curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=${K3S_HOST_VERSION} INSTALL_K3S_EXEC="--write-kubeconfig-mode=777" sh -s -
+
+        kubectl cluster-info
+        kubectl get nodes
+
+        echo "KUBECONFIG=${KUBECONFIG}" >> $GITHUB_ENV
+
+    - name: Setup K3k (from source)
+      if: inputs.k3k_version == ''
+      run: |
+        export REPO=ttl.sh/$(uuidgen)
+        export VERSION=1h
+
+        make build
+        make package
+        make push
+        make install
+
+        # add k3kcli to $PATH
+        echo "${{ github.workspace }}/bin" >> $GITHUB_PATH
+
+    - name: Setup K3k (from release)
+      if: inputs.k3k_version != ''
+      run: |
+        K3K_VERSION="${{ inputs.k3k_version }}"
+        CHART_VERSION="${K3K_VERSION#v}"
+
+        helm repo add k3k https://rancher.github.io/k3k
+        helm repo update
+        helm install --namespace k3k-system --create-namespace --version "${CHART_VERSION}" k3k k3k/k3k
+        
+        wget -qO k3kcli "https://github.com/rancher/k3k/releases/download/${{ inputs.k3k_version }}/k3kcli-linux-amd64"
+        sudo mv k3kcli /usr/local/bin/k3kcli
+        sudo chmod +x /usr/local/bin/k3kcli
+
+    - name: Wait for K3k controller
+      run: |
+        echo "Wait for K3k controller deployment to be available"
+        kubectl wait -n k3k-system deployment -l "app.kubernetes.io/name=k3k" --for=condition=Available --timeout=5m
+
+    - name: Check k3kcli
+      run: k3kcli -v
+
+    - name: Create virtual cluster
+      run: |
+        kubectl create namespace k3k-mycluster
+
+        cat <<EOF | kubectl apply -f -
+        apiVersion: k3k.io/v1beta1
+        kind: Cluster
+        metadata:
+          name: mycluster
+          namespace: k3k-mycluster
+        spec:
+          mode: hcp
+          tlsSANs:
+            - "127.0.0.1"
+            - "192.168.100.1"
+          expose:
+            nodePort:
+              serverPort: 30001
+        EOF
+
+        echo "Wait for bootstrap secret to be available"
+        kubectl wait -n k3k-mycluster --for=create secret k3k-mycluster-bootstrap --timeout=5m
+
+        k3kcli kubeconfig generate --name mycluster
+
+        export KUBECONFIG=${{ github.workspace }}/k3k-mycluster-mycluster-kubeconfig.yaml
+
+        kubectl cluster-info
+        kubectl get nodes
+        kubectl get pods -A
+
+    - name: Install Virtualization Dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y qemu-kvm qemu-utils cloud-image-utils wget
+        sudo usermod -aG kvm $USER
+
+        kvm-ok
+
+    - name: Set up bridge network for VMs
+      run: |
+        # Create a bridge so both VMs share an L2 segment with unique routable IPs.
+        # Required because QEMU `-net user` gives every VM the same 10.0.2.15 NAT
+        # address, which breaks flannel VXLAN between workers.
+        sudo ip link add name k3kbr0 type bridge
+        sudo ip addr add 192.168.100.1/24 dev k3kbr0
+        sudo ip link set k3kbr0 up
+
+        # NAT outbound so VMs can reach the internet (image pulls etc).
+        sudo sysctl -w net.ipv4.ip_forward=1
+        sudo iptables -t nat -A POSTROUTING -s 192.168.100.0/24 ! -o k3kbr0 -j MASQUERADE
+        sudo iptables -A FORWARD -i k3kbr0 -j ACCEPT
+        sudo iptables -A FORWARD -o k3kbr0 -j ACCEPT
+
+        # One TAP per VM, attached to the bridge.
+        sudo ip tuntap add tap-w1 mode tap
+        sudo ip link set tap-w1 master k3kbr0
+        sudo ip link set tap-w1 up
+
+        sudo ip tuntap add tap-w2 mode tap
+        sudo ip link set tap-w2 master k3kbr0
+        sudo ip link set tap-w2 up
+
+    - name: Download Base Cloud Image
+      run: |
+        # PINNED TO UBUNTU 24.04 LTS (noble).
+        #
+        # Newer Ubuntu releases (tested: 26.04 "resolute") ship a stricter
+        # `cri-containerd.apparmor.d` profile that denies inter-thread signal
+        # delivery. The BIND ISC library used by `nslookup` relies on those
+        # signals during shutdown (`isc_app_ctxshutdown()` calls `kill()`),
+        # so when AppArmor denies them nslookup exits 139 with
+        # "kill: Permission denied". The conformance tests
+        #   `[sig-network] Services should be able to change the type from
+        #   {NodePort,ClusterIP} to ExternalName`
+        # both run `nslookup` from an exec pod and fail in that case.
+        #
+        # Before bumping past 24.04, verify those two conformance tests still
+        # pass — or that the containerd AppArmor profile on the newer release
+        # has been relaxed to allow intra-pod signals.
+        wget -q https://cloud-images.ubuntu.com/noble/current/noble-server-cloudimg-amd64.img -O ubuntu-cloudimg.img
+
+    - name: Generate SSH Key and Cloud-Init Seeds
+      run: |
+        ssh-keygen -t rsa -b 4096 -f ./id_rsa -N ""
+        PUBKEY="$(cat ./id_rsa.pub)"
+
+        # Per-VM cloud-init: each worker gets a unique hostname and a static IP
+        # on the bridge subnet via netplan.
+        for i in 1 2; do
+          IP="192.168.100.1${i}"  # 192.168.100.11 / 192.168.100.12
+          cat <<EOF > user-data-${i}
+        #cloud-config
+        hostname: worker-${i}
+        preserve_hostname: false
+        manage_etc_hosts: true
+        # Stop cloud-init from generating its own DHCP netplan that would
+        # conflict with the static one we write in write_files below.
+        network:
+          config: disabled
+        users:
+          - name: ubuntu
+            ssh_authorized_keys:
+              - ${PUBKEY}
+            sudo: ['ALL=(ALL) NOPASSWD:ALL']
+            shell: /bin/bash
+        write_files:
+          - path: /etc/netplan/50-static.yaml
+            permissions: '0600'
+            content: |
+              network:
+                version: 2
+                ethernets:
+                  ens3:
+                    dhcp4: false
+                    addresses: [${IP}/24]
+                    routes:
+                      - to: default
+                        via: 192.168.100.1
+                    nameservers:
+                      addresses: [8.8.8.8, 1.1.1.1]
+        runcmd:
+          - netplan apply
+        EOF
+          cloud-localds seed-${i}.img user-data-${i}
+        done
+
+    - name: Create Worker Disks
+      run: |
+        qemu-img create -f qcow2 -b ubuntu-cloudimg.img -F qcow2 worker-1.qcow2 20G
+        qemu-img create -f qcow2 -b ubuntu-cloudimg.img -F qcow2 worker-2.qcow2 20G
+
+    - name: Launch Worker VMs
+      run: |
+        # Launch Worker 1 — attached to tap-w1 on k3kbr0.
+        sudo qemu-system-x86_64 \
+          -m 2048 -smp 2 -cpu host -enable-kvm -nographic \
+          -drive file=worker-1.qcow2,if=virtio \
+          -drive file=seed-1.img,format=raw,if=virtio \
+          -netdev tap,id=net0,ifname=tap-w1,script=no,downscript=no \
+          -device virtio-net-pci,netdev=net0,mac=52:54:00:12:34:56 \
+          &
+
+        # Wait a moment before launching the second VM
+        sleep 5
+
+        # Launch Worker 2 — attached to tap-w2 on k3kbr0.
+        sudo qemu-system-x86_64 \
+          -m 2048 -smp 2 -cpu host -enable-kvm -nographic \
+          -drive file=worker-2.qcow2,if=virtio \
+          -drive file=seed-2.img,format=raw,if=virtio \
+          -netdev tap,id=net0,ifname=tap-w2,script=no,downscript=no \
+          -device virtio-net-pci,netdev=net0,mac=52:54:00:12:34:57 \
+          &
+
+    - name: Wait for SSH Availability
+      run: |
+        echo "Waiting for Worker 1 (192.168.100.11) to respond..."
+        timeout 180s bash -c '
+          until ssh -i ./id_rsa -o StrictHostKeyChecking=no -o ConnectTimeout=2 ubuntu@192.168.100.11 true 2>/dev/null; do sleep 3; done
+        '
+
+        echo "Waiting for Worker 2 (192.168.100.12) to respond..."
+        timeout 180s bash -c '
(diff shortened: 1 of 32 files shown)


pkg/controller/cluster/hcp.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+		portName := "https"
+
+		endpointSlice.Ports = []discoveryv1.EndpointPort{
+			{
+				Name:     &portName,
+				Port:     &port,
+				Protocol: new(corev1.ProtocolTCP),
+			},
+		}
```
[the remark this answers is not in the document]
enrichman: In Go 1.26 the `new` keyword accept constant values: https://go.dev/doc/go1.26#language

pkg/controller/cluster/server/config.go
```diff
@@ -77,10 +77,29 @@ func buildServerConfig(cluster *v1beta1.Cluster, initServer bool, serviceIP, tok
...
+		// and tears down the remotedialer session, making kubelet streams
+		// flaky. See k3s pkg/agent/tunnel/tunnel.go.
+		serverConfig.EgressSelectorMode = "cluster"
+		// Disable the apiserver's built-in endpoint reconciler so K3k can
+		// own default/kubernetes Endpoints and point it at the externally
+		// reachable host:port (NodePort / LB / Ingress).
+		serverConfig.KubeApiServerArg = append(serverConfig.KubeApiServerArg, "endpoint-reconciler-type=none")
+	case v1beta1.VirtualClusterMode:
+		// no extra config for virtual mode
 	}
```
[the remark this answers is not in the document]
enrichman: This should never happen, but let's keep it.

cli/cmds/kubeconfig.go
```diff
@@ -118,14 +116,38 @@ func generate(appCtx *AppContext, cfg *GenerateKubeconfigConfig) func(cmd *cobra
...
+		return override, nil
+	}
+
+	u, err := url.Parse(restConfigHost)
+	if err != nil {
+		return "", err
 	}
+
+	return strings.Split(u.Host, ":")[0], nil
 }
```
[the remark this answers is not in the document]
bigkevmcd: This looks right, if you really want to split the host... https://pkg.go.dev/net#SplitHostPort might be your best bet.

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -45,11 +45,11 @@ type ClusterSpec struct {
 	// +optional
 	Version string `json:"version,omitempty"`
 
-	// Mode specifies the cluster provisioning mode: "shared" or "virtual".
+	// Mode specifies the cluster provisioning mode: "shared", "virtual" or "hcp".
 	// Defaults to "shared". This field is immutable.
 	//
 	// +kubebuilder:default="shared"
-	// +kubebuilder:validation:Enum=shared;virtual
+	// +kubebuilder:validation:Enum=shared;virtual;hcp
```
bigkevmcd, line 51: I think this is wrong?

`ClusterMode` is declared with `validation:Enum` which I think is why it appears twice here?

https://github.com/rancher/k3k/pull/876/changes#diff-4587d07de4be775d28e57a4c302d2a31ea1b8fdf44c678de13ec4b1fd854c27bR53-R61

Same in the `ClusterSpec`.

pkg/controller/cluster/cluster.go
```diff
@@ -896,6 +916,13 @@ func (c *ClusterReconciler) bindClusterRoles(ctx context.Context, cluster *v1bet
 }
 
 func (c *ClusterReconciler) ensureAgent(ctx context.Context, cluster *v1beta1.Cluster, serviceIP, token string) error {
+	// hcp mode is BYO-node by design: external (out-of-host-cluster) nodes join
+	// using the standard K3s installer command surfaced via Status.HCPRegistration.
+	// k3k therefore does not provision any agent pods on the host cluster.
```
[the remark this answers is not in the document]
bigkevmcd: This looks like it should be fixed?

pkg/controller/cluster/hcp.go
```diff
@@ -0,0 +1,319 @@
...
+		if ip.IsLoopback() {
+			return corev1.EndpointAddress{}, fmt.Errorf("HCP endpoint host %q is a loopback address and cannot be used", host)
+		}
+
+		return corev1.EndpointAddress{IP: host}, nil
+	}
+
+	ips, err := net.LookupIP(host)
+	if err != nil {
+		return corev1.EndpointAddress{}, fmt.Errorf("HCP endpoint host %q is not an IP and does not resolve: %w", host, err)
```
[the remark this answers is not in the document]
bigkevmcd: This should definitely be fixed?

pkg/controller/cluster/hcp.go
```diff
@@ -0,0 +1,319 @@
...
+	if !external {
+		log.Info("HCP cluster has no externally-routable endpoint; skipping registration command",
+			"cluster", cluster.Name, "namespace", cluster.Namespace)
+
+		meta.SetStatusCondition(&cluster.Status.Conditions, metav1.Condition{
+			Type:    ConditionReady,
+			Status:  metav1.ConditionFalse,
+			Reason:  ReasonHCPNoExternalEndpoint,
+			Message: "HCP cluster has no external endpoint; set spec.expose.nodePort, spec.expose.loadBalancer or spec.expose.ingress so external nodes can reach the API server",
+		})
```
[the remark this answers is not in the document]
bigkevmcd: This will need to be fixed?

bigkevmcd (changes requested):

cli/cmds/kubeconfig.go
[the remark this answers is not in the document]
enrichman: Nice, thanks. Since we don't need the port I'll keep it like this, but I would be curious to see why the implementation is different:
- https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/net/ipsock.go;l=165
- http://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/net/url/url.go;l=1183

pkg/apis/k3k.io/v1beta1/types.go
enrichman: IIRC it was because of a strange issue, maybe https://github.com/kubernetes-sigs/controller-tools/issues/461? But probably the controller-gen version we are using is much newer, so I'm going to remove the enum falidation in the Type in order to keep the one in the field.

.github/workflows/test-conformance-hcp.yaml
```diff
@@ -0,0 +1,407 @@
...
+        kubectl get nodes
+        kubectl get pods -A
+
+    - name: Install Virtualization Dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y qemu-kvm qemu-utils cloud-image-utils wget
+        sudo usermod -aG kvm $USER
+
+        kvm-ok
```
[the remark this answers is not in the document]
enrichman: https://github.com/enrichman/k3k/actions/runs/27451128020/job/81146425437

docs/crds/crds.adoc
```diff
@@ -133,8 +133,7 @@ _Underlying type:_ _string_
 
 ClusterMode is the possible provisioning mode of a Cluster.
 
-_Validation:_
-- Enum: [shared virtual]
+
```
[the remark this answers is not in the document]
enrichman: Resolved with a comment.

docs/crds/crds.md
```diff
@@ -102,8 +102,7 @@ _Underlying type:_ _string_
 
 ClusterMode is the possible provisioning mode of a Cluster.
 
-_Validation:_
-- Enum: [shared virtual]
+
```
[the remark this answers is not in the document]
enrichman: Resolved with a comment.

pkg/controller/cluster/server/endpoint.go
```diff
@@ -0,0 +1,120 @@
...
+			ip = cluster.Status.TLSSANs[0]
+		} else {
+			logrus.Warn("IP not found in tlsSANs. This could cause issue with the certificate validation.")
+		}
+	}
+
+	url := "https://" + ip
+	if port != httpsPort {
+		url = fmt.Sprintf("%s:%d", url, port)
+	}
```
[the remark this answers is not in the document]
enrichman: Resolved with:

```go
url := "https://" + net.JoinHostPort(ip, strconv.Itoa(int(port)))
```

pkg/apis/k3k.io/v1beta1/types.go
enrichman: Resolved removing the validation from the type, adding a comment about the possible values.

cli/cmds/cluster_create.go
```diff
@@ -84,10 +83,14 @@ func createAction(appCtx *AppContext, config *CreateConfig) func(cmd *cobra.Comm
 			return errors.New("invalid cluster name")
```
bigkevmcd, line 87: Not an issue in this PR specifically...but the closure returned by `createAction` _could_ use `cmd.Context()`  instead of `context.Background()`.

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -253,10 +253,10 @@ type SecretMount struct {
 	// +optional
 	SubPath string `json:"subPath,omitempty"`
 	// Role is the type of the k3k pod that will be used to mount the secret.
-	// This can be 'server', 'agent', or 'all' (for both).
+	// This can be `server`, `agent`, or `all` (for both).
```
bigkevmcd: It's not clear why some lists of options use back-ticks and others use double-quotes?

e.g. `ClusterMode` above compared to `Role` here?

From looking around types, I would drop the quotation marks or drop the list of valid values

* https://pkg.go.dev/k8s.io/api/core/v1#PodSpec
* https://github.com/openshift/api/blob/master/operator/v1/types.go#L201
* https://github.com/kubernetes-sigs/cluster-api/blob/main/api/core/v1beta2/machinepool_types.go

pkg/controller/cluster/server/endpoint.go
```diff
@@ -0,0 +1,129 @@
...
+	if err := c.Get(ctx, key, &k3kService); err != nil {
+		return "", false, err
+	}
+
+	ip := k3kService.Spec.ClusterIP
+	port := int32(httpsPort)
+	external := false
+
+	if len(k3kService.Spec.Ports) == 0 {
+		logrus.Warn("No ports exposed by the cluster service.")
```
bigkevmcd: I think this should use the [context logger](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log#FromContext)? (There are several uses of logrus here).

pkg/controller/cluster/server/endpoint_test.go
```diff
@@ -0,0 +1,80 @@
...
+					ClusterIP: "10.43.0.50",
+					Ports:     []corev1.ServicePort{{Name: "k3s-server-port", Port: 443}},
+				},
+				Status: corev1.ServiceStatus{
+					LoadBalancer: corev1.LoadBalancerStatus{Ingress: tt.ingress},
+				},
+			}
+
+			c := fake.NewClientBuilder().WithScheme(scheme).WithObjects(svc).Build()
+			url, external, err := ServerURL(context.Background(), c, cluster, "", 0)
```
bigkevmcd:
```suggestion
			url, external, err := ServerURL(t.Context(), c, cluster, "", 0)
```

pkg/controller/cluster/hcp.go
```diff
@@ -0,0 +1,316 @@
...
+
+		return ErrHCPNoExternalEndpoint
+	}
+
+	return nil
+}
+
+// selectNonLoopbackSAN returns the first non-loopback address from the given
+// TLS SANs. Returns empty string if none is found.
+func selectNonLoopbackSAN(sans []string) string {
```
bigkevmcd: nit: I'd call this "find" rather than "select".

pkg/controller/cluster/hcp.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	if !external {
+		// Defensive: reconcile would have already short-circuited with
+		// ErrHCPNoExternalEndpoint via ensureHCPRegistration before reaching
+		// here, but skip gracefully if invoked directly.
+		return nil
+	}
+
+	host, port, err := parseHCPHostPort(rawURL)
```
bigkevmcd: It feels like `server.ServerURL` could return a [`*url.URL` ](https://pkg.go.dev/net/url#URL) to avoid having to reparse it here.

It feels odd that our internal code could return an invalid URL?

bigkevmcd (commented):
Looks good, some minor tidyups.

cli/cmds/cluster_create.go
enrichman: Agree! Thanks

pkg/apis/k3k.io/v1beta1/types.go
enrichman: Ouch.. this looks like a overlook. We should have used title cased names, to make them clearer. I will drop the quotation, but maybe we can think about moving to Titles, still supporting the old ones for backward compatibility.

enrichman: I'll drop entirely the list of allowed values. The kubebuilder annotation still make this clear, and also in the doc it will be highlighted. :+1:

pkg/controller/cluster/server/endpoint.go
enrichman: Good catch! Yes, we should use the one from the controller-runtime that has already other info in it.

pkg/controller/cluster/hcp.go
enrichman: That makes sense. Let me fix that :slightly_smiling_face:

cli/cmds/cluster_create.go
```diff
@@ -179,16 +177,40 @@ func createAction(appCtx *AppContext, config *CreateConfig) func(cmd *cobra.Comm
...
 
+func printHCPJoinInstructions(cluster *v1beta1.Cluster, kc *clientcmdapi.Config) {
+	tokenSecretName := k3kcluster.TokenSecretName(cluster.Name)
+	serverURL := kc.Clusters["default"].Server
+
+	logrus.Infof(`To join an external worker node to this HCP cluster:
+
+	1. On this machine, fetch the cluster token:
+
+		kubectl get secret -n %s %s -o jsonpath='{.data.token}' | base64 -d
```
galal-hussein, line 221:
```suggestion
		TOKEN=$(kubectl get secret -n %s %s -o jsonpath='{.data.token}' | base64 -d)
```

cli/cmds/cluster_create.go
```diff
@@ -179,16 +177,40 @@ func createAction(appCtx *AppContext, config *CreateConfig) func(cmd *cobra.Comm
...
+
+	2. On the worker node, run (replace <TOKEN> with the value from step 1):
```
galal-hussein, line 223:
```suggestion
	2. On the worker node, run:
```

cli/cmds/cluster_create.go
```diff
@@ -179,16 +177,40 @@ func createAction(appCtx *AppContext, config *CreateConfig) func(cmd *cobra.Comm
...
+
+		curl -sfL https://get.k3s.io | K3S_URL=%s K3S_TOKEN=<TOKEN> sh -
```
galal-hussein, line 225:
```suggestion
		curl -sfL https://get.k3s.io | K3S_URL=%s K3S_TOKEN=${TOKEN} sh -
```

pkg/controller/cluster/server/endpoint.go
```diff
@@ -0,0 +1,131 @@
...
+
+	ip := k3kService.Spec.ClusterIP
+	port := int32(httpsPort)
+	external := false
+
+	if len(k3kService.Spec.Ports) == 0 {
+		log.V(1).Info("No ports exposed by the cluster service.")
+	}
+
+	switch k3kService.Spec.Type {
```
galal-hussein, line 82: nit: can we switch on the cluster expose method instead, so that we can include the ingress in the same switch block?

pkg/controller/cluster/server/server.go
```diff
@@ -250,8 +249,10 @@ func (s *Server) podSpec(ctx context.Context, image, name string, persistent boo
 			},
 		},
 	}
-	// start the pod unprivileged in shared mode
-	if s.mode == agent.VirtualNodeMode {
+	// virtual mode runs an embedded kubelet inside the server pod and therefore
+	// requires Privileged. shared and hcp modes are agentless (no kubelet) and
```
galal-hussein, line 253: maybe we should remove the comment altogether since there is exceptions to this rule anyways with different runtimeclasses

pkg/controller/cluster/hcp.go
```diff
@@ -0,0 +1,315 @@
...
+// is configured, which updateStatus surfaces as Ready=False with reason
+// HCPNoExternalEndpoint.
+func (c *ClusterReconciler) ensureHCPRegistration(ctx context.Context, cluster *v1beta1.Cluster) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	host := findNonLoopbackSAN(cluster.Spec.TLSSANs)
+	if host == "" {
+		log.Info("HCP cluster is missing a non-loopback spec.tlsSANs entry for external node registration")
+
+		return ErrHCPNoExternalEndpoint
```
galal-hussein: I think we should differentiate the errors here, since there might be external endpoint but it will still give certificate error because its not configured in the SANs, also I do not think this check is 100% accurate since the user can still supply the TLS SAN in the server-args

pkg/controller/cluster/hcp.go
```diff
@@ -0,0 +1,315 @@
...
+// EndpointSlice object instead.
+func (c *ClusterReconciler) ensureHCPKubernetesEndpointSlice(ctx context.Context, cluster *v1beta1.Cluster) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	rawURL, external, err := server.ServerURL(ctx, c.Client, cluster, findNonLoopbackSAN(cluster.Spec.TLSSANs), 0)
+	if err != nil {
+		return err
+	}
+
+	if !external {
```
galal-hussein: we can drop the extra logic, since we dont really invoke this method directly

pkg/controller/cluster/hcp.go
```diff
@@ -0,0 +1,315 @@
...
+
+func (c *ClusterReconciler) ensureHCPKubernetesEndpoints(ctx context.Context, cluster *v1beta1.Cluster) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	rawURL, external, err := server.ServerURL(ctx, c.Client, cluster, findNonLoopbackSAN(cluster.Spec.TLSSANs), 0)
+	if err != nil {
+		return err
+	}
+
+	if !external {
```
galal-hussein: same as the last comment

pkg/controller/cluster/hcp.go
```diff
@@ -0,0 +1,315 @@
...
+func hcpEndpointAddress(ctx context.Context, host string) (corev1.EndpointAddress, error) {
+	if ip := net.ParseIP(host); ip != nil {
+		if ip.IsLoopback() {
+			return corev1.EndpointAddress{}, fmt.Errorf("HCP endpoint host %q is a loopback address and cannot be used", host)
+		}
+
+		return corev1.EndpointAddress{IP: host}, nil
+	}
+
+	ipAddrs, err := net.DefaultResolver.LookupIPAddr(ctx, host)
```
galal-hussein, line 203: question: will this allow airgap situations since, if the controller only has access to the host-cluster DNS?

cli/cmds/cluster_create.go
enrichman: The command is not an env var on purpose. When you run the second one on the worker node the shell is not going to have the environment set. Also I didn't want to expose in the output the actual token.

I'm resolving this, but feel free to reopen it if you want to discuss it further!

pkg/controller/cluster/hcp.go
enrichman: [7fb141e](https://github.com/rancher/k3k/pull/876/commits/7fb141ee858e87f51df6c101633b2df3ac7009c7)

This was quite "big". Can you have another look at it?

pkg/controller/cluster/server/endpoint.go
enrichman: For [Kevin's comment](https://github.com/rancher/k3k/pull/876#discussion_r3452363682) I did a quite big refactor of this part: [7fb141e](https://github.com/rancher/k3k/pull/876/commits/7fb141ee858e87f51df6c101633b2df3ac7009c7).

Imho it should be cleaner, and I've even removed one unused arg, moved a couple of things, and so on. Everything SHOULD work as before, hopefully, but please have a look!

pkg/controller/cluster/server/server.go
enrichman: Hm, not sure, or at least not currently. Even with different RuntimeClasses the Virtual Mode will still run as a privileged pod. I'm fine dropping the comment though, maybe we can somehow rephrase it.

pkg/controller/cluster/hcp.go
enrichman: I need to think a little bit more about this, I think you are right though.

pkg/controller/cluster/hcp.go
enrichman: I think so, why not? :eyes:

pkg/controller/cluster/hcp.go
enrichman: Removed error.

pkg/controller/cluster/server/server.go
galal-hussein: no its fine, this is just a nit

galal-hussein (approved):
LGTM, Thank you very much 🚀

cli/cmds/cluster_create_flags.go
```diff
@@ -43,30 +43,39 @@ func createFlags(cmd *cobra.Command, cfg *CreateConfig) {
...
+
+var validClusterModes = map[v1beta1.ClusterMode]struct{}{
+	v1beta1.VirtualClusterMode: {},
+	v1beta1.SharedClusterMode:  {},
+	v1beta1.HCPClusterMode:     {},
+}
+
 func validateCreateConfig(cfg *CreateConfig) error {
 	if cfg.servers <= 0 {
 		return errors.New("invalid number of servers")
```
bigkevmcd:
```suggestion
		return errors.New("invalid number of servers - must be 1 or more")
```

cli/cmds/policy_create.go
```diff
@@ -34,10 +34,10 @@ func NewPolicyCreateCmd(appCtx *AppContext) *cobra.Command {
 		Example: "k3kcli policy create [command options] NAME",
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			switch config.mode {
-			case string(v1beta1.VirtualClusterMode), string(v1beta1.SharedClusterMode):
+			case string(v1beta1.VirtualClusterMode), string(v1beta1.SharedClusterMode), string(v1beta1.HCPClusterMode):
 				return nil
 			default:
-				return errors.New(`mode should be one of "shared" or "virtual"`)
+				return errors.New(`mode should be one of "shared", "virtual" or "hcp"`)
```
bigkevmcd:
```suggestion
				return errors.New(`mode must be one of "shared", "virtual" or "hcp"`)
```

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -45,11 +45,10 @@ type ClusterSpec struct {
 	// +optional
 	Version string `json:"version,omitempty"`
 
-	// Mode specifies the cluster provisioning mode: "shared" or "virtual".
-	// Defaults to "shared". This field is immutable.
+	// Mode specifies the cluster provisioning mode.
```
bigkevmcd:
```suggestion
	// Mode specifies the cluster provisioning mode. This field is immutable.
```

pkg/controller/cluster/server/endpoint.go
```diff
@@ -0,0 +1,147 @@
...
+	var rawURL string
+	if port != int32(443) {
+		rawURL = fmt.Sprintf("https://%s", net.JoinHostPort(ip, strconv.Itoa(int(port))))
+	} else {
+		rawURL = fmt.Sprintf("https://%s", ip)
+	}
+
+	u, err := url.Parse(rawURL)
+
+	return u, err
```
bigkevmcd: nit: you could just return `url.Parse(rawURL)`

Or, arguably better...you could do what k8s does... https://github.com/kubernetes/kubernetes/blob/eea41af21b16fbe3b85690e867788aff211e1c22/cmd/kubeadm/app/util/endpoint.go#L143

bigkevmcd (approved):
Looks good, some minor nits but I'm happy with this.

pkg/controller/cluster/server/endpoint.go
enrichman: Yes, I kinda like the k8s approach. I renamed also the var from `ip` to `host`, because it was misleading.

PR #879, opened 2026-06-05
External datastore support

galal-hussein:
WIP
- #859 
- #304 

The PR adds support for external datastore configuration, currently there are two obstacles preventing supporting this server arg, which are:
- The code tries to get the `/v1-k3s/server-bootstrap` which is disabled in external datastore mode in k3s
- In statefulset controller, the controller tries to add `etcdpod.k3k.io/finalizer` finalizer which is not applicable in external datstore configuration

The PR solves both problems by querying the k3s server in runtime for `/v1-k3s/config` to see if `ClusterInit` flag is set to true or false, and reads the bootstrap data directly from the pod if its disabled, and skips adding the finalizer.


pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -241,6 +241,12 @@ type ClusterSpec struct {
 type SecretMount struct {
 	// Embeds SecretName, Items, DefaultMode, and Optional
 	corev1.SecretVolumeSource `json:",inline"`
+	// Name is the name of the secret mount volume that will be used
+	// as the name of volume and volume mount for the server or agent pod
+	// if empty then the secret name will be used instead.
+	//
+	// +optional
+	Name string `json:"name,omitempty"`
```
bigkevmcd, line 252: Is there any reasonable validation for this - perhaps something like?

```go
// +kubebuilder:validation:MinLength=1
// +kubebuilder:validation:MaxLength=253
// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
```

https://pkg.go.dev/k8s.io/api/core/v1#Volume

The regex is taken from IsValidDNS1123Label which is used to validate the name in K8s itself.

pkg/k3s/bootstrap_test.go
```diff
@@ -0,0 +1,156 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mux := http.NewServeMux()
+
+			mockServer := httptest.NewUnstartedServer(mux)
+			if tt.isServerRunning {
+				mockServer.StartTLS()
```
bigkevmcd, line 57: You'll need to close this?

bigkevmcd (changes requested):

pkg/apis/k3k.io/v1beta1/types.go
galal-hussein: sure, I will update the PR with the validation pattern example provided in the kubebuilder book https://book.kubebuilder.io/reference/markers/crd-validation.html for DNSSubdomain

pkg/controller/cluster/mounts/mounts.go
```diff
@@ -41,8 +41,13 @@ func buildSecretMountVolume(secretMount v1beta1.SecretMount) (corev1.Volume, cor
 		},
 	}
 
+	volName := secretMount.Name
+	if volName == "" {
+		volName = secretMount.SecretName
+	}
```
enrichman, line 47: Not a big fan, but another option to avoid the if is the [`cmp.Or`](https://pkg.go.dev/cmp#Or):


```go
volName := cmp.Or(secretMount.Name, secretMount.SecretName)
```

pkg/controller/cluster/server/bootstrap/bootstrap.go
```diff
@@ -19,17 +20,29 @@ import (
...
+
+	if config.ClusterInit {
+		log.V(1).Info("Fetching bootstrap data from K3s server API")
+		return k3s.GetServerBootstrap(client)
+	}
+
+	log.V(1).Info("Fetching bootstrap data from K3s server Pod")
+
+	return k3s.ReadBootstrapFromK3sPod(ctx, restConfig, clusterName, clusterNamespace)
 }
```
enrichman: The signature and usage of this func gives me some code smells. It has `ip` and `token` are not used directly. They are needed for the k3s client. If so it's probably better to build the `k3s` client before, and pass it to this func.

So we should do this basically (I think we can also pass the whole Cluster object entirely:

```go
k3sClient := k3s.New(k3s.ClientConfig{
	ServerIP: ip,
	Token:    token,
})

data, err := bootstrap.Fetch(ctx, k3sClient, c.RestCfg, cluster)
```

pkg/controller/cluster/cluster.go
```diff
@@ -114,6 +113,7 @@ func Add(ctx context.Context, mgr manager.Manager, config *Config, maxConcurrent
 		DiscoveryClient: discoveryClient,
 		Client:          mgr.GetClient(),
 		Scheme:          mgr.GetScheme(),
+		RestCfg:         mgr.GetConfig(),
```
enrichman, line 113: We have lots of `mgr.XXX`. The `mgr.GetConfig` is unavoidable, but we can remove the `Scheme` field, because it can be derived from the Client itself.

pkg/controller/cluster/cluster.go
```diff
@@ -462,7 +453,7 @@ func (c *ClusterReconciler) ensureBootstrapSecret(ctx context.Context, cluster *
 	log := ctrl.LoggerFrom(ctx)
 	log.V(1).Info("Ensuring bootstrap secret")
 
-	data, err := bootstrap.Fetch(ctx, serviceIP, token)
+	data, err := bootstrap.Fetch(ctx, serviceIP, token, cluster.Name, cluster.Namespace, c.RestCfg)
```
enrichman: Same as before. We should create the k3sClient and pass it.

pkg/controller/cluster/statefulset.go
```diff
@@ -106,6 +106,26 @@ func (p *StatefulSetReconciler) Reconcile(ctx context.Context, req reconcile.Req
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	k3sClient := k3s.New(k3s.ClientConfig{
+		ServerIP: fmt.Sprintf("%s.%s", server.ServiceName(cluster.Name), cluster.Namespace),
+		Token:    clusterToken,
+	})
+
+	config, err := k3s.GetServerConfig(k3sClient)
```
enrichman: I'm not sure why this are func and not methods:


```suggestion
	config, err := k3sClient.GetServerConfig()
```

pkg/controller/cluster/token.go
```diff
@@ -113,3 +114,21 @@ func TokenSecretObj(token, name, namespace string) corev1.Secret {
 func TokenSecretName(clusterName string) string {
 	return controller.SafeConcatNameWithPrefix(clusterName, "token")
 }
+
+func getClusterToken(ctx context.Context, cluster *v1beta1.Cluster, client client.Client) (string, error) {
```
enrichman: nit: I would pass the "dependency" before, then the arguments:

```suggestion
func getClusterToken(ctx context.Context, client client.Client, cluster *v1beta1.Cluster) (string, error) {
```

pkg/k3s/bootstrap.go
```diff
@@ -20,6 +36,21 @@ type cert struct {
...
+func GetServerConfig(c *Client) (*K3SConfig, error) {
+	endpoint := "/v1-k3s/config"
+
+	k3sConfig, err := do[K3SConfig](c, endpoint, "node", http.MethodGet)
+	if err != nil {
+		return nil, err
+	}
+
+	return k3sConfig, nil
+}
```
enrichman:
```suggestion
func (c *Client) GetServerConfig(*K3SConfig, error) {
	endpoint := "/v1-k3s/config"
	return do[*K3SConfig](c, endpoint, "node", http.MethodGet)
}
```

The `do` should just return the same type you ask for, otherwise it's quite strange to ask for a `K3SConfig` and get a pointer instead.

To fix it:

```go
func do[T any](c *Client, endpoint, user, method string) (T, error) {
	var resp T

	respBody, err := c.do(endpoint, user, method)
	if err != nil {
		return resp, err
	}

	// unmarshal the json data to the generic struct
	if err := json.Unmarshal(respBody, resp); err != nil {
		return resp, err
	}

	return resp, nil
}
```

pkg/k3s/client.go
```diff
@@ -71,17 +77,17 @@ func New(config ClientConfig) *Client {
 	}
 }
 
-func do[T any](c *Client, endpoint, user, method string) (T, error) {
-	var response T
+func do[T any](c *Client, endpoint, user, method string) (*T, error) {
+	response := new(T)
```
enrichman: Ah, so the previous version was correct. We don't need to change it.

pkg/apis/k3k.io/v1beta1/types.go
enrichman: @galal-hussein be careful, because the VolumeName is supposed to be a [DNS_LABEL](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names), not a [DNS_SUBDOMAIN](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names):

```go
type Volume struct {
	// name of the volume.
	// Must be a DNS_LABEL and unique within the pod.
	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
	Name [string](https://pkg.go.dev/builtin#string) `json:"name" protobuf:"bytes,1,opt,name=name"`
}
```

The differences are in the length (63 vs 253 characters) and allowed characters (labels cannot use `.`).

So the correct validation would be:

```go
// +kubebuilder:validation:MinLength=1
// +kubebuilder:validation:MaxLength=63
// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
```

enrichman (changes requested):

pkg/apis/k3k.io/v1beta1/types.go
galal-hussein: oh sorry I confused it with the secret name not volume name, I will fix

pkg/controller/cluster/mounts/mounts.go
galal-hussein: I think the rest of the code base is using the same pattern, but we should consider changing that in a separate PR perhaps

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -241,6 +241,15 @@ type ClusterSpec struct {
 type SecretMount struct {
 	// Embeds SecretName, Items, DefaultMode, and Optional
 	corev1.SecretVolumeSource `json:",inline"`
+	// Name is the name of the secret mount volume that will be used
+	// as the name of volume and volume mount for the server or agent pod
+	// if empty then the secret name will be used instead.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
```
```suggestion
	// +kubebuilder:validation:MaxLength=63
```

pkg/controller/cluster/cluster.go
```diff
@@ -462,12 +450,17 @@ func (c *ClusterReconciler) ensureBootstrapSecret(ctx context.Context, cluster *
...
+		Token:    token,
+	})
+
+	data, err := bootstrap.Fetch(ctx, k3sClient, cluster, c.RestCfg)
 	if err != nil {
 		return err
 	}
 
-	return bootstrap.SaveToSecret(ctx, c.Client, c.Scheme, cluster, data)
+	return bootstrap.SaveToSecret(ctx, c.Client, c.Client.Scheme(), cluster, data)
```
enrichman: This arg can be removed as well:

```suggestion
	return bootstrap.SaveToSecret(ctx, c.Client, cluster, data)
```

pkg/controller/cluster/cluster.go
```diff
@@ -896,7 +889,7 @@ func (c *ClusterReconciler) bindClusterRoles(ctx context.Context, cluster *v1bet
 }
 
 func (c *ClusterReconciler) ensureAgent(ctx context.Context, cluster *v1beta1.Cluster, serviceIP, token string) error {
-	config := agent.NewConfig(cluster, c.Client, c.Scheme)
+	config := agent.NewConfig(cluster, c.Client, c.Client.Scheme())
```
enrichman:
```suggestion
	config := agent.NewConfig(cluster, c.Client)
```

pkg/k3s/bootstrap.go
```diff
@@ -20,7 +36,22 @@ type cert struct {
...
+func (c *Client) GetServerConfig() (*K3SConfig, error) {
+	endpoint := "/v1-k3s/config"
+
+	k3sConfig, err := do[K3SConfig](c, endpoint, "node", http.MethodGet)
+	if err != nil {
+		return nil, err
+	}
+
+	return &k3sConfig, nil
+}
```
enrichman, line 46:
```suggestion
func (c *Client) GetServerConfig() (*K3SConfig, error) {
	endpoint := "/v1-k3s/config"
	return do[*K3SConfig](c, endpoint, "node", http.MethodGet)
}
```

pkg/k3s/bootstrap.go
enrichman:
```suggestion
bootstrap, err := do[*BootstrapData](c, endpoint, "server", http.MethodGet)
if err != nil {
return nil, err
}

// we still need to decode each certs since the bootstrap data endpoint base64 encode each cert
if err := decode(bootstrap); err != nil {
return nil, fmt.Errorf("failed to decode bootstrap secret: %w", err)
}

return bootstrap, nil
```

pkg/k3s/bootstrap.go
```diff
@@ -87,3 +118,76 @@ func decode(data *BootstrapData) error {
...
+	}
+
+	parameterCodec := runtime.NewParameterCodec(scheme)
+
+	req.VersionedParams(&corev1.PodExecOptions{
+		Command: command,
+		Stdout:  true,
+	}, parameterCodec)
+
+	exec, err := remotecommand.NewSPDYExecutor(config, "POST", req.URL())
```
enrichman: nit

```suggestion
	exec, err := remotecommand.NewSPDYExecutor(config, http.MethodPost, req.URL())
```

pkg/k3s/client_test.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func getServerAddress(serverURL string) string {
+	serverAddress := strings.Split(serverURL, "//")
+	if len(serverAddress) == 2 {
+		return serverAddress[1]
+	}
+
+	return ""
+}
```
enrichman:
```suggestion
func getServerIP(t *testing.T, serverURL string) string {
	t.Helper()

	u, err := url.Parse(serverURL)
	require.NoError(t, err)

	return u.Host
}
```

Not sure if this needs to be even in a separate func though.

pkg/k3s/bootstrap_test.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+
+			mux.Handle("/v1-k3s/config", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if _, err := w.Write([]byte(tt.serverResponse)); err != nil {
+					t.Fatalf("failed to write server response: %v", err)
+				}
+			}))
+
+			if tt.clientConfig.ServerIP == "" {
+				tt.clientConfig.ServerIP = getServerAddress(mockServer.URL)
```
enrichman:
```suggestion
				u, err := url.Parse(mockServer.URL)
				require.NoError(t, err)

				tt.clientConfig.ServerIP = u.Host
```

Something like this

enrichman (commented):
I would use assert and require, instead of the std pattern. Personally I find them more readable, and we already use them.

```
"github.com/stretchr/testify/assert"
"github.com/stretchr/testify/require"
```

Left also some other small comments. Thanks!

enrichman (approved):
Awesome, thanks! LGTM

bigkevmcd (approved):
Thanks for the fixes!

Sorry for the bad example in the comment, I should've edited it to 63.

galal-hussein:
> Thanks for the fixes!
> 
> Sorry for the bad example in the comment, I should've edited it to 63.

No worries at all, Thank you

PR #925, opened 2026-06-19
Add `--tls-sans` flag to `k3kcli`

enrichman:
## Add `--tls-sans` flag to `k3kcli cluster create`

Adds a repeatable `--tls-sans` flag to specify additional Subject Alternative Names for the virtual cluster server's TLS certificate:

```bash
k3kcli cluster create --tls-sans extra.example.com --tls-sans 192.0.2.1 mycluster
```

Previously `Cluster.Spec.TLSSANs` only contained the CLI-derived server host, so clients reaching the API server via any other name/IP hit cert validation errors. The provided SANs are now appended to the host and deduplicated. Includes updated CLI docs and a cli test.

### DIFF
--- a/cli/cmds/cluster_create.go
+++ b/cli/cmds/cluster_create.go
@@ -131,14 +131,9 @@ func createAction(appCtx *AppContext, config *CreateConfig) func(cmd *cobra.Comm
 		}
 
 		// add Host IP address as an extra TLS-SAN to expose the k3k cluster
-		url, err := url.Parse(appCtx.RestConfig.Host)
+		host, err := extractHost(appCtx.RestConfig.Host, config.kubeconfigServerHost)
 		if err != nil {
 			return err
-		}
-
-		host := url.Hostname()
-		if config.kubeconfigServerHost != "" {
-			host = config.kubeconfigServerHost
 		}
 
 		sans := append([]string{host}, config.tlsSANs...)
@@ -484,6 +479,22 @@ func getClusterDetails(cluster *v1beta1.Cluster) (string, error) {
 	}
 
 	return buf.String(), nil
+}
+
+// extractHost returns the hostname from the given server URL, stripping any
+// port and correctly handling IPv6 addresses (e.g. "[2001:db8::1]:6443").
+// If override is non-empty it is returned as-is.
+func extractHost(serverURL, override string) (string, error) {
+	if override != "" {
+		return override, nil
+	}
+
+	u, err := url.Parse(serverURL)
+	if err != nil {
+		return "", err
+	}
+
+	return u.Hostname(), nil
 }
 
 // uniqueStrings returns a new slice with duplicates removed, preserving the first occurrence order.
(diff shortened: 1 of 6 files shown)


cli/cmds/cluster_create.go
```diff
@@ -138,7 +139,7 @@ func createAction(appCtx *AppContext, config *CreateConfig) func(cmd *cobra.Comm
 			host = []string{config.kubeconfigServerHost}
 		}
 
-		cluster.Spec.TLSSANs = []string{host[0]}
+		cluster.Spec.TLSSANs = uniqueStrings(append([]string{host[0]}, config.tlsSANs...))
```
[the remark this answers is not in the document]
andreas-kupries: @enrichman This seems to be a correct observation, and it does not look to be fixed yet ?!

Better use the https://pkg.go.dev/net/url#URL.Hostname method.

cli/cmds/cluster_create.go
andreas-kupries:
```suggestion
		host := url.Hostname()
```

cli/cmds/cluster_create.go
```diff
@@ -481,3 +482,18 @@ func getClusterDetails(cluster *v1beta1.Cluster) (string, error) {
 
 	return buf.String(), nil
 }
+
+// uniqueStrings returns a new slice with duplicates removed, preserving the first occurrence order.
```
andreas-kupries, line 500: Are there go packages for generic slice deduplication (and preserving order) ?

tests/cli/cli_test.go
```diff
@@ -113,6 +113,36 @@ var _ = When("using the k3kcli", Label("cli"), func() {
...
+			clusterNamespace := namespace.Name
+
+			DeferCleanup(func() {
+				fwk3k.DeleteNamespaces(k8s, clusterNamespace)
+			})
+
+			_, stderr, err = K3kcli("cluster", "create",
+				"--namespace", clusterNamespace,
+				"--tls-sans", "extra.example.com",
+				"--tls-sans", "192.0.2.1",
```
andreas-kupries, line 134: Add hostnames with ports, and ipv6 hostnames, without and without ports, to test hostname extraction.

andreas-kupries (changes requested):

cli/cmds/cluster_create.go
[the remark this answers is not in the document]
enrichman: Yep, addressed adding a separate func, so it's easier to test this. 👍

tests/cli/cli_test.go
enrichman: Done!

cli/cmds/cluster_create.go
enrichman: Unfortunately not. Asking also Gemini there is a sets in the k8s packages, but it's not going to preserve the original order.

andreas-kupries (approved):

PR #931, opened 2026-06-22
Update advanced-usage.md

ppzhukov:
The documentation appears to contain an incorrect value for the nginx.ingress.kubernetes.io/backend-protocol annotation.


galal-hussein (approved):

galal-hussein:
This needs to be fixed also in
 https://github.com/rancher/k3k-product-docs/blob/main/versions/v1.1.0/modules/en/pages/advanced-usage.adoc
and other versions as well

PR #938, opened 2026-06-24
Refactor kubeconfig URL generation

enrichman:
This PR refactors the kubeconfig generation workflow returning now a `*url.URL` instead of a raw string.
The behavior was slightly changed to address and fix some quirks or issues that probably were never faced yet.

The `serverPort` parameter is no longer supported (it was not used). Now Port is always read from the service definition.

The kubeconfig is also now retrieved from the secret where is stored, removing the legacy behavior creating a new one from the bootstrap secret.

Note: this refactor was made to address https://github.com/rancher/k3k/pull/876#discussion_r3452363682, adding before the tests with https://github.com/rancher/k3k/pull/939 to avoid any regression.


cli/cmds/cluster_create.go
```diff
@@ -174,12 +175,21 @@ func createAction(appCtx *AppContext, config *CreateConfig) func(cmd *cobra.Comm
 			Steps:    25,
 		}
 
-		cfg := kubeconfig.New()
-
 		var kubeconfig *clientcmdapi.Config
 
 		if err := retry.OnError(availableBackoff, apierrors.IsNotFound, func() error {
-			kubeconfig, err = cfg.Generate(ctx, client, cluster, host[0], 0)
+			kubeconfigSecretKey := types.NamespacedName{
```
JonCrowther, line 181: When a cluster is created, is a kubeconfig generated somewhere else? I'm just wondering if we should have a flow like
```
getKubeconfigSecret()
if err == NotFound {
  createKubeconfig()
}
```

enrichman: No, because the kubeconfig is created from the controller when the cluster is ready. If the secret is not there, it means the Cluster creation failed, or something else, but still nothing the cli can/should fix.

bigkevmcd (approved):

galal-hussein (approved):

Issue #872, opened 2026-05-27
Support custom annotation for the expose service


jp-gouin:
User should be able to set custom annotations for the service set in the `expose` settings. 
This is useful for the expose `loadbalancer` settings to specify custom behaviour for the load balancer (setting IPs etc)

fillipehmeireles:
# QA Validation

## Environment

- k3k Helm chart: `1.2.0-rc3`
- Host cluster: k3d (`k3k-872-host`)
- Kubernetes version: `v1.35.5+k3s1`
- Host nodes: 1 server and 1 agent
- Cluster mode: Shared
- Expose type: LoadBalancer

## Validation

| # | Test | Expected result | Status |
|---:|---|---|:---:|
| 1 | CRD field support | `spec.expose.annotations` is available in the installed CRD. | PASS ✅ |
| 2 | Primary annotation propagation | `qa.k3k.io/test=original-value` is added to the generated Service. | PASS ✅ |
| 3 | Multiple annotations | `qa.k3k.io/second=second-value` is also added to the generated Service. | PASS ✅ |
| 4 | Service type | The annotated Service remains a `LoadBalancer`. | PASS ✅ |
| 5 | Annotation update | Updating the Cluster changes the Service annotation to `updated-value`. | PASS ✅ |
| 6 | Reconciliation | A subsequent Cluster update changes the annotation to `updated-again`. | PASS ✅ |
| 7 | Existing annotation preservation | `external.qa.k3k.io/retain-me=yes` remains after reconciliation. | PASS ✅ |
| 8 | Regression without annotations | A Cluster without custom annotations creates its LoadBalancer Service normally. | PASS ✅ |

Final annotations on the generated Service:

```yaml
external.qa.k3k.io/retain-me: "yes"
qa.k3k.io/second: "second-value"
qa.k3k.io/test: "updated-again"
```
## Final Result

**PASS ✅**

PR #950, opened 2026-06-26
Support annotations on exposed Services

bigkevmcd:
This adds a new .spec.expose.annotations field which is applied to the Service when it's created.

Addresses #872


pkg/controller/cluster/server/service_test.go
```diff
@@ -0,0 +1,208 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/rancher/k3k/pkg/apis/k3k.io/v1beta1"
+)
+
+func TestNoExposeService(t *testing.T) {
```
galal-hussein: I think all of the 4 test functions can be merged only into one function with multiple cases since we are basically testing the same thing:
```go
service := Service(cluster)
assert.Equal(t, want, service)
```

Also we can have some helper to generate the cluster spec with specific expose option to avoid repeating code

galal-hussein (approved):
Only one comment, but LGTM for me

galal-hussein (approved):

PR #989, opened 2026-07-06
Isolate only virtual cluster workload Pods

### LINKED ISSUE #443: Shared mode - Fix network policy and Cilium
**Environmental Info:**
Host Cluster Version: 
RKE2 v1.31.5+rke2r1 
CNI: Cilium

Host Cluster Configuration:
1 servers, 1 agents
CNI Cilium

K3K Cluster Configuration:
`shared` mode virtual cluster
1 server

**Describe the bug:**
The shared kubelet cannot connect with the host kubeapi server
```
~> k logs -n mdk3k-ns2 k3k-mdclustersh1-kubelet-ksjlz
W0714 23:02:34.921323       1 client_config.go:659] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
{"level":"fatal","timestamp":"2025-07-14T23:03:04.923Z","msg":"failed to create new virtual kubelet instance","error":"failed to get API group resources: unable to retrieve the complete list of server APIs: k3k.io/v1alpha1: Get \"https://10.43.0.1:443/apis/k3k.io/v1alpha1\": dial tcp 10.43.0.1:443: i/o timeout"}
```

When disabling the `network policy` using a `VirtualClusterPolicy`, the K3k cluster is up and running

```
apiVersion: k3k.io/v1alpha1
kind: VirtualClusterPolicy
metadata:
  name: policy-example-cilium
spec:
  disableNetworkPolicy: true
```

**Expected behavior:**
The virtual cluster should be up with the embedded network policy

Relates to https://github.com/rancher/k3k/issues/405 

enrichman:
## Summary

Shared-mode virtual clusters are isolated from one another by a per-namespace `NetworkPolicy`.

This PR enhances and fixes this isolation in two ways:

1. Scope the policy to synced workload pods via a `podSelector` on the `k3k.io/clusterName` label.
2. Derive the excluded pod CIDR from the real host network instead of a single hardcoded/guessed value, so cross-cluster pod-to-pod traffic is actually blocked.

## Background: how the isolation actually works

In shared mode all virtual clusters share a flat host pod network, and each cluster's pods are scattered across every node. Two pods from *different* virtual clusters can land on the same node in the same `/24`, so there is no CIDR that separates clusters. Isolation is therefore enforced by **namespace**, not by address range.

The egress rules are _additive-allow_ on top of _default-deny_:

```
1. to ipBlock 0.0.0.0/0  except <all host pod CIDRs>   # only non-pod / external traffic
2. to namespaceSelector: <own namespace>               # re-admit own virtual cluster
3. to namespaceSelector: kube-system / kube-dns        # DNS
```

Rule 1's `except` removes the **entire** pod network from the `0.0.0.0/0` allow, so it permits no pod-to-pod traffic at all. Every pod-to-pod connection must instead match rule 2, the per-namespace selector, which only re-admits pods in the cluster's own namespace. Cross-cluster pods match neither rule 1 (excluded) nor rule 2 (different namespace) and are dropped.

## The bugs this fixes

**1. The `except` block was a guess, not the real pod network.**

The controller path excluded a single value (`cluster.Status.ClusterCIDR`) and the policy path had an ad-hoc node loop that didn't guard against an empty `PodCIDR`. When the value didn't match the host's real pod CIDRs, rule 1's `0.0.0.0/0` allow *matched other clusters' pod IPs*, so isolation silently leaked (cross-cluster pods were reachable).

**2. The policy relied on brittle assumptions about which pods it selected.**

Adding an explicit `podSelector` on `k3k.io/clusterName` makes the policy target exactly the synced workload pods, which is the intended and self-documenting behavior.

## Changes

- **`pkg/controller/policy/networkpolicy.go`**
  - New exported `FindPodCIDRs(ctx, client, clusterCIDR)`: returns `--cluster-cidr` if set, otherwise infers from Nodes' `Spec.PodCIDRs`/`Spec.PodCIDR`, logging a warning for the empty case.
  - Added the `podSelector` on `k3k.io/clusterName` to the isolation policy.
- **`pkg/controller/cluster/cluster.go`**
  - `ensureNetworkPolicy` now uses `FindPodCIDRs` for the `except` list (was a single guessed value) and applies the same `podSelector`.

## Notes on pod-CIDR detection

`node.Spec.PodCIDR` is only populated when kube-controller-manager runs with `--allocate-node-cidrs`.

**K3s and RKE2 enable it by default** (cluster-cidr `10.42.0.0/16`), so the node-inference path works out of the box there, regardless of CNI. Some setups don't set it, i.e. Cilium with cluster-pool IPAM (Cilium's default), which stores per-node CIDRs in the `v2.CiliumNode` CRD and doesn't need Kubernetes to hand out PodCIDRs. In that case set the controller's `--cluster-cidr` flag; the controller logs a warning if it can't determine the CIDR otherwise.

See: https://docs.cilium.io/en/stable/network/concepts/ipam/cluster-pool/

### Reference

- #443

### DIFF
--- a/pkg/controller/policy/networkpolicy.go
+++ b/pkg/controller/policy/networkpolicy.go
@@ -11,6 +11,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 
+	"github.com/rancher/k3k/k3k-kubelet/translate"
 	"github.com/rancher/k3k/pkg/apis/k3k.io/v1beta1"
 	k3kcontroller "github.com/rancher/k3k/pkg/controller"
 )
@@ -32,9 +33,17 @@ func (c *VirtualClusterPolicyReconciler) reconcileNetworkPolicy(ctx context.Cont
 		for _, node := range nodeList.Items {
 			if len(node.Spec.PodCIDRs) > 0 {
 				cidrList = append(cidrList, node.Spec.PodCIDRs...)
-			} else {
+			} else if node.Spec.PodCIDR != "" {
 				cidrList = append(cidrList, node.Spec.PodCIDR)
 			}
+		}
+
+		// Some CNIs (e.g. Cilium with cluster-pool IPAM) don't populate node.Spec.PodCIDR.
+		// Without it the egress rule can't exclude the pod network, so cross-cluster pod
+		// isolation would silently not be enforced. Set --cluster-cidr in that case.
+		if len(cidrList) == 0 {
+			log.Info("Could not determine the pod CIDR from the nodes; cross-cluster pod isolation " +
+				"will not be enforced. Set the --cluster-cidr flag on the k3k controller to fix this.")
 		}
 	}
 
@@ -79,6 +88,15 @@ func networkPolicy(namespaceName string, policy *v1beta1.VirtualClusterPolicy, c
 			},
 		},
 		Spec: networkingv1.NetworkPolicySpec{
+			// Isolate synced workload pods
+			PodSelector: metav1.LabelSelector{
+				MatchExpressions: []metav1.LabelSelectorRequirement{
+					{
+						Key:      translate.ClusterNameLabel,
+						Operator: metav1.LabelSelectorOpExists,
+					},
+				},
+			},
 			PolicyTypes: []networkingv1.PolicyType{
 				networkingv1.PolicyTypeIngress,
 				networkingv1.PolicyTypeEgress,
--- a/pkg/controller/policy/networkpolicy.go
+++ b/pkg/controller/policy/networkpolicy.go
@@ -2,6 +2,7 @@ package policy
 
 import (
 	"context"
+	"sort"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -77,6 +78,8 @@ func FindPodCIDRs(ctx context.Context, cl client.Client, clusterCIDR string) ([]
 			cidrList = append(cidrList, node.Spec.PodCIDR)
 		}
 	}
+
+	sort.Strings(cidrList)
 
 	// node.Spec.PodCIDR is only populated when kube-controller-manager runs with --allocate-node-cidrs.
 	// K3s and RKE2 enable it by default (cluster-cidr 10.42.0.0/16),
--- a/pkg/controller/policy/networkpolicy.go
+++ b/pkg/controller/policy/networkpolicy.go
@@ -2,8 +2,8 @@ package policy
 
 import (
 	"context"
-	"sort"
 
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	corev1 "k8s.io/api/core/v1"
@@ -69,17 +69,17 @@ func FindPodCIDRs(ctx context.Context, cl client.Client, clusterCIDR string) ([]
 		return nil, err
 	}
 
-	var cidrList []string
+	cidrs := sets.New[string]()
 
 	for _, node := range nodeList.Items {
-		if len(node.Spec.PodCIDRs) > 0 {
-			cidrList = append(cidrList, node.Spec.PodCIDRs...)
-		} else if node.Spec.PodCIDR != "" {
-			cidrList = append(cidrList, node.Spec.PodCIDR)
+		cidrs.Insert(node.Spec.PodCIDRs...)
+
+		if node.Spec.PodCIDR != "" {
+			cidrs.Insert(node.Spec.PodCIDR)
 		}
 	}
 
-	sort.Strings(cidrList)
+	cidrList := sets.List(cidrs)
 
 	// node.Spec.PodCIDR is only populated when kube-controller-manager runs with --allocate-node-cidrs.
 	// K3s and RKE2 enable it by default (cluster-cidr 10.42.0.0/16),
--- a/pkg/controller/policy/networkpolicy.go
+++ b/pkg/controller/policy/networkpolicy.go
@@ -51,11 +51,11 @@ func (c *VirtualClusterPolicyReconciler) reconcileNetworkPolicy(ctx context.Cont
 	return err
 }
 
-// FindPodCIDRs returns the CIDR ranges to exclude from the isolation NetworkPolicy's egress
-// allow-list, so that pod-to-pod traffic on the host's real pod network is blocked. If clusterCIDR
-// is set (the --cluster-cidr controller flag) it's used as-is; otherwise it's inferred from the
-// host Nodes' Spec.PodCIDR(s). Returns an empty list, and logs a warning, if it can't be
-// determined either way.
+// FindPodCIDRs returns the CIDR ranges to exclude from the isolation NetworkPolicy's egress allow-list,
+// so that pod-to-pod traffic on the host's real pod network is blocked.
+// If clusterCIDR is set (the --cluster-cidr controller flag) it's used as-is;
+// otherwise it's inferred from the host Nodes' Spec.PodCIDR(s).
+// Returns an empty list, and logs a warning, if it can't be determined either way.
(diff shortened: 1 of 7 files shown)


pkg/controller/policy/networkpolicy_test.go
```diff
@@ -0,0 +1,115 @@
...
+	// a Node with a different PodCIDR is present, but the explicit override must win and
+	// the Node must not even be consulted
+	node := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{Name: "node-1"},
+		Spec:       corev1.NodeSpec{PodCIDR: "10.244.0.0/24"},
+	}
+
+	cl := createFakeClient(t, node)
+
+	cidrs, err := policy.FindPodCIDRs(context.Background(), cl, "10.42.0.0/16")
```
bigkevmcd:
```suggestion
	cidrs, err := policy.FindPodCIDRs(t.Context(), cl, "10.42.0.0/16")
```

pkg/controller/policy/networkpolicy.go
```diff
@@ -64,6 +52,51 @@ func (c *VirtualClusterPolicyReconciler) reconcileNetworkPolicy(ctx context.Cont
...
+		return nil, err
+	}
+
+	var cidrList []string
+
+	for _, node := range nodeList.Items {
+		if len(node.Spec.PodCIDRs) > 0 {
+			cidrList = append(cidrList, node.Spec.PodCIDRs...)
+		} else if node.Spec.PodCIDR != "" {
+			cidrList = append(cidrList, node.Spec.PodCIDR)
```
bigkevmcd: nit: You could possibly simplify this to append the `node.Spec.PodCIDRs` always and if `node.Spec.PodCIDR != ""` append it?

If `node.Spec.PodCIDRs` is empty nothing would be appended...

```go
for _, node := range nodeList.Items {
        cidrList = append(cidrList, node.Spec.PodCIDRs...)

        if node.Spec.PodCIDR != "" {
			cidrList = append(cidrList, node.Spec.PodCIDR)
        }
}
```

pkg/controller/policy/networkpolicy_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+func TestFindPodCIDRs_FromNodePodCIDR(t *testing.T) {
+	node := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{Name: "node-1"},
+		Spec:       corev1.NodeSpec{PodCIDR: "192.168.77.0/24"},
+	}
+
+	cl := createFakeClient(t, node)
+
+	cidrs, err := policy.FindPodCIDRs(context.Background(), cl, "")
```
bigkevmcd:
```suggestion
	cidrs, err := policy.FindPodCIDRs(t.Context(), cl, "")
```

pkg/controller/policy/networkpolicy_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+func TestFindPodCIDRs_FromNodePodCIDRs(t *testing.T) {
+	node := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{Name: "node-1"},
+		Spec:       corev1.NodeSpec{PodCIDRs: []string{"192.168.77.0/24", "fd00:1::/64"}},
+	}
+
+	cl := createFakeClient(t, node)
+
+	cidrs, err := policy.FindPodCIDRs(context.Background(), cl, "")
```
bigkevmcd:
```suggestion
	cidrs, err := policy.FindPodCIDRs(t.Context(), cl, "")
```

pkg/controller/policy/networkpolicy_test.go
```diff
@@ -0,0 +1,115 @@
...
+		Spec:       corev1.NodeSpec{PodCIDR: "192.168.77.0/24"},
+	}
+	node2 := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{Name: "node-2"},
+		Spec:       corev1.NodeSpec{PodCIDR: "192.168.78.0/24"},
+	}
+
+	cl := createFakeClient(t, node1, node2)
+
+	cidrs, err := policy.FindPodCIDRs(context.Background(), cl, "")
```
bigkevmcd:
```suggestion
	cidrs, err := policy.FindPodCIDRs(t.Context(), cl, "")
```

pkg/controller/policy/networkpolicy_test.go
```diff
@@ -0,0 +1,115 @@
...
+func TestFindPodCIDRs_CannotDetermine(t *testing.T) {
+	// no --cluster-cidr override, and the Node carries no PodCIDR at all (e.g. Cilium with
+	// cluster-pool IPAM) -- must fail open (empty list) rather than error
+	node := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{Name: "node-1"},
+	}
+
+	cl := createFakeClient(t, node)
+
+	cidrs, err := policy.FindPodCIDRs(context.Background(), cl, "")
```
bigkevmcd:
```suggestion
	cidrs, err := policy.FindPodCIDRs(t.Context(), cl, "")
```

tests/e2e/cluster_network_test.go
```diff
@@ -92,4 +100,33 @@ var _ = When("two virtual clusters are installed", Label(e2eTestLabel), Label(ne
 		Expect(err).To(HaveOccurred())
 		Expect(stdout).To(Not(ContainSubstring("Welcome to nginx!")))
 	})
+
+	It("excludes the real host pod CIDR from the isolation NetworkPolicy, not a hardcoded guess", func() {
+		ctx := context.Background()
```
bigkevmcd: I think you can change this

```go
It("excludes the real host pod CIDR from the isolation NetworkPolicy, not a hardcoded guess", func() {
```

to

```go
It("excludes the real host pod CIDR from the isolation NetworkPolicy, not a hardcoded guess", func(ctx context.Context) {
```

Rather than using `context.Background()`.

tests/e2e/cluster_pod_test.go
```diff
@@ -338,6 +338,80 @@ var _ = Context("In a shared cluster", Label(e2eTestLabel), Ordered, func() {
 		})
 	})
 
+	When("updating the labels of a synced Pod", func() {
+		var virtualPod *corev1.Pod
+
+		BeforeEach(func() {
```
bigkevmcd: Again, I think you can declare the `func()` here to be `func(ctx context.Context)`

tests/e2e/cluster_pod_test.go
```diff
@@ -338,6 +338,80 @@ var _ = Context("In a shared cluster", Label(e2eTestLabel), Ordered, func() {
...
+			}
+
+			var err error
+
+			virtualPod, err = virtualCluster.Client.CoreV1().Pods(p.Namespace).Create(ctx, p, metav1.CreateOptions{})
+			Expect(err).To(Not(HaveOccurred()))
+		})
+
+		It("should keep the clusterName isolation label on the host Pod", func() {
+			ctx := context.Background()
```
bigkevmcd: See above about `func()`.

bigkevmcd (approved):
Nice solution.

These are all minor or `nit` feel free to ignore :-)

pkg/controller/policy/networkpolicy.go
enrichman: I guess it's worth to use a Kubernetes Set then. I am not sure if it's always the case but they could be both set, probably for backards/legacy reasons.

I.e. in my current env:

```yaml
spec:
  podCIDR: 10.42.0.0/24
  podCIDRs:
  - 10.42.0.0/24
  providerID: rke2://rke2-lab-0
```

enrichman: :ok_hand: 52b60e0be37519c98d388ac0e1e245ad52cf8fae

enrichman:
Thank you @bigkevmcd, TIL about the `func(ctx context.Context)`, nice one! :smile:

galal-hussein (approved):

Issue #380, opened 2025-06-15
Backup and Restore virtual clusters


jp-gouin:
When `dynamic` persistence is enabled, it should be possible to backup a virtual cluster ETCD database. 
 It should be possible to restore the ETCD volume and to specify a virtual cluster to use the restored volume.

Out of scope: snapshot volumes of running workloads - third party solution should be documented

PR #993, opened 2026-07-07
feat: ETCD Snapshots

galal-hussein:
Adding Snapshot feature to k3k:

- Add ETCDSnapshot and ETCDSnapshotList types
- Add Snapshot Controller to k3k
- Add snapshot operations helpers to k3s client pkg including SaveSnapshot/DeleteSnapshot/ListSnapshots

- #380


pkg/k3s/snapshot.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (c *Client) SaveSnapshot(snapshot *v1beta1.ETCDSnapshot, s3Config *EtcdS3) (*SnapshotResult, error) {
+	endpoint := "/db/snapshot"
+
+	req := SnapshotRequest{
+		Operation: SnapshotOperationSave,
+		Name:      []string{snapshot.Name},
+		Dir:       new(snapshot.Spec.SnapshotDir),
+		Compress:  new(snapshot.Spec.SnapshotCompress),
+		S3:        s3Config,
```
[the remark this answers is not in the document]
galal-hussein: it compiles with go 1.26

pkg/k3s/snapshot.go
```diff
@@ -0,0 +1,121 @@
...
+func (c *Client) DeleteSnapshot(snapshot *v1beta1.ETCDSnapshot, s3Config *EtcdS3) (*SnapshotResult, error) {
+	endpoint := "/db/snapshot"
+
+	k3sSnapshotName := filepath.Base(snapshot.Status.Address)
+
+	req := SnapshotRequest{
+		Operation: SnapshotOperationDelete,
+		Name:      []string{k3sSnapshotName},
+		Dir:       new(snapshot.Spec.SnapshotDir),
+		S3:        s3Config,
```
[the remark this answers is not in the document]
galal-hussein: it does compile with go 1.26

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,212 @@
...
+	snapshotFileList, err := k3sClient.ListSnapshots(s3Config)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range snapshotFileList.Items {
+		if file.Spec.SnapshotName == snapshotName {
+			snapshotFile = &file
+		}
+	}
```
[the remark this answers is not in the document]
galal-hussein: this is safe since 1.22 I think

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +885,90 @@ type VirtualClusterPolicyList struct {
...
+	// +kubebuilder:default={}
+	// +optional
+	Status ETCDSnapshotStatus `json:"status,omitempty"`
+}
+
+// ETCDSnapshotSpec defines the desired state of a ETCDSnapshot.
+type ETCDSnapshotSpec struct {
+	// ClusterName is the name of the cluster where the snapshot will be taken
+	//
+	// +optional
```
enrichman: This should be required, not optional:
```suggestion
	// +kubebuilder:validation:MinLength=1
	// +required
```

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +885,90 @@ type VirtualClusterPolicyList struct {
...
+	S3ConfigSecret *corev1.SecretReference `json:"s3ConfigSecret,omitempty"`
+
+	// SnapshotDir defines the location where the snapshot will be created, if
+	// left empty k3k will use the directory configured for k3s when the
+	// cluster was created
+	//
+	// +optional
+	SnapshotDir string `json:"snapshotDir,omitempty"`
```
enrichman: I think we can omit the Snapshot prefix, it stutters a bit. What about `Dir`, `Directory` or `TargetDir`?

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +885,90 @@ type VirtualClusterPolicyList struct {
...
+
+	// SnapshotCompress specifies if the snapshot should be compressed
+	//
+	// +optional
+	SnapshotCompress bool `json:"snapshotCompress,omitempty"`
```
enrichman: Same as before, I would simplify this as just `Compress`

```suggestion
	Compress bool `json:"compress,omitempty"`
```

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +885,90 @@ type VirtualClusterPolicyList struct {
...
+}
+
+// ETCDSnapshotStatus reflects the observed state of a ETCDSnapshot.
+type ETCDSnapshotStatus struct {
+	// Address is the absolute file:// or s3:// URI address of the snapshot.
+	//
+	// +optional
+	Address string `json:"address,omitempty"`
```
enrichman: What about `Location`? It mirrors better with the upstream https://github.com/k3s-io/k3s/blob/main/pkg/etcd/snapshot/types.go#L62

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +885,90 @@ type VirtualClusterPolicyList struct {
...
+
+	// CreationTime is the timestamp when the snapshot was taken by etcd.
+	//
+	// +optional
+	CreationTime *metav1.Time `json:"creationTime,omitempty" column:""`
+
+	// ReadyToUse indicates that the snapshot is available to be restored.
+	//
+	// +optional
+	ReadyToUse *bool `json:"readyToUse,omitempty"`
```
enrichman: Hm, I'm not convinced about this. It's a bit long/strange. If we want to keep it as a bool I would just go with a `Ready`, and we can probably even omit the optional and omitempty (it's either ready or not).

Otherwise I would also evaluate a more precise Enum, like we have in Cluster, even though I know it's not really the best pattern to adopt.

/cc @bigkevmcd

pkg/controller/cluster/cluster.go
```diff
@@ -91,12 +91,12 @@ type ClusterReconciler struct {
 	RestCfg         *rest.Config
 	PortAllocator   *agent.PortAllocator
 
-	record.EventRecorder
+	EventRecorder events.EventRecorder
```
enrichman: If the name is the same of the interface it can be omitted
```suggestion
	events.EventRecorder
```

pkg/controller/cluster/status.go
```diff
@@ -52,7 +52,7 @@ func (c *ClusterReconciler) updateStatus(ctx context.Context, cluster *v1beta1.C
 			Message: reconcileErr.Error(),
 		})
 
-		c.Event(cluster, corev1.EventTypeWarning, ReasonValidationFailed, reconcileErr.Error())
+		c.EventRecorder.Eventf(cluster, nil, corev1.EventTypeWarning, ReasonValidationFailed, "ReconcilingResource", reconcileErr.Error())
```
enrichman: I would create a const for the `action`s:

```go
ActionReconciling = "Reconciling"
```

enrichman:
```suggestion
		c.Eventf(cluster, nil, corev1.EventTypeWarning, ReasonValidationFailed, ActionReconciling, reconcileErr.Error())
```

pkg/controller/cluster/status.go
```diff
@@ -79,7 +79,7 @@ func (c *ClusterReconciler) updateStatus(ctx context.Context, cluster *v1beta1.C
 			Message: reconcileErr.Error(),
 		})
 
-		c.Event(cluster, corev1.EventTypeWarning, ReasonProvisioningFailed, reconcileErr.Error())
+		c.EventRecorder.Eventf(cluster, nil, corev1.EventTypeWarning, ReasonProvisioningFailed, "ReconcilingResource", reconcileErr.Error())
```
enrichman: Same
```suggestion
		c.Eventf(cluster, nil, corev1.EventTypeWarning, ReasonProvisioningFailed, ActionReconciling, reconcileErr.Error())
```

pkg/controller/cluster/status.go
```diff
@@ -95,7 +95,7 @@ func (c *ClusterReconciler) updateStatus(ctx context.Context, cluster *v1beta1.C
 
 	// Only emit event on transition to Ready
 	if !meta.IsStatusConditionPresentAndEqual(cluster.Status.Conditions, ConditionReady, metav1.ConditionTrue) {
-		c.Event(cluster, corev1.EventTypeNormal, ReasonProvisioned, newCondition.Message)
+		c.EventRecorder.Eventf(cluster, nil, corev1.EventTypeNormal, ReasonProvisioned, "ReconcilingResource", newCondition.Message)
```
enrichman: Same
```suggestion
		c.Eventf(cluster, nil, corev1.EventTypeNormal, ReasonProvisioned, ActionReconciling, newCondition.Message)
```

pkg/controller/snapshot/s3.go
```diff
@@ -0,0 +1,87 @@
...
+	"time"
+
+	"github.com/go-logr/logr"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/rancher/k3k/pkg/k3s"
+)
+
+func getS3ConfigFromSecret(s3ConfigSecret *corev1.Secret, logger logr.Logger) (*k3s.EtcdS3, error) {
```
enrichman: This func returns an error, but we are never failing.

Should we fail, instead of just logging? I.e. with the parse/validations:

```go
logger.V(1).Error(err, "Failed to parse etcd-s3-timeout value from S3 config secret")
```

pkg/controller/snapshot/s3.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	// Set timeout from secret if set
+	if v, ok := s3ConfigSecret.Data["etcd-s3-timeout"]; ok {
+		if duration, err := time.ParseDuration(string(v)); err != nil {
+			logger.V(1).Error(err, "Failed to parse etcd-s3-timeout value from S3 config secret")
+		} else {
+			etcdS3.Timeout.Duration = duration
+		}
+	}
```
enrichman: I think we should return an error, this means the secret is somehow misconfigured. Also I think an Error should not be in a less verbose mode (debug), but it should be always logged:

```suggestion
	if v, ok := s3ConfigSecret.Data["etcd-s3-timeout"]; ok {
		duration, err := time.ParseDuration(string(v))
		if err != nil {
			logger.V(1).Error(err, "Failed to parse etcd-s3-timeout value from S3 config secret")
			return nil, err
		}
		
		etcdS3.Timeout.Duration = duration
	}
```

pkg/controller/snapshot/s3.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// configure insecure http, if value can be parsed
+	if v, ok := s3ConfigSecret.Data["etcd-s3-insecure"]; ok {
+		if b, err := strconv.ParseBool(string(v)); err != nil {
+			logger.V(1).Error(err, "Failed to parse etcd-s3-insecure value from S3 config secret")
+		} else {
+			etcdS3.Insecure = b
+		}
+	}
```
enrichman: Same as before (return err?).

pkg/controller/snapshot/s3.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// encode CA bundles from value, and keys in configmap if one is named
+	caBundles := []string{}
+	// Add inline CA bundle if set
+	if len(s3ConfigSecret.Data["etcd-s3-endpoint-ca"]) > 0 {
+		caBundles = append(caBundles, base64.StdEncoding.EncodeToString(s3ConfigSecret.Data["etcd-s3-endpoint-ca"]))
+	}
+
+	// Concatenate all requested CA bundle strings into config var
+	etcdS3.EndpointCA = strings.Join(caBundles, " ")
```
enrichman, line 115: I am a bit unsure about this part. Also we are missing the `etcd-s3-endpoint-ca-name` field: https://docs.k3s.io/cli/etcd-snapshot#s3-configuration-secret-support

If we are not going to support that maybe we should at least log?

pkg/apis/k3k.io/v1beta1/types.go
galal-hussein: I was actually trying to mimic what we have for k3s api https://github.com/k3s-io/api/blob/main/k3s.cattle.io/v1/types.go#L96-L106

pkg/controller/snapshot/s3.go
galal-hussein: I think the way k3s do parse it is a bit more tolerant, thats why it just fallback to default instead of failing, but that makes sense too, I will return an error

pkg/apis/k3k.io/v1beta1/types.go
bigkevmcd: This looks like a ref name? Should this be `ClusterRef` ?

https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#object-references

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +885,90 @@ type VirtualClusterPolicyList struct {
 
 	Items []VirtualClusterPolicy `json:"items"`
 }
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:JSONPath=".status.location",name="Address",type="string"
+// +kubebuilder:printcolumn:JSONPath=".status.size",name=Size,type=string
+// +kubebuilder:printcolumn:JSONPath=".status.readyToUse",name="Ready To Use",type=boolean
```
bigkevmcd: Maybe this should come from a condition?

```go
// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status",description=""
```

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +885,90 @@ type VirtualClusterPolicyList struct {
...
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	ClusterName string `json:"clusterName"`
+
+	// S3ConfigSecret defines the S3 configuration secret that contains all
+	// s3 configuration, the configuration items are expected to match the following
+	// https://docs.k3s.io/cli/etcd-snapshot?_highlight=snapshot#s3-compatible-object-store-support
+	//
+	// +optional
+	S3ConfigSecret *corev1.SecretReference `json:"s3ConfigSecret,omitempty"`
```
bigkevmcd: This is a reference too? should be `S3ConfigSecretRef`?

pkg/apis/k3k.io/v1beta1/types.go
bigkevmcd: This feels like it should be a condition?

pkg/controller/cluster/server/service.go
```diff
@@ -171,6 +171,6 @@ func ServiceName(clusterName string) string {
 	return controller.SafeConcatNameWithPrefix(clusterName, "service")
 }
 
-func headlessServiceName(clusterName string) string {
+func HeadlessServiceName(clusterName string) string {
```
bigkevmcd, line 176: nit: This is an exported function, it should have a doc comment?

pkg/controller/cluster/token.go
```diff
@@ -115,7 +115,7 @@ func TokenSecretName(clusterName string) string {
 	return controller.SafeConcatNameWithPrefix(clusterName, "token")
 }
 
-func getClusterToken(ctx context.Context, client client.Client, cluster *v1beta1.Cluster) (string, error) {
+func GetClusterToken(ctx context.Context, client client.Client, cluster *v1beta1.Cluster) (string, error) {
```
bigkevmcd, line 121: nit: This is exported and should have a doc comment?

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,222 @@
...
+			// remove the snapshot from k3s cluster
+			log.Info("Deleting snapshot from cluster", "Location", snapshot.Status.Location)
+
+			snapshotResp, err := client.DeleteSnapshot(&snapshot, s3Config)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+
+			if len(snapshotResp.Deleted) <= 0 {
+				return reconcile.Result{}, fmt.Errorf("failed to delete snapshot")
```
bigkevmcd: If there were none to delete, should this prevent finalisation?

pkg/k3s/snapshot.go
```diff
@@ -0,0 +1,121 @@
...
+	Region        string          `json:"region,omitempty"`
+	SecretKey     string          `json:"secretKey,omitempty"`
+	SessionToken  string          `json:"sessionToken,omitempty"`
+	Insecure      bool            `json:"insecure,omitempty"`
+	SkipSSLVerify bool            `json:"skipSSLVerify,omitempty"`
+	Retention     int             `json:"retention,omitempty"`
+	Timeout       metav1.Duration `json:"timeout,omitempty"`
+}
+
+var DefaultEtcdS3 = &EtcdS3{
```
bigkevmcd, line 36: nit: this is exported, it should have a doc-comment.

pkg/k3s/snapshot.go
```diff
@@ -0,0 +1,121 @@
...
+	Endpoint: "s3.amazonaws.com",
+	Region:   "us-east-1",
+	Timeout: metav1.Duration{
+		Duration: 5 * time.Minute,
+	},
+	Retention: 5,
+}
+
+var ErrSnapshotRequest = errors.New("failed to execute snapshot request")
```
bigkevmcd: nit: This is exported, it's common to explain when that error is returned.

https://github.com/kubernetes/kubernetes/blob/0fe889a76062e58098cb7582ee7aa91cf2649812/staging/src/k8s.io/cli-runtime/pkg/resource/builder.go#L134

https://github.com/kubernetes/kubernetes/blob/0fe889a76062e58098cb7582ee7aa91cf2649812/vendor/go.etcd.io/raft/v3/raft.go#L88

pkg/k3s/snapshot.go
```diff
@@ -0,0 +1,121 @@
...
+
+type SnapshotOperation string
+
+const (
```
bigkevmcd, line 58: nit: Exported...

https://go.dev/wiki/CodeReviewComments#doc-comments

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +885,90 @@ type VirtualClusterPolicyList struct {
...
+
+	// Dir defines the location where the snapshot will be created, if
+	// left empty k3k will use the directory configured for k3s when the
+	// cluster was created
+	//
+	// +optional
+	Dir string `json:"dir,omitempty"`
+
+	// cifies if the snapshot should be compressed
```
bigkevmcd:
```suggestion
	// Specifies if the snapshot should be compressed
```

bigkevmcd (changes requested):
There are some nits around exported names being exported.

Otherwise looks pretty good.

pkg/apis/k3k.io/v1beta1/types.go
galal-hussein: it is a reference to a secret yes, I will fix

pkg/controller/snapshot/snapshot.go
galal-hussein: hmm, I guess this could happen if the snapshots on disk/s3 were deleted manually, okay I think if we did not recieve an error from k3s, then I think we are safe to return from reconciliation, will fix.

pkg/apis/k3k.io/v1beta1/types.go
galal-hussein: Ref usually points to name/namespace, however this is different, since we are pointing to the cluster in the same namespace, I am open to change it what do you think?

pkg/apis/k3k.io/v1beta1/types.go
galal-hussein: it should be a condition, but I am trying to be similar to what k3s api has for etcdsnapshotfiles, https://github.com/k3s-io/api/blob/main/k3s.cattle.io/v1/types.go#L96-L105

pkg/apis/k3k.io/v1beta1/types.go
bigkevmcd: I don't feel very strongly about this, when I saw the field it wasn't clear to me what it was the name of, where `Ref` would've told me that it pointed to a cluster.

It could be a https://pkg.go.dev/k8s.io/api/core/v1#LocalObjectReference (which is used to link secrets into pods amongst other things).

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +884,67 @@ type VirtualClusterPolicyList struct {
...
+	// left empty k3k will use the directory configured for k3s when the
+	// cluster was created
+	//
+	// +optional
+	Dir string `json:"dir,omitempty"`
+
+	// Specifies if the snapshot should be compressed
+	//
+	// +optional
+	Compress bool `json:"compress,omitempty"`
```
```suggestion
	// Compress specifies if the snapshot should be compressed
	//
	// +optional
	Compress bool `json:"compress,omitempty"`
```

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +884,67 @@ type VirtualClusterPolicyList struct {
...
+}
+
+// ETCDSnapshotStatus reflects the observed state of a ETCDSnapshot.
+type ETCDSnapshotStatus struct {
+	// SnapshotFileName is the name of the ETCDSnapshotFile object in the virtual
+	// cluster that backs this snapshot.
+	//
+	// +optional
+	SnapshotFileName string `json:"snapshotFileName,omitempty"`
```
enrichman: Maybe only FileName words a little better:

```suggestion
	// FileName is the name of the ETCDSnapshotFile object in the virtual
	// cluster that backs this snapshot.
	//
	// +optional
	FileName string `json:"snapshotFileName,omitempty"`
```

also the `printcolumn` should be renamed eventually.

Is it an Object, or the path of a file? Just because in case of an object maybe a FileRef could fit better, but since this is a string I am assuming is a path.

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/rancher/k3k/pkg/controller/cluster/server"
+	"github.com/rancher/k3k/pkg/k3s"
+)
+
+const (
+	snapshotController    = "k3k-snapshot-controller"
+	snapshotFinalizerName = "snapshot.k3k.io/finalizer"
+)
+
+type SnapshotReconciler struct {
```
enrichman: IMHO this can be just a `Reconciler`, because it lives in the `snapshot` package.

From an external usage this will be `snapshot.SnapshotReconciler`, but it stutters, so it's easier to read as `snapshot.Reconciler`.

Note: if this was copy-pasted we should probably rename the others, probably not in this PR. I will open an issue eventually.

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,219 @@
...
+	Client        client.Client
+	K8sClient     *kubernetes.Clientset
+	RestConfig    *rest.Config
+	Scheme        *runtime.Scheme
+	PortAllocator *agent.PortAllocator
+
+	events.EventRecorder
```
enrichman: It seems that ALL these fields are unused. I would drop them, if that's the case.

I think we should emit some events though.

```suggestion
```

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// Add adds a new controller to the manager
```
enrichman: nit

```suggestion
// Add adds a new snapshot controller to the manager
```

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,219 @@
...
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1beta1.ETCDSnapshot{}).
+		WithOptions(ctrlcontroller.Options{MaxConcurrentReconciles: maxConcurrentReconciles}).
+		Complete(&reconciler)
+}
+
+func (r *SnapshotReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := log.FromContext(ctx)
+
+	log.V(1).Info("reconciling snapshot")
```
enrichman: This is probably the only info we need to log in Info, to check the Snapshot was triggered:
```suggestion
	log := log.FromContext(ctx)
	log.Info("Reconciling ETCDSnapshot")
```

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,219 @@
...
+		cluster.Namespace +
+		":6443"
+
+	client := k3s.New(k3s.ClientConfig{
+		Token:    token,
+		ServerIP: endpoint,
+	})
+
+	// if DeletionTimestamp is not Zero -> finalize the object
+	if !snapshot.DeletionTimestamp.IsZero() {
```
enrichman: I would move the deletion right on top, in its own func, like we have in the Cluster Reconciler.

This is because it's a "special" case that it's better to handle immediately:

```go
	// if DeletionTimestamp is not Zero -> finalize the object
	if !cluster.DeletionTimestamp.IsZero() {
		return c.finalizeCluster(ctx, &cluster)
	}
```

I.e. with the current implementation a Cluster deletion made before a ETCDSnapshot deletion will cause an infinite loop here. The get Cluster will trigger a NotFound error, and a requeue, and the snapshot will never be deleted (or synced?).

enrichman: I would move the deletion right on top, in its own func, like we have in the Cluster Reconciler.

This is because it's a "special" case that it's better to handle immediately:

```go
	// if DeletionTimestamp is not Zero -> finalize the object
	if !cluster.DeletionTimestamp.IsZero() {
		return c.finalizeCluster(ctx, &cluster)
	}
```

I.e. with the current implementation a Cluster deletion made before a ETCDSnapshot deletion will cause an infinite loop here. The get Cluster will trigger a NotFound error, and a requeue, and the snapshot will never be deleted (or synced?).

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+
+		return reconcile.Result{}, nil
+	}
+
+	if controllerutil.AddFinalizer(&snapshot, snapshotFinalizerName) {
+		if err := r.Client.Update(ctx, &snapshot); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
```
enrichman: Reading about this somewhere I learnt the Update will bump the `resourceVersion` of the object, triggering another reconciliation. Basically we should/can do this:

```suggestion
	if controllerutil.AddFinalizer(&snapshot, snapshotFinalizerName) {
		return reconcile.Result{}, r.Client.Update(ctx, &snapshot)
	}
```

because another pass will be done anyway. I hope. 😆

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,219 @@
...
+
+	// only request the snapshot from the first server
+	endpoint := controller.SafeConcatNameWithPrefix(cluster.Name, "server-0") +
+		"." +
+		server.HeadlessServiceName(cluster.Name) +
+		"." +
+		cluster.Namespace +
+		":6443"
+
+	client := k3s.New(k3s.ClientConfig{
```
enrichman: Can we rename this to `k3sClient`? I was a bit confused looking later in the code

```suggestion
	k3sClient := k3s.New(k3s.ClientConfig{
```

go.mod
```diff
@@ -38,6 +38,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.24.1
 )
 
+require github.com/k3s-io/api v0.1.4
```
enrichman: Nit: can we have all the deps together in the same `require` directive?

pkg/k3s/snapshot.go
```diff
@@ -0,0 +1,129 @@
...
+	Region:   "us-east-1",
+	Timeout: metav1.Duration{
+		Duration: 5 * time.Minute,
+	},
+	Retention: 5,
+}
+
+var (
+	errSnapshotRequest  = errors.New("failed to execute snapshot request")
+	ErrSnapshotNotFound = errors.New("snapshot not found")
```
enrichman: I think `errSnapshotRequest` can be exported (it's something probably useful to check form a caller point of view), and I would add a comment for each one of them.

pkg/k3s/snapshot.go
```diff
@@ -0,0 +1,129 @@
...
+		Name:      []string{snapshot.Name},
+		Dir:       new(snapshot.Spec.Dir),
+		Compress:  new(snapshot.Spec.Compress),
+		S3:        s3Config,
+	}
+
+	snapshotResult, err := do[*SnapshotResult](c, endpoint, "server", http.MethodPost, req)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", errSnapshotRequest, err)
+	}
```
enrichman: Having a look on how the `errSnapshotRequest` is used it's probably better to have separate errors, or we are going to lose _where_ the error happened.

I.e. I think if the request failed because of a timeout we will get a:
```
"failed to execute snapshot request: timeout context deadline whatever"
```

or even worst if the request was wrong for whatever reason:

```
"failed to execute snapshot request: wrong request body cannot marshal"
```

but it's then impossible to say from the logs that this was after a `SaveSnapshot`, `ListSnapshots`, or `DeleteSnapshot`.

I would add a dedicated error per request or at leaset some context around it:


```suggestion
	snapshotResult, err := do[*SnapshotResult](c, endpoint, "server", http.MethodPost, req)
	if err != nil {
		return nil, fmt.Errorf("%w: save snapshot failed: %w", errSnapshotRequest, err)
	}
```

pkg/k3s/snapshot.go
```diff
@@ -0,0 +1,129 @@
...
+	endpoint := "/db/snapshot"
+
+	req := snapshotRequest{
+		Operation: snapshotOperationList,
+		S3:        s3Config,
+	}
+
+	snapshotFileList, err := do[*k3sv1.ETCDSnapshotFileList](c, endpoint, "server", http.MethodPost, req)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", errSnapshotRequest, err)
```
enrichman: Same, we are missing some context.

pkg/k3s/snapshot.go
```diff
@@ -0,0 +1,129 @@
...
+	req := snapshotRequest{
+		Operation: snapshotOperationDelete,
+		Name:      []string{snapshot.Status.SnapshotFileName},
+		Dir:       new(snapshot.Spec.Dir),
+		S3:        s3Config,
+	}
+
+	snapshotResult, err := do[*SnapshotResult](c, endpoint, "server", http.MethodPost, req)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", errSnapshotRequest, err)
```
enrichman: Same, we are missing some context.

go.mod
enrichman: Nit: can we have all the deps together in the same `require` directive?

main.go
```diff
@@ -141,6 +142,10 @@ func run(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to add clusterpolicy controller: %v", err)
 	}
 
+	if err := snapshot.Add(ctx, mgr, maxConcurrentReconciles); err != nil {
+		return err
+	}
+
```
enrichman, line 150: We are missing the logger:
```suggestion
	logger.Info("adding snapshot controller")

	if err := snapshot.Add(ctx, mgr, maxConcurrentReconciles); err != nil {
		return err
	}

```

enrichman (changes requested):
Thanks @galal-hussein, I left some comments, I also would like to try it, but it looks generally fine to me. 😄

pkg/apis/k3k.io/v1beta1/types.go
galal-hussein: This represents a path to a file, we can point to the object directly, but due to a bug in k3s with agentless servers (which was fixed recently but not released yet), the objects on k3s are sometimes deleted, so its better to point to the snapshot name, the snapshot name is basically `k3sv1.ETCDSnapshotFile.Spec.SnapshotName`

pkg/controller/snapshot/snapshot.go
galal-hussein: Done I did separate the finalize and reconcilation, and emitted events in both

pkg/controller/snapshot/snapshot.go
galal-hussein: Done

pkg/k3s/snapshot.go
galal-hussein: I did separate the errors to represent save/list/delete errors for snaspshot requests

galal-hussein:
> Thanks @galal-hussein, I left some comments, I also would like to try it, but it looks generally fine to me. 😄

I have addressed the comments, the major changes are:

- Separated the logic for deletion and reconcilation into separate functions.
- Snapshot requests return explicit errors for the operation name.
- Added events logs for succeeded and failed operation of snapshot (creation or deletion) from the cluster.

Let me know if thats what you had in mind.

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,235 @@
...
+
+		r.Eventf(&snapshot, nil, corev1.EventTypeNormal, SuccessfulDeleteSnapshotReason, actionSnapshotReconciling, "Snapshot was successfully deleted")
+
+		return reconcile.Result{}, nil
+	}
+
+	// avoid recreating snapshot if status is populated
+	if snapshot.Status.FileName != "" {
+		return reconcile.Result{}, nil
+	}
```
enrichman, line 90: Should we check for a condition/phase here, like `SnapshotCreated`?

Also I would add a debug log to explain what's happened, maybe it was not expected, and we know where we are stop reconciling.

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,235 @@
...
+
+	if err := r.reconcileSnapshot(ctx, &snapshot); err != nil {
+		r.Eventf(&snapshot, nil, corev1.EventTypeWarning, FailedCreateSnapshotReason, actionSnapshotReconciling, err.Error())
+		return reconcile.Result{}, err
+	}
+
+	r.Eventf(&snapshot, nil, corev1.EventTypeNormal, SuccessfulCreateSnapshotReason, actionSnapshotReconciling, "Snapshot was successfully created")
```
enrichman: I would emit this event in a different place, or we should check again if the Snapshot was created. Since we are returning with no-err when we add the finalizer it means we are logging "Snapshot was successfully created" even when the creation didn't happened yet.

```go
func (r *Reconciler) reconcileSnapshot(ctx context.Context, snapshot *v1beta1.ETCDSnapshot) error {
	// ...

	if controllerutil.AddFinalizer(snapshot, snapshotFinalizerName) {
		return r.Client.Update(ctx, snapshot)
	}

	// ...
}
```

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +884,67 @@ type VirtualClusterPolicyList struct {
...
+	Compress bool `json:"compress,omitempty"`
+}
+
+// ETCDSnapshotStatus reflects the observed state of a ETCDSnapshot.
+type ETCDSnapshotStatus struct {
+	// FileName is the name of the snapshot file created in
+	// the virtual cluster
+	//
+	// +optional
+	FileName string `json:"fileName,omitempty"`
```
enrichman: @bigkevmcd was leaning a little bit more over the `filename`, instead of `fileName`. Thinking a bit more about this I probably agree as well, sorry 😅 

```suggestion
	// Filename is the name of the snapshot file created in the virtual cluster
	//
	// +optional
	Filename string `json:"filename,omitempty"`
```

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,235 @@
...
+	// FailedCreateSnapshotReason is added in an event when a snapshot is failed to be created.
+	FailedCreateSnapshotReason = "FailedCreate"
+	// SuccessfulCreateSnapshotReason is added in an event when a snapshot is successfully created.
+	SuccessfulCreateSnapshotReason = "SuccessfulCreate"
+	// FailedDeleteSnapshotReason is added in an event when a snapshot is failed to be deleted.
+	FailedDeleteSnapshotReason = "FailedDelete"
+	// SuccessfulDeleteSnapshotReason is added in an event when a snapshot is successfully deleted.
+	SuccessfulDeleteSnapshotReason = "SuccessfulDelete"
+
+	actionSnapshotReconciling = "Reconciling"
```
enrichman: Should this be the other way around?

```suggestion
	snapshotReconcilingAction = "Reconciling"
```

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,235 @@
...
+)
+
+type Reconciler struct {
+	Client client.Client
+
+	events.EventRecorder
+}
```
enrichman, line 51: I've missed the name of the Client is the same name of the interface. We can remove one of the three (!):

```suggestion
type Reconciler struct {
	client.Client
	events.EventRecorder
}
```

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,235 @@
...
+	initServerPodName := controller.SafeConcatNameWithPrefix(cluster.Name, "server-0")
+	k3sClient := k3s.New(k3s.ClientConfig{
+		Token:    token,
+		ServerIP: fmt.Sprintf("%s.%s.%s:6443", initServerPodName, server.HeadlessServiceName(cluster.Name), snapshot.Namespace),
+	})
+
+	snapshotResp, err := k3sClient.SaveSnapshot(snapshot, s3Config)
+	if err != nil {
+		return err
+	}
```
enrichman: Two minor nit here:
1. I think returning `nil, nil` is a bad habit, so I would move the check for the nil outside of the `getS3ConfigFromSecret`. Returning an `ErrS3ConfigNotFound` is probably overkill.
2. I would keep the `s3Config` closer where it's used

```suggestion
	token, err := k3kcluster.GetClusterToken(ctx, r.Client, &cluster)
	if err != nil {
		return err
	}

	initServerPodName := controller.SafeConcatNameWithPrefix(cluster.Name, "server-0")
	k3sClient := k3s.New(k3s.ClientConfig{
		Token:    token,
		ServerIP: fmt.Sprintf("%s.%s.%s:6443", initServerPodName, server.HeadlessServiceName(cluster.Name), snapshot.Namespace),
	})
	
	var s3Config *k3s.EtcdS3

	if snapshot.Spec.S3ConfigSecretRef != nil {
		s3Config, err = r.getS3ConfigFromSecret(ctx, snapshot)
		if err != nil {
			return err
		}
	}

	snapshotResp, err := k3sClient.SaveSnapshot(snapshot, s3Config)
	if err != nil {
		return err
	}
```

pkg/controller/snapshot/s3.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/rancher/k3k/pkg/apis/k3k.io/v1beta1"
+	"github.com/rancher/k3k/pkg/k3s"
+)
+
+func (r *Reconciler) getS3ConfigFromSecret(ctx context.Context, snapshot *v1beta1.ETCDSnapshot) (*k3s.EtcdS3, error) {
+	if snapshot.Spec.S3ConfigSecretRef == nil {
+		return nil, nil
+	}
+
```
enrichman: As the previous comment I would move this outside.

```suggestion
```

enrichman: As the previous comment I would move this outside.

```suggestion
```

enrichman (commented):
Some more small comments, almost there! Thanks @galal-hussein 

@bigkevmcd I would like to ask you if we should stick with the lower case, or upper case `etcd` vs `ETCD`. Because of the unexported/exported feature of the language we are sometimes forced to use it in the first letter, but it seems the official name is `etcd`, because it's not an acronym.

Probably for the same reason in `k3s` they use `k3s.Etcd`. Should we go that way as well?

pkg/controller/snapshot/snapshot.go
galal-hussein: sure we can still add a condition, but if the fileName exists, that means that backpopulation was successful and the snapshot got created on the virtual cluster, what do you think?

pkg/controller/snapshot/snapshot.go
galal-hussein: correct, I will fix this part, I verified that on a happy path we get two events emitted

bigkevmcd:
> Some more small comments, almost there! Thanks @galal-hussein
> 
> @bigkevmcd I would like to ask you if we should stick with the lower case, or upper case `etcd` vs `ETCD`. Because of the unexported/exported feature of the language we are sometimes forced to use it in the first letter, but it seems the official name is `etcd`, because it's not an acronym.
> 
> Probably for the same reason in `k3s` they use `k3s.Etcd`. Should we go that way as well?

I'd say yes.

Looking at the Kube source they use the same naming...

galal-hussein:
> > Some more small comments, almost there! Thanks @galal-hussein
> > @bigkevmcd I would like to ask you if we should stick with the lower case, or upper case `etcd` vs `ETCD`. Because of the unexported/exported feature of the language we are sometimes forced to use it in the first letter, but it seems the official name is `etcd`, because it's not an acronym.
> > Probably for the same reason in `k3s` they use `k3s.Etcd`. Should we go that way as well?
> 
> I'd say yes.
> 
> Looking at the Kube source they use the same naming...

I can change the type to ~EtcdSnapshot` instead of `ETCDSnapshot` however note that we have used ETCD name for all other fields throughout the spec to represent ETCD properties like:

```
	// +optional
	ETCDPort *int32 `json:"etcdPort,omitempty"`
...
	// ETCDServerCA specifies the etcd-server-ca cert/key pair.
	ETCDServerCA CredentialSource `json:"etcdServerCA"`

	// ETCDPeerCA specifies the etcd-peer-ca cert/key pair.
	ETCDPeerCA CredentialSource `json:"etcdPeerCA"`

```
@bigkevmcd @enrichman should those also be changed? it wont break the yaml fields since its lowercased anyhow

enrichman:
> > > Some more small comments, almost there! Thanks @galal-hussein
> > > @bigkevmcd I would like to ask you if we should stick with the lower case, or upper case `etcd` vs `ETCD`. Because of the unexported/exported feature of the language we are sometimes forced to use it in the first letter, but it seems the official name is `etcd`, because it's not an acronym.
> > > Probably for the same reason in `k3s` they use `k3s.Etcd`. Should we go that way as well?
> > 
> > 
> > I'd say yes.
> > Looking at the Kube source they use the same naming...
> 
> I can change the type to ~EtcdSnapshot`instead of`ETCDSnapshot` however note that we have used ETCD name for all other fields throughout the spec to represent ETCD properties like:
> 
> ```
> 	// +optional
> 	ETCDPort *int32 `json:"etcdPort,omitempty"`
> ...
> 	// ETCDServerCA specifies the etcd-server-ca cert/key pair.
> 	ETCDServerCA CredentialSource `json:"etcdServerCA"`
> 
> 	// ETCDPeerCA specifies the etcd-peer-ca cert/key pair.
> 	ETCDPeerCA CredentialSource `json:"etcdPeerCA"`
> ```
> 
> @bigkevmcd @enrichman should those also be changed? it wont break the yaml fields since its lowercased anyhow

I would say yes. :slightly_smiling_face:

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,266 @@
...
+	var s3Config *k3s.EtcdS3
+
+	if snapshot.Spec.S3ConfigSecretRef != nil {
+		var err error
+
+		s3Config, err = r.getS3ConfigFromSecret(ctx, snapshot)
+		if err != nil {
+			return err
+		}
+	}
```
enrichman, line 149: We already have an err in scope, so we can drop this one:
```suggestion
	if snapshot.Spec.S3ConfigSecretRef != nil {
		s3Config, err = r.getS3ConfigFromSecret(ctx, snapshot)
		if err != nil {
			return err
		}
	}
```

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,266 @@
...
```
enrichman, line 260: Same as the other comment:


```suggestion
	if snapshot.Spec.S3ConfigSecretRef != nil {
		s3Config, err = r.getS3ConfigFromSecret(ctx, snapshot)
		if err != nil {
			return err
		}
	}
```

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,266 @@
...
+
+	clusterKey := types.NamespacedName{
+		Name:      snapshot.Spec.ClusterRef.Name,
+		Namespace: snapshot.Namespace,
+	}
+
+	err := r.Get(ctx, clusterKey, &cluster)
+	if err != nil {
+		return client.IgnoreNotFound(err)
+	}
```
enrichman: Nit: if the line is short a one-liner is usually preferred to keep the error scoped into the inner block:


```suggestion
	if err := r.Get(ctx, clusterKey, &cluster); err != nil {
		return client.IgnoreNotFound(err)
	}
```

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,266 @@
...
+
+	// skip deletion if the clsuter is terminating
```
enrichman:
```suggestion
	// skip deletion if the cluster is terminating
```

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +884,68 @@ type VirtualClusterPolicyList struct {
...
+// +kubebuilder:printcolumn:JSONPath=".status.filename",name="File Name",type="string"
+
+type EtcdSnapshot struct {
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	metav1.TypeMeta   `json:",inline"`
+
+	// +required
+	Spec EtcdSnapshotSpec `json:"spec"`
+
+	// +kubebuilder:default={}
```
bigkevmcd: I'm pretty sure that because this isn't a pointer type, it'll default to `{}` because it's the zero-value of the `EtcdSnapshotSpec`, this is probably a use for [`omitzero`](https://pkg.go.dev/github.com/go-json-experiment/json#hdr-JSON_Representation_of_Go_structs).

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -884,3 +884,68 @@ type VirtualClusterPolicyList struct {
...
+	//
+	// +kubebuilder:validation:XValidation:message="s3ConfigSecretRef is immutable",rule="self == oldSelf"
+	// +optional
+	S3ConfigSecretRef *corev1.SecretReference `json:"s3ConfigSecretRef,omitempty"`
+
+	// Compress specifies if the snapshot should be compressed
+	//
+	// +kubebuilder:validation:XValidation:message="compress is immutable",rule="self == oldSelf"
+	// +optional
+	Compress bool `json:"compress,omitempty"`
```
bigkevmcd: I don't think this will quite work...it's `omitempty` and `bool` and `optional`...

I think if it's true, you can set it to false by omitting it in an edit.

You will likely need to to make it a pointer, or leave out `omitempty`.

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	// handle snapshot deletion
+	if !snapshot.DeletionTimestamp.IsZero() {
+		if err := r.finalizeSnapshot(ctx, &snapshot); err != nil {
+			r.Eventf(&snapshot, nil, corev1.EventTypeWarning, FailedDeleteSnapshotReason, snapshotReconcilingAction, err.Error())
+			return reconcile.Result{}, err
+		}
+
+		r.Eventf(&snapshot, nil, corev1.EventTypeNormal, SuccessfulDeleteSnapshotReason, snapshotReconcilingAction, "Snapshot was successfully deleted")
```
bigkevmcd: _successfully_ is probably extraneous here

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	if err := r.reconcileSnapshot(ctx, &snapshot); err != nil {
+		r.Eventf(&snapshot, nil, corev1.EventTypeWarning, FailedCreateSnapshotReason, snapshotReconcilingAction, err.Error())
+		return reconcile.Result{}, err
+	}
+
+	// only emit event when the file is actually created and populated to the status
+	if snapshot.Status.Filename != "" {
+		r.Eventf(&snapshot, nil, corev1.EventTypeNormal, SuccessfulCreateSnapshotReason, snapshotReconcilingAction, "Snapshot was successfully created")
```
bigkevmcd: Again "successfully" is probably extraneous...you're emitting "success" events...

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,266 @@
...
+		Name:      snapshot.Spec.ClusterRef.Name,
+		Namespace: snapshot.Namespace,
+	}
+
+	if err := r.Get(ctx, nn, &cluster); err != nil {
+		return err
+	}
+
+	if cluster.Status.Phase != v1beta1.ClusterReady {
+		return fmt.Errorf("cluster is not ready")
```
bigkevmcd: I wouldn't return an error in this case - it will trigger backoff behaviour.

I'd watch Clusters and trigger reconciliations of dependent EtcdSnapshots (which may already be up-to-date).

pkg/controller/snapshot/snapshot.go
```diff
@@ -0,0 +1,266 @@
...
+
+		s3Config, err = r.getS3ConfigFromSecret(ctx, snapshot)
+		if err != nil {
+			return err
+		}
+	}
+
+	snapshotResp, err := k3sClient.SaveSnapshot(snapshot, s3Config)
+	if err != nil {
+		return err
```
bigkevmcd: Maybe wrap this?

pkg/controller/snapshot/snapshot.go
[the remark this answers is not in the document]
bigkevmcd: Not quite...it will still end up pointing at the last non-local snap shot even if multiple snapshots match and given that there's no guaranteed ordering that could be odd?

It's likely that there is only one - in which case maybe just `break` as it suggests?

bigkevmcd (changes requested):

pkg/controller/snapshot/snapshot.go
galal-hussein: Can I use the pattern we already have in the code base:
```go
if reconcilerErr != nil {
		if errors.Is(reconcilerErr, k3s.ErrServerNotReady) {
			log.V(1).Info("Server not ready, requeueing")
			return reconcile.Result{RequeueAfter: time.Second * 10}, nil
		}

		return reconcile.Result{}, reconcilerErr
	}
```
We already do that for for the Cluster reconciler, I think this can save us a lot of logic handling instead of watching the cluster object?

bigkevmcd: Sure...if that's the pattern we use...it does mean there's a delay but maybe that's not important.

pkg/k3s/client.go
```diff
@@ -132,7 +135,7 @@ func (c *Client) do(endpoint, user, method string, reader io.Reader) ([]byte, er
 	}
 
 	if resp.StatusCode >= http.StatusBadRequest {
```
enrichman: I would try to match the same K3s status codes: https://github.com/k3s-io/k3s/blob/164ae1b40c2f09001458e9ae8b870cda5a5bf15f/pkg/clientaccess/token.go#L544

```go
if resp.StatusCode < 200 || resp.StatusCode > 299 {
    // ...
}
```

pkg/k3s/client.go
```diff
@@ -141,3 +144,23 @@ func (c *Client) do(endpoint, user, method string, reader io.Reader) ([]byte, er
...
+	if err := json.Unmarshal(body, &status); err != nil {
+		return fmt.Errorf("failed to unmarshal response body for failed request: %w", err)
+	}
+
+	if status.Status != metav1.StatusFailure {
+		return fmt.Errorf("error status did not match failed request status")
+	}
+
+	return &apierrors.StatusError{ErrStatus: status}
+}
```
enrichman, line 161: This can be slightly simplified, and also adding some info if the status is different from the expected. Not sure if we really need to check this though.

```go
// statusError return back the error from failed requests
func statusError(body io.Reader) error {
	var status metav1.Status

	if err := json.NewDecoder(body).Decode(&status); err != nil {
		return fmt.Errorf("failed to unmarshal response body for failed request: %w", err)
	}

	if status.Status != metav1.StatusFailure {
		return fmt.Errorf("error status did not match failed request status: %s: %s", status.Status, status.Message)
	}

	return &apierrors.StatusError{ErrStatus: status}
}
```

enrichman (commented):
A small NIT but then LGTM! Thanks @galal-hussein , great work! 👏

pkg/k3s/client.go
galal-hussein: I would keep the status in case of any server mismatch, since the status can be either failure or success

enrichman (approved):

bigkevmcd (approved):

bigkevmcd:
Looks great to me!

PR #1010, opened 2026-07-10
Move `e2e` tests to a multi-node cluster, add kubelet restart test

enrichman:
This PR adds an e2e regression test reproducing a bug introduced during the development/investigation of https://github.com/rancher/k3k/issues/443: restarting the k3k-kubelet will delete pods on **both** the host cluster
  and the virtual cluster, even though nothing about the pod itself changed. This happened also in the past, but this time it was not caught by the CI because it's happening only in a **multi-node cluster** scenario (not covered by our tests).

It changes the e2e tests to run against a multi-node cluster, extracting the logic of the HCP mode conformance tests into a composite Github Action. This should be the most common scenario in a production environment, and it enables us to create tests for multi-node clusters.

### Additional notes related to the issue (for future reference)

The deletion is driven by the `virtual-kubelet` (in [`node/podcontroller.go`](https://github.com/virtual-kubelet/virtual-kubelet/blob/67ed37d25731123d2dc80df5a40b2bdab460d2ce/node/podcontroller.go#L421-L424)):

```go
// Perform a reconciliation step that deletes any dangling pods from the provider.
// This happens only when the virtual-kubelet is starting, and operates on a "best-effort" basis.
// If by any reason the provider fails to delete a dangling pod, it will stay in the provider and deletion won't be retried.
pc.deleteDanglingPods(ctx, podSyncWorkers)
```

 - `PodController.Run()` calls `deleteDanglingPods()` once, every time the k3k-kubelet process starts.
 - `deleteDanglingPods` calls `Provider.GetPods()` (`k3k-kubelet/provider/provider.go`), which lists **host** pods filtered only by `k3k.io/clusterName == <cluster-name>` cluster-wide, not scoped to any particular node.
 - It then looks each one up in the **virtual** cluster's pod lister, which *is* scoped to `spec.nodeName == <this instance's own node>`. Anything not found there is deleted from the host.

On a single-node cluster there's only one k3k-kubelet instance, so the unscoped list and the node-scoped lister cover the same ground — no mismatch. On a **multi-node** cluster, every restarting instance sees the *entire* cluster's pods via `GetPods()`, but its own lister only knows about its own node: any pod belonging to a different node looks "dangling" and gets deleted, even though it's perfectly healthy.


.github/actions/provision-k3s-vm-workers/action.yml
```diff
@@ -0,0 +1,183 @@
...
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Install Virtualization Dependencies
+      shell: bash
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y qemu-kvm qemu-utils cloud-image-utils wget
```
galal-hussein: maybe we need to pin versions here, to avoid any possible supply chain attacks

.github/actions/provision-k3s-vm-workers/action.yml
```diff
@@ -0,0 +1,183 @@
...
+        kvm-ok
+
+    - name: Set up bridge network for VMs
+      shell: bash
+      run: |
+        # Create a bridge so both VMs share an L2 segment with unique routable IPs.
+        # Required because QEMU `-net user` gives every VM the same 10.0.2.15 NAT
+        # address, which breaks flannel VXLAN between workers.
+        sudo ip link add name k3kbr0 type bridge
+        sudo ip addr add 192.168.100.1/24 dev k3kbr0
```
galal-hussein: can add the CIDR to an environment variable to avoid any inconsistencies if we decide to change it

.github/actions/provision-k3s-vm-workers/action.yml
```diff
@@ -0,0 +1,183 @@
...
+    - name: Generate SSH Key and Cloud-Init Seeds
+      shell: bash
+      run: |
+        ssh-keygen -t rsa -b 4096 -f ./id_rsa -N ""
+        PUBKEY="$(cat ./id_rsa.pub)"
+
+        # Per-VM cloud-init: each worker gets a unique hostname and a static IP
+        # on the bridge subnet via netplan.
+        for i in $(seq 1 ${{ inputs.worker-count }}); do
+          IP="192.168.100.1${i}"  # 192.168.100.11 / 192.168.100.12 / ...
```
galal-hussein: same comment as https://github.com/rancher/k3k/pull/1010/changes#r3569809771 maybe we can add this as CIDR_PREFIX or something

galal-hussein (commented):
few comments, but LGTM for me, thank you!

galal-hussein (approved):

PR #1019, opened 2026-07-14
Fix logged events in syncer

bigkevmcd:
Some events were being logged with the incorrect namespace, this fixes this to ensure that events are logged with the correct name/namespace.

### DIFF
--- a/k3k-kubelet/controller/syncer/events.go
+++ b/k3k-kubelet/controller/syncer/events.go
@@ -63,6 +63,7 @@ func AddEventSyncer(ctx context.Context, virtMgr, hostMgr manager.Manager, clust
 // between the host and virtual clusters.
 func (s *EventSyncer) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	logger := ctrl.LoggerFrom(ctx).WithValues("cluster", s.ClusterName, "clusterNamespace", s.ClusterNamespace)
+	ctx = ctrl.LoggerInto(ctx, logger)
 
 	event := &corev1.Event{}
 	if err := s.HostClient.Get(ctx, req.NamespacedName, event); err != nil {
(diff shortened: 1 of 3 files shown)


k3k-kubelet/controller/syncer/events.go
```diff
@@ -62,7 +62,7 @@ func AddEventSyncer(ctx context.Context, virtMgr, hostMgr manager.Manager, clust
 // Reconcile implements reconcile.Reconciler and synchronizes relevant events
 // between the host and virtual clusters.
 func (s *EventSyncer) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
-	logger := ctrl.LoggerFrom(ctx).WithValues("cluster", s.ClusterName, "clusterNamespace", s.ClusterName)
+	logger := ctrl.LoggerFrom(ctx).WithValues("cluster", s.ClusterName, "clusterNamespace", s.ClusterNamespace)
```
enrichman, line 65: This syncer is also missing the LoggerInto:

```suggestion
	logger := ctrl.LoggerFrom(ctx).WithValues("cluster", s.ClusterName, "clusterNamespace", s.ClusterNamespace)
	ctx = ctrl.LoggerInto(ctx, logger)
```

Nit: should we reuse the same `log` var name?

enrichman (commented):
Thanks @bigkevmcd!

k3k-kubelet/controller/syncer/events.go
bigkevmcd: I'd turn that around and ask, why shadow the `log` package?

bigkevmcd: (the thing that it creates is a logger, and it doesn't have the same interface as the log package?)

enrichman (approved):

PR #1021, opened 2026-07-14
Fix dangling Pods

### LINKED ISSUE #520: Spike - virtual kubelet and child pod deletion
Currently when you delete the virtual kubelet, it'll delete all pods assigned to the virtual kubelet.
Similar to delete a node type of experience.

We should define if this is a wanted behaviour

enrichman:
In **shared mode** on a **multi-node** host cluster, restarting a k3k-kubelet could delete healthy synced pods, on both the host and the virtual cluster.

The deletion is driven by the `virtual-kubelet` library's `deleteDanglingPods`, which runs once every time a k3k-kubelet instance starts. It calls `Provider.GetPods()` to learn which pods the provider "has", then deletes any of them tre absent from that instance's **virtual pod lister**, which is scoped to `spec.nodeName == <this instance's node>`.

`GetPods()` listed host pods **cluster-wide** (filtered only by the `k3k.io/clusterName` label), so on a multi-node host every restarting instance saw pods belonging to *other* nodes, found them missing from its own node-scoped lister, and deleted them.

On a single-node cluster there is only one instance, so the two sets always match and the bug is not happening.

## Fix

Scope `GetPods()` to the pods this instance owns. Ownership is the **virtual pod's** `spec.nodeName` (the same signal the framework's lister uses), not the host pod's physical node. Host pods are scheduled with only a soft, sometimes-absent affinity toward their owner, so their physical placement is unreliable.

`GetPods()` now returns only the pods assigned to the virtual kubelet for that cluster, filtering them with a new `k3k.io/agentName` label.

This PR also fixes an existing bug removing the `k3k.io/clusterName` labels and k3k annotations.

### DIFF
--- a/k3k-kubelet/provider/provider.go
+++ b/k3k-kubelet/provider/provider.go
@@ -794,9 +794,12 @@ func (p *Provider) GetPods(ctx context.Context) ([]*corev1.Pod, error) {
 
 	var hostPods corev1.PodList
 
-	listOpts := client.MatchingLabels{translate.ClusterNameLabel: p.ClusterName}
+	listOpts := []client.ListOption{
+		client.InNamespace(p.ClusterNamespace),
+		client.MatchingLabels{translate.ClusterNameLabel: p.ClusterName},
+	}
 
-	if err := p.Host.Client.List(ctx, &hostPods, listOpts); err != nil {
+	if err := p.Host.Client.List(ctx, &hostPods, listOpts...); err != nil {
 		p.logger.Error(err, "Error listing pods from host cluster")
 		return nil, err
 	}
--- a/k3k-kubelet/provider/provider.go
+++ b/k3k-kubelet/provider/provider.go
@@ -15,8 +15,6 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/virtual-kubelet/virtual-kubelet/node/api"
 	"github.com/virtual-kubelet/virtual-kubelet/node/nodeutil"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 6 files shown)


k3k-kubelet/provider/provider.go
```diff
@@ -782,32 +780,54 @@ func (p *Provider) getPodFromHostCluster(ctx context.Context, hostPodName string
 // The Pods returned are expected to be immutable, and may be accessed
 // concurrently outside of the calling goroutine. Therefore it is recommended
 // to return a version after DeepCopy.
+//
+// It returns only the Pods this instance owns: those whose virtual Pod is assigned to this node
```
galal-hussein: maybe the comment should just describe what GetPods does, and we can mention the issue number, no need to describe the issue I think

k3k-kubelet/provider/provider.go
```diff
@@ -782,32 +780,54 @@ func (p *Provider) getPodFromHostCluster(ctx context.Context, hostPodName string
...
-	selector := labels.NewSelector()
+	var hostPods corev1.PodList
 
-	requirement, err := labels.NewRequirement(translate.ClusterNameLabel, selection.Equals, []string{p.ClusterName})
-	if err != nil {
-		p.logger.Error(err, "Error creating label selector for GetPods")
-		return nil, err
+	listOpts := []client.ListOption{
+		client.InNamespace(p.ClusterNamespace),
+		client.MatchingLabels{translate.ClusterNameLabel: p.ClusterName},
```
galal-hussein: As  a simpler approach I think we can have a label when a pod is synced that shows which k3k-kubelet agent has synced it, this will eliminate any additional calls to the virtual cluster:

```suggestion
		client.MatchingLabels{translate.ClusterNameLabel: p.ClusterName, translate.AgentNameLabel: p.agentHostname},
```

galal-hussein (changes requested):
I think this approach is fine to resolve the problem, but I think it can be simpler, since we are adding labels to the synced pods we can append that with the name of the k3k-kubelet agent that did the actual sync, and then GetPods can just get the right list and avoid mistakenly deleting the wrong pods

galal-hussein (approved):
LGTM, thank you!

PR #1049, opened 2026-07-21
Add completion to `--namespace` flag

enrichman:
This PR adds shell completion for the `--namespace` / `-n` flag across the `k3kcli`.

The `--namespace` / `-n` flag now suggests real namespaces from the cluster. Two completers are wired depending on what the command operates on:

- **`completeNamespaces`**: every namespace in the host cluster. Used where any namespace is a valid target: `cluster create` and `policy create`.
- **`completeClusterNamespaces`**: only the namespaces that already contain a K3k Virtual Cluster. Used where you act on an existing cluster: `cluster delete`, `cluster list`, `cluster update`, and `kubeconfig generate`.

`policy create --namespace` is repeatable (`StringSlice`), so its completion also excludes namespaces already selected on the command line, avoiding duplicate suggestions.

### Note

Cobra does not run `PersistentPreRunE` during completion, so the shared client isn't available. `completionClient` builds its own client, reading the `--kubeconfig` flag so completion targets the same cluster the command would.

[Screencast from 2026-07-21 13-56-16.webm](https://github.com/user-attachments/assets/9e1ae12f-b1b1-4977-b07d-7e35ae0d05d5)

### DIFF
--- a/cli/cmds/completion.go
+++ b/cli/cmds/completion.go
@@ -6,6 +6,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	corev1 "k8s.io/api/core/v1"
@@ -58,9 +59,18 @@ func completeNamespaces(appCtx *AppContext) cobra.CompletionFunc {
 			return nil, cobra.ShellCompDirectiveError
 		}
 
-		names := make([]string, 0, len(namespaces.Items))
+		// Exclude already selected namespaces
+		selected := sets.New[string]()
+		if vals, err := cmd.Flags().GetStringSlice("namespace"); err == nil {
+			selected.Insert(vals...)
+		}
+
+		names := []string{}
+
 		for _, ns := range namespaces.Items {
-			names = append(names, ns.Name)
+			if !selected.Has(ns.Name) {
+				names = append(names, ns.Name)
+			}
 		}
 
 		return names, cobra.ShellCompDirectiveNoFileComp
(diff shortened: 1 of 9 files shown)


cli/cmds/completion.go
```diff
@@ -24,6 +30,68 @@ var completePersistenceMode = cobra.FixedCompletions(
...
+	}
+
+	return namespaceCompletions(cmd, client)
+}
+
+// namespaceCompletions lists every namespace in the host cluster, excluding any
+// already provided to the command's "namespace" flag.
+func namespaceCompletions(cmd *cobra.Command, cl client.Client) ([]string, cobra.ShellCompDirective) {
+	var namespaces corev1.NamespaceList
+	if err := cl.List(context.Background(), &namespaces); err != nil {
```
bigkevmcd:
```suggestion
	if err := cl.List(cmd.Context(), &namespaces); err != nil {
```

cli/cmds/completion.go
```diff
@@ -24,6 +30,68 @@ var completePersistenceMode = cobra.FixedCompletions(
...
+
+	// Exclude already selected namespaces
+	selected := sets.New[string]()
+	if vals, err := cmd.Flags().GetStringSlice("namespace"); err == nil {
+		selected.Insert(vals...)
+	}
+
+	names := []string{}
+
+	for _, ns := range namespaces.Items {
```
bigkevmcd: You could make more use of the sets here...

The difference i.e. `a \ b` is the elements in `a` that are not in `b` i.e. if `names` contains all the known names, then `names \ selected` is all the elements from `names` that are not in `selected`.

Simply...in Go 😄 

```go
names := sets.New[string]()
for _, ns := range namespaces.Items {
     names.Insert(ns.Name)
}   

return sets.List(names.Difference(selected)), cobra.ShellCompDirectiveNoFileComp
```

cli/cmds/completion.go
```diff
@@ -24,6 +30,68 @@ var completePersistenceMode = cobra.FixedCompletions(
...
+	}
+
+	return clusterNamespaceCompletions(client)
+}
+
+// clusterNamespaceCompletions lists the unique namespaces that contain at least
+// one k3k Cluster.
+func clusterNamespaceCompletions(cl client.Client) ([]string, cobra.ShellCompDirective) {
+	var clusters v1beta1.ClusterList
+	if err := cl.List(context.Background(), &clusters); err != nil {
```
bigkevmcd: This could accept the context from the `*cobra.Command` ?

bigkevmcd (changes requested):

bigkevmcd (approved):

PR #1051, opened 2026-07-21
Fix cluster delete for missing clusters

### LINKED ISSUE #536: k3kcli cluster delete should error/warn when the namespace is not provided not mislead with 'deleting' message

nightcityblade:
## Summary

- verify the target cluster exists before logging deletion or removing related PVCs
- return an actionable error that names the attempted namespace and points to `--namespace`
- cover the missing-cluster behavior with a focused unit test

Fixes #536

## Testing

- `go test ./cli/cmds -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --timeout=5m ./cli/cmds`

### DIFF
--- a/cli/cmds/cluster_delete.go
+++ b/cli/cmds/cluster_delete.go
@@ -3,6 +3,7 @@ package cmds
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -47,14 +48,22 @@ func delete(appCtx *AppContext) func(cmd *cobra.Command, args []string) error {
 
 		namespace := appCtx.Namespace(name)
 
-		logrus.Infof("Deleting '%s' cluster in namespace '%s'", name, namespace)
-
 		cluster := v1beta1.Cluster{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      name,
 				Namespace: namespace,
 			},
 		}
+		if err := client.Get(ctx, ctrlclient.ObjectKeyFromObject(&cluster), &cluster); err != nil {
+			if apierrors.IsNotFound(err) {
+				return fmt.Errorf("cluster %q not found in namespace %q; specify the correct namespace with --namespace", name, namespace)
+			}
+
+			return err
+		}
+
+		logrus.Infof("Deleting '%s' cluster in namespace '%s'", name, namespace)
+
 		// keep bootstrap secrets and tokens if --keep-data flag is passed
 		if keepData {
 			// skip removing tokenSecret
--- a/cli/cmds/cluster_delete.go
+++ b/cli/cmds/cluster_delete.go
@@ -56,7 +56,7 @@ func delete(appCtx *AppContext) func(cmd *cobra.Command, args []string) error {
 		}
 		if err := client.Get(ctx, ctrlclient.ObjectKeyFromObject(&cluster), &cluster); err != nil {
 			if apierrors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


cli/cmds/cluster_delete.go
```diff
@@ -47,14 +48,22 @@ func delete(appCtx *AppContext) func(cmd *cobra.Command, args []string) error {
...
-
 		cluster := v1beta1.Cluster{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      name,
 				Namespace: namespace,
 			},
 		}
+		if err := client.Get(ctx, ctrlclient.ObjectKeyFromObject(&cluster), &cluster); err != nil {
+			if apierrors.IsNotFound(err) {
+				return fmt.Errorf("cluster %q not found in namespace %q; specify the correct namespace with --namespace", name, namespace)
```
enrichman: I would drop the suggestion about the namespace, the NotFound could be also due to a wrong provided cluster name:

```
➜ k3kcli cluster delete --namespace correct-ns wrong-cluster
Error: cluster "wrong-cluster" not found in namespace "correct-ns"; specify the correct namespace with --namespace
```

```suggestion
				return fmt.Errorf("cluster %q not found in namespace %q", name, namespace)
```

enrichman (commented):
Thanks @nightcityblade! It looks good. If you can address the only comment I left I will approve and merge it. 🙂

cli/cmds/cluster_delete.go
nightcityblade: Updated the error message and matching test in 8bc9cec to remove the namespace suggestion. The focused Go tests and golangci-lint both pass.

enrichman (approved):

PR #1053, opened 2026-07-21
Expose loadBalancerClass through LoadBalancer configuration

### LINKED ISSUE #860: expose config: Allow setting a loadBalancerClass in the loadbalancer config
**Environmental Info:**
Host Cluster Version: 
not relevant

not relevant

Host Cluster Configuration:
I have a host cluster that has 2 loadbalancer implementation for different purposes.

K3K Cluster Configuration:
not relevant

**Describe the bug:**
If I want to spin up a k3k cluster and expose it via loadbalancer, k3k always picks the default implementation. I would like to be able to configure the desired loadBalancerClass in the cluster CRD.

See https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class

**Steps To Reproduce:**
In a cluster with two or more available loadBalancer implementations, create a shared cluster with the following spec:

```yaml
apiVersion: k3k.io/v1beta1
kind: Cluster
  name: test-cluster
  namespace: testing
spec:
  expose:
    loadBalancer: {}
  mode: shared
```

**Expected behavior:**
I can set the desired `loadBalancerClass` in the CRD.

**Actual behavior:**
The controller tries to create a loadbalancer on the default loadbalancer implementation for the cluster.

ihopenre-eng:
## Summary

- Add an optional loadBalancerClass field to LoadBalancerConfig.
- Propagate the configured class to the generated Kubernetes LoadBalancer Service.
- Update the generated CRD and API reference documentation.
- Cover the propagation behavior in the service unit test.

Closes #860

## Testing

- go test ./pkg/controller/cluster/server
- go test ./... (from pkg/apis)
- go vet ./pkg/controller/cluster/server
- golangci-lint run --timeout=5m --new-from-rev=HEAD


galal-hussein (approved):
LGTM

Issue #873, opened 2026-05-28
[Bug] kubelet fails to start on clusters with cis enabled


wstutt:
**Environmental Info:**
Host Cluster Version: 
 RKE2 v1.35.4 

Node(s) CPU architecture, OS, and Version: 
sle-micro 6.2/6.1

Host Cluster Configuration:
3 Servers

K3K Cluster Configuration:
3 servers

**Describe the bug:**
Right now it appears that the k3k-system namespace needs to be exempted from PSA policy, but even then, clusters are failing to provision.

```# kubectl logs -n kube-system -l cluster=k3k-test -f
W0521 11:47:22.425285       1 client_config.go:667] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
time="2026-05-21T11:45:30Z" level=info msg="Handling backend connection request [k3k-k3k-test-server-0]"
time="2026-05-21T11:45:30Z" level=info msg="Connected to proxy" url="wss://10.42.1.131:6443/v1-k3s/connect"
time="2026-05-21T11:45:30Z" level=info msg="Remotedialer connected to proxy" url="wss://10.42.1.131:6443/v1-k3s/connect"
{"level":"info","ts":"2026-05-21T11:45:40.186704Z","caller":"traceutil/trace.go:172","msg":"trace[1748979259] transaction","detail":"{read_only:false; response_revision:369; number_of_response:1; }","duration":"176.170705ms","start":"2026-05-21T11:45:40.010519Z","end":"2026-05-21T11:45:40.186690Z","steps":["trace[1748979259] 'process raft request'  (duration: 176.060734ms)"],"step_count":1}
{"level":"info","ts":"2026-05-21T11:45:40.211000Z","caller":"traceutil/trace.go:172","msg":"trace[1629555927] transaction","detail":"{read_only:false; response_revision:370; number_of_response:1; }","duration":"200.17755ms","start":"2026-05-21T11:45:40.010809Z","end":"2026-05-21T11:45:40.210986Z","steps":["trace[1629555927] 'process raft request'  (duration: 200.096421ms)"],"step_count":1}
time="2026-05-21T11:46:37Z" level=info msg="Opening etcd client connection with endpoints [https://127.0.0.1:2379/]"
time="2026-05-21T11:46:37Z" level=info msg="Serving HTTP bootstrap from datastore for 10.42.1.73:59106"
[...]

PR #1058, opened 2026-07-22
Add PodSecurityContext and Security Context to helm chart and cluster spec

galal-hussein:
- #873 

The PR adds new options to the helm chart allowing k3k controller to be deployed in hardened `restricted` environment, the new options are:
- `securityContext`: for container security context
- `podSecurityContext`: for pod security context

The previous `cluster.spec.securityContext` represents the container security context, the PR also adds new field to cluster.spec for the `podSecurityContext` allowing users to add pod security context to servers and agents pods.

Finally the PR harden the CoreDNS by default for all cases in shared mode, which should be harmless and work out of the box.


tests/e2e/cluster_policy_pss_test.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+var _ = When("a shared mode cluster is created in a namespace with a restricted pss configured policy", Ordered, func() {
+	var (
+		ctx            context.Context
+		virtualCluster *VirtualCluster
+		vcp            *v1beta1.VirtualClusterPolicy
+	)
+
+	BeforeAll(func() {
```
bigkevmcd:
```suggestion
	BeforeAll(func(ctx context.Context) {
```

And you can remove the `ctx = context.Background()` and the `ctx` variable too.

tests/e2e/cluster_policy_pss_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+		DeferCleanup(func() {
+			fwk3k.DeleteNamespaces(k8s, ns.Name)
+
+			err = k8sClient.Delete(ctx, vcp)
+			Expect(err).To(Not(HaveOccurred()))
+		})
+	})
+
+	It("has the server and agent pods created with security context", func() {
```
bigkevmcd:
```suggestion
	It("has the server and agent pods created with security context", func(ctx context.Context) {
```

tests/e2e/cluster_policy_pss_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+		client, restConfig, kubeconfig := NewVirtualK8sClientAndKubeconfig(clusterObj)
+		virtualCluster = &VirtualCluster{
+			Cluster:    clusterObj,
+			RestConfig: restConfig,
+			Client:     client,
+			Kubeconfig: kubeconfig,
+		}
+
+		DeferCleanup(func() {
```
bigkevmcd:
```suggestion
		DeferCleanup(func(ctx context.Context) {
```

pkg/controller/cluster/cluster.go
```diff
@@ -942,7 +949,26 @@ func (c *ClusterReconciler) bindClusterRoles(ctx context.Context, cluster *v1bet
 	return err
 }
 
-func (c *ClusterReconciler) ensureAgent(ctx context.Context, cluster *v1beta1.Cluster, serviceIP, token string) error {
+func (c *ClusterReconciler) hardened(ctx context.Context, cluster *v1beta1.Cluster) (bool, error) {
```
bigkevmcd: Maybe this could have a comment to explain what the `bool` is actually returning?

pkg/controller/cluster/server/server.go
```diff
@@ -251,23 +254,43 @@ func (s *Server) podSpec(ctx context.Context, image, name string, persistent boo
...
 	securityContext := s.cluster.Spec.SecurityContext
+
 	if s.cluster.Status.Policy != nil && s.cluster.Status.Policy.SecurityContext != nil {
 		log.V(1).Info("Using securityContext configuration from policy", "policyName", s.cluster.Status.PolicyName, "clusterName", s.cluster.Name)
 		securityContext = s.cluster.Status.Policy.SecurityContext
 	}
 
-	if securityContext != nil {
-		podSpec.Containers[0].SecurityContext = securityContext
+	// harden server pod in shared mode
```
bigkevmcd:
```suggestion
	// harden server pod in shared or HCP mode
```

bigkevmcd (approved):

enrichman:
Thanks @galal-hussein. I was thinking to have a configurable securityContext, to avoid any possible issue/clash and eventually exend it easily. Basically having an object, instead of a plain bool with hardcoded values.

Something like:

```yaml
# values.yaml

host:
  hardened:
    # enabled turns on hardened mode.
    enabled: false
    # securityContext overrides the restricted-PSS default for the controller pod
    securityContext: {}
```

```yaml
# deployment.yaml
          ...
          {{- if .Values.host.hardened.enabled }}
          securityContext:
            {{- with .Values.host.hardened.securityContext }}
            {{- toYaml . | nindent 12 }}
            {{- else }}
            {{- include "k3k.restrictedSecurityContext" . | nindent 12 }}
            {{- end }}
          {{- end }}
          ...
```

```yaml
# _helpers.tpl

{{/*
Default restricted-PSS compliant securityContext
*/}}
{{- define "k3k.restrictedSecurityContext" -}}
runAsUser: 65534
runAsGroup: 65534
runAsNonRoot: true
allowPrivilegeEscalation: false
capabilities:
  drop:
  - "ALL"
seccompProfile:
  type: "RuntimeDefault"
{{- end }}
```

and I would add an additional `pss` package with the common func/things.

Not sure about the created Pods though. Do they need the same securityContext in order to run in the Host?

galal-hussein:
> Thanks @galal-hussein. I was thinking to have a configurable securityContext, to avoid any possible issue/clash and eventually exend it easily. Basically having an object, instead of a plain bool with hardcoded values.
> 
> Something like:
> 
> ```yaml
> # values.yaml
> 
> host:
>   hardened:
>     # enabled turns on hardened mode.
>     enabled: false
>     # securityContext overrides the restricted-PSS default for the controller pod
>     securityContext: {}
> ```
> 
> ```yaml
> # deployment.yaml
>           ...
>           {{- if .Values.host.hardened.enabled }}
>           securityContext:
>             {{- with .Values.host.hardened.securityContext }}
>             {{- toYaml . | nindent 12 }}
>             {{- else }}
>             {{- include "k3k.restrictedSecurityContext" . | nindent 12 }}
>             {{- end }}
>           {{- end }}
>           ...
> ```
> 
> ```yaml
> # _helpers.tpl
> 
> {{/*
> Default restricted-PSS compliant securityContext
> */}}
> {{- define "k3k.restrictedSecurityContext" -}}
> runAsUser: 65534
> runAsGroup: 65534
> runAsNonRoot: true
> allowPrivilegeEscalation: false
> capabilities:
>   drop:
>   - "ALL"
> seccompProfile:
>   type: "RuntimeDefault"
> {{- end }}
> ```
> 
> and I would add an additional `pss` package with the common func/things.
> 
> Not sure about the created Pods though. Do they need the same securityContext in order to run in the Host?

That makes more sense, however we need to consider the running pods, there are basically two cases:

1- we run with VCP policy that applies `restricted` admission level, we can fetch for that information and configure the pods accordingly.
2- is that the whole host cluster is configured with hardened CIS profile like [RKE2](https://docs.rke2.io/security/hardening_guide) which we cant really check for.

we still can document that the user can use `spec.SecurityContext` which is also a valid solution, instead of passing a hardcoded sec context, with the only exception of CoreDNS which will not run unless the code is modified.

galal-hussein:
After discussing with @enrichman I think the correct way as he suggests is as follows:
- Adding securityContext field to the chart
- Use the already existing `cluster.spec.SecurityContext` for servers and agents/k3k-kubelet pods
- Harden the synced CoreDNS addon for shared mode by default

cc @bigkevmcd

galal-hussein:
@bigkevmcd @enrichman I have applied the suggestions you made, the pr now allows the controller and shared/hcp clusters to be deployed in restricted environment, but does not enforce any security context, however it allows the users to supply their own container and pod secuirtyContexts, for testing you can install the chart in restricted environment with the following values:

```
helm upgrade --install --namespace k3k-system --create-namespace \
                --set controller.podSecurityContext.runAsNonRoot=true \
                --set controller.podSecurityContext.runAsUser=65532 \
                --set controller.podSecurityContext.runAsGroup=65532 \
                --set controller.podSecurityContext.seccompProfile.type=RuntimeDefault \
                --set controller.securityContext.runAsNonRoot=true \
                --set controller.securityContext.allowPrivilegeEscalation=false \
                --set controller.securityContext.privileged=false \
                --set controller.securityContext.readOnlyRootFilesystem=true \
                --set 'controller.securityContext.capabilities.drop[0]=ALL' \
                --set controller.securityContext.seccompProfile.type=RuntimeDefault
                k3k ./charts/k3k/
```

And the cluster.yaml
```
---
apiVersion: k3k.io/v1beta1
kind: Cluster
metadata:
  name: testcluster
  namespace: test-hardened
spec:
  agents: 0
  expose:
    nodePort: {}
  mode: shared
  persistence:
    type: ephemeral
  servers: 1
  serverArgs:
    # to allow running k3s as rootless since we are running as NonRoot
    - "--rootless"
    - "--data-dir=/var/lib/rancher/k3s"
    - "--write-kubeconfig=/var/lib/rancher/k3s/k3s.yaml"
  # applied to both the server and the shared agent (k3k-kubelet) pods
  podSecurityContext:
    runAsNonRoot: true
    runAsUser: 65534
    runAsGroup: 65534
    # the k3s data dir volume must be group owned by the user k3s runs as
    fsGroup: 65534
    fsGroupChangePolicy: OnRootMismatch
    seccompProfile:
      type: RuntimeDefault
  # applied to the server and shared agent containers
  securityContext:
    runAsNonRoot: true
    runAsUser: 65534
    runAsGroup: 65534
    privileged: false
    allowPrivilegeEscalation: false
    capabilities:
      drop:
      - ALL
    seccompProfile:
      type: RuntimeDefault
```

bigkevmcd (approved):

enrichman (approved):

PR #1069, opened 2026-07-23
Fix event-mirroring

bigkevmcd:
Lookup the InvolvedObject in the Host cluster and use the annotations on that to get the name of the Pod from the virtual cluster.

The reverse translation of the name is no longer used.

https://github.com/rancher/k3k/issues/346

### DIFF
--- a/k3k-kubelet/controller/syncer/events.go
+++ b/k3k-kubelet/controller/syncer/events.go
@@ -74,18 +74,44 @@ func (s *EventSyncer) Reconcile(ctx context.Context, req reconcile.Request) (rec
 		return reconcile.Result{}, nil
 	}
 
-	virtualRef := s.Translator.TranslateObjectReferenceFrom(event.InvolvedObject)
+	hostPod := &corev1.Pod{}
+	if err := s.HostClient.Get(ctx, client.ObjectKey{Name: event.InvolvedObject.Name, Namespace: event.InvolvedObject.Namespace}, hostPod); err != nil {
+		if client.IgnoreNotFound(err) != nil {
+			return reconcile.Result{}, fmt.Errorf("could not load host object: %w", err)
+		}
+
+		logger.V(3).Info("host object not found, skipping event emission",
+			"involvedObject.kind", event.InvolvedObject.Kind,
+			"involvedObject.name", event.InvolvedObject.Name,
+			"involvedObject.namespace", event.InvolvedObject.Namespace,
+		)
+
+		return reconcile.Result{}, nil
+	}
+
+	s.Translator.TranslateFrom(hostPod)
+
+	if hostPod.Name == "" {
+		logger.V(3).Info("Host object has no name - skipping event",
+			"involvedObject.kind", event.InvolvedObject.Kind,
+			"involvedObject.name", event.InvolvedObject.Name,
+			"involvedObject.namespace", event.InvolvedObject.Namespace,
+		)
+
+		return reconcile.Result{}, nil
+	}
 
 	// Look up the corresponding object in the virtual cluster.
 	virtPod := &corev1.Pod{}
-	if err := s.VirtualClient.Get(ctx, client.ObjectKey{Name: virtualRef.Name, Namespace: virtualRef.Namespace}, virtPod); err != nil {
+	if err := s.VirtualClient.Get(ctx, client.ObjectKey{Name: hostPod.Name, Namespace: hostPod.Namespace}, virtPod); err != nil {
 		if client.IgnoreNotFound(err) != nil {
 			return reconcile.Result{}, fmt.Errorf("could not load virtual object: %w", err)
 		}
 
 		logger.Info("virtual object not found, skipping event emission",
-			"virtualInvolvedObject.kind", virtualRef.Kind,
-			"virtualInvolvedObject.name", virtualRef.Name,
+			"virtualInvolvedObject.kind", hostPod.Kind,
+			"virtualInvolvedObject.name", hostPod.Name,
+			"virtualInvolvedObject.namespace", hostPod.Namespace,
 		)
 
 		return reconcile.Result{}, nil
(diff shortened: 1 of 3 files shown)


k3k-kubelet/controller/syncer/events.go
```diff
@@ -74,18 +74,44 @@ func (s *EventSyncer) Reconcile(ctx context.Context, req reconcile.Request) (rec
 		return reconcile.Result{}, nil
 	}
 
-	virtualRef := s.Translator.TranslateObjectReferenceFrom(event.InvolvedObject)
+	hostPod := &corev1.Pod{}
+	if err := s.HostClient.Get(ctx, client.ObjectKey{Name: event.InvolvedObject.Name, Namespace: event.InvolvedObject.Namespace}, hostPod); err != nil {
+		if client.IgnoreNotFound(err) != nil {
```
enrichman:
```suggestion
		if apierrors.IsNotFound(err){
```

Internally:
```go
// IgnoreNotFound returns nil on NotFound errors.
// All other values that are not NotFound errors or nil are returned unmodified.
func IgnoreNotFound(err error) error {
		return nil
	}
	return err
}
```

k3k-kubelet/controller/syncer/events.go
```diff
@@ -74,18 +74,44 @@ func (s *EventSyncer) Reconcile(ctx context.Context, req reconcile.Request) (rec
 		return reconcile.Result{}, nil
 	}
 
-	virtualRef := s.Translator.TranslateObjectReferenceFrom(event.InvolvedObject)
+	hostPod := &corev1.Pod{}
+	if err := s.HostClient.Get(ctx, client.ObjectKey{Name: event.InvolvedObject.Name, Namespace: event.InvolvedObject.Namespace}, hostPod); err != nil {
```
enrichman, line 79: I think we should find a convention about this, just for consistency and readability. We use a lot the controller-runtime client, and we use it in too many different ways.

Not related to the PR, and not functional though, I was just thinking about it, because of the very-long inline second arg.

I.e. I think I like the following approach:

```go
key := types.NamespacedName{
	Name: c.ClusterName,
	Namespace: c.ClusterNamespace
}

var myobj v1.MyObj
if err := myClient.Get(ctx, key, &myobj); err != nil {
		// handle err
}
```

Not sure if swapped is a little bit better:
```go

var myobj v1.MyObj

key := types.NamespacedName{
	Name: c.ClusterName,
	Namespace: c.ClusterNamespace
}

if err := myClient.Get(ctx, key, &myobj); err != nil {
		// handle err
}
```

What's your preference, or what are the guidelines about this?

(bikeshedding?)

k3k-kubelet/controller/syncer/events.go
```diff
@@ -74,18 +74,44 @@ func (s *EventSyncer) Reconcile(ctx context.Context, req reconcile.Request) (rec
...
+		if apierrors.IsNotFound(err){
+			return reconcile.Result{}, fmt.Errorf("could not load host object: %w", err)
+		}
+
+		logger.V(3).Info("host object not found, skipping event emission",
```
enrichman: I think the `debug` level when enabled will just increase the verbosity to 1, so everything above will be skipped. Eventually we could think about a `trace` level, but currently I would stick with info and debug (`V(1)`).

```suggestion
		logger.V(1).Info("host object not found, skipping event emission",
```
---
A very opinionated article, like everything about logs: https://dave.cheney.net/2015/11/05/lets-talk-about-logging

> We’ve ruled out warnings, argued that nothing should be logged at error level, and shown that only the top level of the application should have some kind of log.Fatal behaviour. What’s left ?
> 
> I believe that there are only two things you should log:
> 
>  Things that developers care about when they are developing or debugging software.
Things that users care about when using your software.
Obviously these are debug and info levels, respectively.

k3k-kubelet/controller/syncer/events.go
```diff
@@ -74,18 +74,44 @@ func (s *EventSyncer) Reconcile(ctx context.Context, req reconcile.Request) (rec
...
+			"involvedObject.kind", event.InvolvedObject.Kind,
+			"involvedObject.name", event.InvolvedObject.Name,
+			"involvedObject.namespace", event.InvolvedObject.Namespace,
```
enrichman: Is it possible to log an object/map? Not sure how that works, I think we have it somewhere and in the json format you will get a nested map, not sure about the current/default text `key=value` one.

k3k-kubelet/controller/syncer/events.go
```diff
@@ -74,18 +74,44 @@ func (s *EventSyncer) Reconcile(ctx context.Context, req reconcile.Request) (rec
...
+		)
+
+		return reconcile.Result{}, nil
+	}
+
+	s.Translator.TranslateFrom(hostPod)
+
+	if hostPod.Name == "" {
+		logger.V(3).Info("Host object has no name - skipping event",
```
enrichman:
```suggestion
		logger.V(1).Info("Host object has no name - skipping event",
```

k3k-kubelet/controller/syncer/events.go
```diff
@@ -74,18 +74,44 @@ func (s *EventSyncer) Reconcile(ctx context.Context, req reconcile.Request) (rec
...
 
 	// Look up the corresponding object in the virtual cluster.
 	virtPod := &corev1.Pod{}
-	if err := s.VirtualClient.Get(ctx, client.ObjectKey{Name: virtualRef.Name, Namespace: virtualRef.Namespace}, virtPod); err != nil {
+	if err := s.VirtualClient.Get(ctx, client.ObjectKey{Name: hostPod.Name, Namespace: hostPod.Namespace}, virtPod); err != nil {
 		if client.IgnoreNotFound(err) != nil {
 			return reconcile.Result{}, fmt.Errorf("could not load virtual object: %w", err)
 		}
 
 		logger.Info("virtual object not found, skipping event emission",
```
enrichman: Maybe this is useful on a debug level?
```suggestion
		logger.V(1).Info("virtual object not found, skipping event emission",
```

enrichman (commented):
Some random NIT comments!

enrichman (approved):

PR #1079, opened 2026-07-24
Add k3k upgrade tests

### LINKED ISSUE #559: Add test to upgrade K3k version
We need a test to validate the safe upgrade from the current supported version of K3k to the latest. 
The test should include cluster creation and ensuring no clusters are deleted in the process.

enrichman:
This PR adds an upgrade test workflow to verify that upgrading from the last released version to the main branch is safe. This will hopefully prevent us from shipping features that break existing functionality.

Note that this is just a quick smoke test: it creates virtual clusters with the latest release, upgrades them, and checks cluster availability. Since shared mode also runs pods on the host, we check a small app deployment in that scenario as well.

If needed, we can add more test cases for this scenario and refactor some common methods used across the test suite.

Note: The GitHub Action is currently part of the e2e workflow solely for testing purposes, as we cannot manually run new actions that do not yet exist in main. Before merging, I will move test-upgrade into its own dedicated GitHub Action.

Note 2: To test this locally, you need to export a `REPO` variable. This is used to push the latest images to a registry from which the Kubernetes cluster can pull.

### DIFF
--- a/tests/upgrade/helpers_test.go
+++ b/tests/upgrade/helpers_test.go
@@ -1,0 +1,186 @@
+package upgrade_test
+
+// The cluster helpers below are a copy of the ones in tests/e2e/common_test.go,
+// and runCmd is a copy of the one in tests/cli/cli_test.go. The duplication is
+// deliberate: sharing them means moving them into tests/framework, which is a
+// refactor of the e2e and cli suites and does not belong in this change. Only
+// what this suite actually uses was copied.
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os/exec"
+	"time"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/kubernetes/pkg/api/v1/pod"
+	"k8s.io/utils/ptr"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+
+	"github.com/rancher/k3k/pkg/apis/k3k.io/v1beta1"
+	"github.com/rancher/k3k/pkg/controller/certs"
+	"github.com/rancher/k3k/pkg/controller/kubeconfig"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// runCmd executes cmdName with args and returns its stdout, stderr and error.
+func runCmd(cmdName string, args ...string) (string, string, error) {
+	stdout, stderr := &bytes.Buffer{}, &bytes.Buffer{}
+
+	cmd := exec.CommandContext(context.Background(), cmdName, args...)
+	cmd.Stdout = stdout
+	cmd.Stderr = stderr
+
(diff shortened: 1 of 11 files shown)


tests/upgrade/helpers_test.go
```diff
@@ -0,0 +1,186 @@
+package upgrade_test
+
+// The cluster helpers below are a copy of the ones in tests/e2e/common_test.go,
+// and runCmd is a copy of the one in tests/cli/cli_test.go. The duplication is
+// deliberate: sharing them means moving them into tests/framework, which is a
+// refactor of the e2e and cli suites and does not belong in this change. Only
```
galal-hussein, line 6: I am not sure if thats a bad thing actually, I think we should move them to tests/framework and use less code here and avoid copying the same functions

tests/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,220 @@
...
+		clusterHCP     *v1beta1.Cluster
+		podUIDs        []types.UID
+	)
+
+	ctx := context.Background()
+
+	BeforeAll(func() {
+		// Guard early: REPO has to point at the images built from this checkout and available to every node
+		// (in CI they are tagged `k3k.local/...` and imported into containerd directly).
+		Expect(os.Getenv("REPO")).NotTo(BeEmpty(), "REPO must be set to the image repository of the build under test")
```
galal-hussein, line 38: nit: can we check directly on the existence of the images using containerd client?

Only caveat is that we will require to pull containerd dependency, maybe we can check with direct os.exec

tests/upgrade/helpers_test.go
enrichman: Yes, I tried to export them twice but the changes growth exponentially because of some shared resources that needs to be exported, and moved, so I opted to postpone the refactor to a different PR. :/

i.e. https://github.com/rancher/k3k/pull/1157/changes

tests/upgrade/upgrade_test.go
enrichman: I don't want to reintroduce the containerd dependency. We removed it because it was only used in tests and introduced a lot of transitive dependencies with CVEs that were triggering our alerts.

While it's not an issue for CI or local testing, the REPO environment variable is used by make install a few lines later to specify the registry Helm should pull the images from.

Basically, when you run `make test-upgrade`, `make install` installs the latest K3k version (a few lines down):

```go
By("Upgrading k3k to the build from source")
expectCmd(runCmd("make", "-C", "../..", "install"))
```

And this executes the upgrade:

```Makefile
.PHONY: install
install:    ## Install K3k with Helm on the targeted Kubernetes cluster
    helm upgrade --install --namespace k3k-system --create-namespace \
        --set controller.extraEnv[0].name=DEBUG \
        --set-string controller.extraEnv[0].value=true \
        --set controller.image.repository=$(REPO)/k3k \
        --set controller.image.tag=$(VERSION) \
        --set agent.shared.image.repository=$(REPO)/k3k-kubelet \
        --set agent.shared.image.tag=$(VERSION) \
        k3k ./charts/k3k/
```

So, while the image is already available (though you do need to build and push it beforehand), the subsequent helm upgrade still needs to know which registry to fetch it from.

galal-hussein (approved):

PR #1080, opened 2026-07-24
Add `cluster delete` autocompletion

### LINKED ISSUE #448: Prevent deleting wrong cluster when names are ambiguous
When two or more clusters share the same name but exist in different namespaces, running `k3kcli cluster delete <cluster-name>` without the `--namespace` flag can easily delete the wrong cluster.

The command currently defaults to the `k3k-<cluster-name>` namespace if one exists, which may not be the user's intention. This can lead to accidental data loss.

```
-> % k3kcli cluster list      
NAMESPACE   NAME                  MODE     STATUS    POLICY
default     foo                   shared   Unknown   <none>
k3k-foo     foo                   shared   Unknown   <none>
```

### Expected Behavior

The CLI should detect the name collision and fail with an error message, requiring the user to specify the target with the `--namespace` flag to prevent accidental deletion.

### LINKED ISSUE #473: Add autocompletion for `k3kcli cluster delete`
cluster delete
The k3kcli cluster delete [NAME] expects a single argument with the cluster to delete. We can provide an autocompletion to select from the available clusters.

NOTE: the command could also have a --namespace flag, to select the namespace. Different clusters could be part of different namespaces. A "workaround" could be dropping the --namespace flag, and/or handling the namespaces within the name, like k3kcli cluster delete myns/cluster. This will ease the deletion of different clusters in different namespaces.

To improve the usability of `k3kcli cluster delete [NAME]`, we should provide shell autocompletion for the cluster `[NAME]` argument. The completion logic should suggest a list of available clusters.

Following the note in the epic, we should consolidate the cluster identifier into a single namespace/cluster-name argument. This simplifies the command by removing the need for a separate `--namespace` flag and makes deleting multiple clusters across different namespaces more intuitive.

```
# User types this and presses <TAB>
$ k3kcli cluster delete myns/

# CLI suggests completions
myns/cluster-1   myns/cluster-2

# The command can handle multiple deletions
$ k3kcli cluster delete myns/cluster-1 anotherns/cluster-foo
```

## Acceptance Criteria:

- [ ] The cluster delete command accepts one or more arguments in the format `<namespace>/<cluster-name>`.
- [ ] Shell autocompletion is implemented to suggest available clusters when the user presses `<TAB>`.
- [ ] The `--namespace` flag is deprecated or removed in favor of the new argument format.

enrichman:
This PR adds shell autocompletion for the cluster name argument of `k3kcli cluster delete`, plus an `--all/-A` flag to delete every cluster in a namespace.

Completions are emitted in the Kubernetes `namespace/name` format, so the selected cluster is always unambiguous. A plain name can't tell you which namespace it lives in, and there is the possibility to have two virtual clusters with the same name in different namespaces.

For backwards compatibility `k3kcli cluster delete mycluster` still resolves to the `k3k-mycluster` namespace (or `-n` when provided).

Screenshot from 2026-07-28 15-50-43

image

### DIFF
--- a/cli/cmds/cluster_delete.go
+++ b/cli/cmds/cluster_delete.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -19,7 +20,10 @@ import (
 	k3kcluster "github.com/rancher/k3k/pkg/controller/cluster"
 )
 
-var keepData bool
+var (
+	keepData  bool
+	deleteAll bool
+)
 
 func NewClusterDeleteCmd(appCtx *AppContext) *cobra.Command {
 	cmd := &cobra.Command{
@@ -27,10 +31,13 @@ func NewClusterDeleteCmd(appCtx *AppContext) *cobra.Command {
 		Short:   "Delete an existing cluster.",
 		Example: "k3kcli cluster delete [command options] NAME",
 		RunE:    delete(appCtx),
-		Args:    cobra.ExactArgs(1),
+		Args:    cobra.MaximumNArgs(1),
 	}
 
 	cmd.Flags().BoolVar(&keepData, "keep-data", false, "keeps persistence volumes created for the cluster after deletion")
+	cmd.Flags().BoolVarP(&deleteAll, "all", "A", false, "delete all the clusters in the namespace")
+
+	cmd.ValidArgsFunction = completeClusterNames
 
 	CobraFlagNamespace(appCtx, cmd, completeClusterNamespaces)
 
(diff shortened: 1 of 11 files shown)


cli/cmds/cluster_delete.go
```diff
@@ -41,14 +48,48 @@ func delete(appCtx *AppContext) func(cmd *cobra.Command, args []string) error {
 	return func(cmd *cobra.Command, args []string) error {
 		ctx := context.Background()
```
bigkevmcd:
```suggestion
		ctx := cmd.Context()
```

cli/cmds/cluster_delete.go
```diff
@@ -41,14 +48,48 @@ func delete(appCtx *AppContext) func(cmd *cobra.Command, args []string) error {
...
+		if deleteAll {
+			if len(args) > 0 {
+				return errors.New("cannot specify a cluster name together with --all")
+			}
+
+			if appCtx.namespace == "" {
+				return errors.New("--all requires a namespace, set it with --namespace/-n")
+			}
+
+			var clusters v1beta1.ClusterList
```
bigkevmcd, line 61: Maybe use `client.DeleteAllOf()` ?

cli/cmds/completion.go
```diff
@@ -89,6 +112,27 @@ func clusterNamespaceCompletions(ctx context.Context, client client.Client) ([]s
...
+
+	var clusters v1beta1.ClusterList
+	if err := cl.List(ctx, &clusters, opts...); err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	names := make([]string, 0, len(clusters.Items))
+	for _, cluster := range clusters.Items {
+		names = append(names, cluster.Namespace+"/"+cluster.Name)
+	}
```
bigkevmcd, line 131: Slightly more efficient to allocate the slice and assign the elements?

```go
names := make([]string, len(clusters.Items))
for i, cluster := range clusters.Items {
	names[i] = cluster.Namespace+"/"+cluster.Name
}
```

bigkevmcd (approved):
Tested it and it works.

cli/cmds/cluster_delete.go
enrichman: I looked also for other contexts! 😄

cli/cmds/cluster_delete.go
enrichman: Hm, yes, that would be nice, but looking at the single deletion of a cluster we are doing some additional logic around the PVC to keep/delete the data, so it's probably sligthly safer to do this sequencially.

We can think about a `--force` or similar flag to do something like that eventually.

cli/cmds/completion.go
enrichman: Since the array is already allocated with the length I think the difference is negligible, and I'm always "scared" about assigning directly the elements. :laughing: 

If you don't mind I will keep the append!

PR #1081, opened 2026-07-26
fix: add consistent cluster pod labels

### LINKED ISSUE #531: Inconsistencies in labels for cluster pods
If I check the configuration for the server pod, I get the following:
```yaml
kind: Pod
metadata:
  labels:
    cluster: mycluster
    role: server
```

But for the kubelet pod, it shows:
```yaml
kind: Pod
metadata:
  labels:
    cluster: mycluster
    mode: shared
    type: agent
```

I was thinking maybe it would be useful to change `type: agent` to `role: agent` for consistency, perhaps?
Also, maybe adding the mode label to the server, though I do not know if that's useful?

I would be willing to implement this myself if I could get a couple pointers in the right direction.

sidsri14:
## Summary

- add consistent `role`, `mode`, and `k3k.io/clusterName` labels to server and agent workloads and pod templates
- use the same complete label sets for workload selectors and metadata
- retain the existing `cluster` and agent `type` labels
- centralize the label maps and cover them in the existing package tests

Closes #531

## Validation

- `go test ./pkg/controller/cluster/agent ./pkg/controller/cluster/server`
- repository unit target equivalent (16 Ginkgo suites passed; integration and E2E suites skipped)
- `go vet ./pkg/controller/cluster/agent ./pkg/controller/cluster/server`
- `golangci-lint v2.12.2` new-change scan for the touched packages (0 issues)

A package-wide lint scan also reports existing `gci` findings in three untouched files: `pkg/controller/cluster/agent/ports.go`, `pkg/controller/cluster/agent/virtual_test.go`, and `pkg/controller/cluster/server/config.go`.


pkg/controller/cluster/agent/agent.go
```diff
@@ -29,6 +29,24 @@ type Config struct {
 	scheme  *runtime.Scheme
 }
 
+func agentSelectorLabels(clusterName, mode string) map[string]string {
+	return map[string]string{
+		"cluster": clusterName,
+		"type":    "agent",
+		"mode":    mode,
```
galal-hussein: the agent selector is missing the new labels here

pkg/controller/cluster/agent/labels_test.go
```diff
@@ -0,0 +1,25 @@
+package agent
```
galal-hussein: no need for a new file here, this can be added to the existing tests

galal-hussein (changes requested):

pkg/controller/cluster/server/server.go
```diff
@@ -52,6 +52,22 @@ type Server struct {
 	imagePullSecrets []string
 }
 
+func serverSelectorLabels(clusterName string) map[string]string {
+	return map[string]string{
+		"cluster": clusterName,
+		"role":    "server",
```
galal-hussein: same as https://github.com/rancher/k3k/pull/1081/changes#r3712655994

pkg/controller/cluster/agent/agent.go
sidsri14: Updated in af45864: agent selectors now use the same complete agentLabels map as workload and pod metadata, including k3k.io/clusterName and role=agent.

pkg/controller/cluster/agent/labels_test.go
sidsri14: Updated in af45864: moved TestAgentLabels into shared_test.go and removed the standalone labels_test.go file.

pkg/controller/cluster/server/server.go
sidsri14: Updated in af45864: the server selector now uses the complete serverLabels map, matching the StatefulSet and pod-template metadata.

Issue #844, opened 2026-05-07
Cold start - cluster reset failed


xelab04:
**Environmental Info:**
Host Cluster Version: 
3 K3s nodes, `v1.35.4+k3s1`

Node(s) CPU architecture, OS, and Version: 
openSUSE Leap 15.6, x86, kernel version 6.4

Host Cluster Configuration:
3 nodes both control plane and agent.

K3K Cluster Configuration:
```yaml
apiVersion: k3k.io/v1beta1
kind: Cluster
spec:
  agents: 0
  clusterCIDR: 10.42.0.0/16
  clusterDNS: 10.43.0.10
  mode: shared
  persistence:
    storageRequestSize: 2G
    type: dynamic
  serverLimit:
    cpu: 600m
    memory: 1000Mi
  servers: 1
  serviceCIDR: 10.43.0.0/16
  sync:
  tlsSANs:
  - k-1-cvetest.mydomain
  workerLimit:
    cpu: 30m
    memory: 75Mi
```

**Describe the bug:**
```bash
> k logs -n k3k-k-1-cvetest k3k-k-1-cvetest-server-0
[INFO] [Thu May  7 09:19:30 2026] Starting single node setup...
[INFO] [Thu May  7 09:19:30 2026] Existing data found in single node setup. Performing cluster-reset to ensure quorum...
[FATAL] [Thu May  7 09:19:32 2026]  cluster reset failed!
```

**Steps To Reproduce:**
- Created a cluster through a yaml manifest, resembling the manifest from above.
- Use cluster for a while.
- Hard power off nodes. The k3k cluster will not come back up

**Expected behavior:**
The K3k cluster would pass the cluster-reset.

xelab04:
I don't know if it's helpful, but I edited the statefulset for the server pod and changed the command to not check whether the cluster reset failed. The pod then ran (as in didn't immediately crash) and I now see the following error:

```bash
time="2026-05-07T10:00:11Z" level=info msg="Failed to test etcd connection: failed to get etcd status: rpc error: code = Unavailable desc = connection error: desc = \"transport: authentication handshake failed: context deadline exceeded\""
``` 
So is it possibly a problem with etcd? I've only noticed this problem with old clusters, where maybe the etcd store is too full?

xelab04:
I exec'd into the pod and restored the etcd database from a previous snapshot. Though that caused another problem:
```
[...]

PR #1117, opened 2026-08-04
fix(server pod): avoid cluster reset crash loop

xelab04:
Hi, I have looked into a potential fix for the cluster reset bug which I encountered in #844 

This fix removes the cluster reset flag before attempting a reset. It can happen that a stale flag remains if the previous cluster reset failed or was interrupted.
Additionally, it will output reset errors into stdout, which can make such problems easier to diagnose in the future.


galal-hussein (approved):

enrichman (approved):

enrichman:
Thanks @xelab04! I have rebased, run the tests, and merged.

Issue #1048, opened 2026-07-20
Configure DNS forwarder for Clusters in Shared Mode via Custom Resource Spec


dquilbd:
Hello K3K Development Team,
As part of an internal company evaluation, I am currently testing K3K (Rancher Virtual Clusters) in Shared Mode within an air-gapped environment. In this environment, we use internal DNS servers to resolve endpoints such as Rancher, and external public DNS servers are not reachable.
As I understand it, in Shared Mode the CoreDNS configuration is generated by the kubelet and uses public DNS forwarders (e.g., 8.8.8.8) as a fallback, while the DNS configuration of the worker node (e.g., /etc/resolv.conf) is not being used.
Would it be possible to configure custom DNS forwarders in the Custom Resource so that the controller can automatically add them to the CoreDNS configuration (Corefile)?

_Example yaml:_

```
apiVersion: k3k.io/v1beta1
kind: Cluster
metadata:
  name: test-shared-2
  namespace: k3k-clusters
spec:
  mode: shared
  customDNS:
    forwarders:
    - 10.20.30.40
```

Thank you in advance.
Best regards,
Daniel Quilitzsch

PR #1119, opened 2026-08-04
Allow Custom DNS configuration

bigkevmcd:
This adds support for configuring custom recursors in the virtual cluster coredns setup.

```yaml
apiVersion: k3k.io/v1beta1
kind: Cluster
metadata:
  name: mycluster
spec:
  expose:
    nodePort: {}
  mode: shared
  servers: 1
  customDNS:
    overrides:
      forwarders:
      - ips: ["8.8.8.8"]
```

https://github.com/rancher/k3k/issues/1048


galal-hussein (approved):
LGTM Thanks!

enrichman (commented):
Thanks @bigkevmcd for putting this together!

The current implementation handles configuring the CoreDNS forwarders, however I am concerned that exposing only specific, hardcoded keys makes the configuration a bit too rigid. If users need to customize other standard CoreDNS plugins in the future, such as rewrites, custom hosts, cache settings, or logging, we would be forced to make continuous API changes and codebase modifications to support them.

To make this more future-proof and align with standard Kubernetes patterns, I think it would be better to rely on a more flexible approach. We could either allow users to provide a reference to an external ConfigMap containing their custom CoreDNS configuration snippets, or expose a generic key-value array in the API to dynamically generate that ConfigMap directly from the k3k Custom Resource. These alternatives would provide a much better balance of flexibility and long-term maintainability without requiring schema updates for every new CoreDNS feature.

Here is how the two approaches could look in practice. The first option relies on an external ConfigMap reference. The user creates the ConfigMap, and the k3k cluster simply points to it:

```yaml
apiVersion: k3k.io/v1alpha1
kind: Cluster
metadata:
  name: my-k3k-cluster
spec:
  dns:
    coreDNS:
      customConfigRef:
        name: custom-coredns-snippets
```

Alternatively, we could use an inline array of files and their contents. The k3k controller would then take this array and translate it into the underlying ConfigMap automatically:

```yaml
apiVersion: k3k.io/v1alpha1
kind: Cluster
metadata:
  name: my-k3k-cluster
spec:
  dns:
    coreDNS:
      customConfig:
      - name: forward.conf
        value: |
          forward . 8.8.8.8 1.1.1.1
      - name: rewrite.conf
        value: |
          rewrite stop name regex (.*)\.old\.local {1}.new.local
```

What do you think?

/cc @galal-hussein

bigkevmcd:
Hey @enrichman thanks for weighing in.

I'm definitely open to changing this...so..let's see

> The current implementation handles configuring the CoreDNS forwarders, however I am concerned that exposing only specific, hardcoded keys makes the configuration a bit too rigid. If users need to customize other standard CoreDNS plugins in the future, such as rewrites, custom hosts, cache settings, or logging, we would be forced to make continuous API changes and codebase modifications to support them.

Yeah...this was deliberate, but maybe it's too focussed on usability.

> To make this more future-proof and align with standard Kubernetes patterns, I think it would be better to rely on a more flexible approach. We could either allow users to provide a reference to an external ConfigMap containing their custom CoreDNS configuration snippets, or expose a generic key-value array in the API to dynamically generate that ConfigMap directly from the k3k Custom Resource. These alternatives would provide a much better balance of flexibility and long-term maintainability without requiring schema updates for every new CoreDNS feature.

> Here is how the two approaches could look in practice. The first option relies on an external ConfigMap reference. The user creates the ConfigMap, and the k3k cluster simply points to it:

```yaml
apiVersion: k3k.io/v1alpha1
kind: Cluster
metadata:
  name: my-k3k-cluster
spec:
  dns:
    coreDNS:
      customConfigRef:
        name: custom-coredns-snippets
```        

Why not just rely on ConfigMap synchronisation?

The `coredns` in the virtual cluster will automatically mount a `ConfigMap` named `coredns-custom` and keys with `*.server` and `*.override` will be picked up automatically?

https://github.com/k3s-io/k3s/blob/main/manifests/coredns.yaml#L73-L77

We wouldn't need to do much if anything?

> Alternatively, we could use an inline array of files and their contents. The k3k controller would then take this array and translate it into the underlying ConfigMap automatically:

```yaml
apiVersion: k3k.io/v1alpha1
kind: Cluster
metadata:
  name: my-k3k-cluster
spec:
  dns:
    coreDNS:
      customConfig:
      - name: forward.conf
        value: |
          forward . 8.8.8.8 1.1.1.1
      - name: rewrite.conf
        value: |
          rewrite stop name regex (.*)\.old\.local {1}.new.local
```

Given that the key naming is significant, maybe we have `.override` and `.server` elements here and combine them. We can't do much to syntax check the config, but that's probably fine?

I'm not sure we even need this branch if we go down the "just use ConfigMap synchronisation route"?

enrichman:
I'm not completely sure the entire approach will work though. Thinking a little bit more I'm wondering if the ConfigMap, in order to be applied, needs to exists also on the virtual cluster.

Anyway I think we should probably rely on the inner spec approach. In this way we can override the config with a VirtualClusterPolicy (?).

bigkevmcd:
> I'm not completely sure the entire approach will work though. Thinking a little bit more I'm wondering if the ConfigMap, in order to be applied, needs to exists also on the virtual cluster.

The ConfigMap for the coredns in the virtual cluster does need to be in the virtual cluster - this change creates the ConfigMap in the virtual cluster for the coredns pod to mount (it's flagged as optional).

This would need to be expanded a bit?

```yaml
spec:
  dns:
    coreDNS:
      customConfig:
      - name: forward.conf
        value: |
          forward . 8.8.8.8 1.1.1.1
      - name: rewrite.conf
        value: |
          rewrite stop name regex (.*)\.old\.local {1}.new.local
```

We'd combine the `.spec.dns.coreDNS.customConfig` elements (the values) into a `ConfigMap` called `coredns-custom` in the `kube-system` namespace.

It's up to the user to name the elements as `.override` or `.server` depending on the behaviour they want.

This wouldn't be very hard to do, it's not quite clear what we'd do with the `VirtualClusterPolicy` tho'.

Does it overwrite known keys in the `.spec.dns.coreDNS.customConfig` or does it replace the `.spec.dns.coreDNS.customConfig`?

To be fair, we don't need to decide on this just now :-)

Let me implement this...and the policy part can get figured out later.

PR #1126, opened 2026-08-07
Allow customising coredns for virtual clusters.

bigkevmcd:
This allows providing custom snippets for configuring coredns in the virtualcluster via the mounted configmap mechanism supported by k3s coredns deployment.


bigkevmcd:
@enrichman I think this implements your suggestion?

pkg/apis/k3k.io/v1beta1/types.go
```diff
@@ -239,6 +239,29 @@ type ClusterSpec struct {
...
+// configuration in the virtual cluster.
+type CustomDNSConfig struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
+// CoreDNS provides the configuration for coredns.
+type CoreDNS struct {
+	CustomConfig []CustomDNSConfig `json:"customConfig"`
 }
```
enrichman, line 265: Naming is hard, so I'm a bit unsure about the `Custom` prefix. Since this is allowing more global (?) DNS configuration maybe we can use something like just `DNS`, or `DNSConfig`? And also the inner `CustomDNSConfig` could be a `CustomConfigMap`, since it's already inside the `CoreDNS` struct. Not sure about the `CustomConfigMap` or just `CustomConfig`.


```go
	// other ...

	// DNS provides custom configuration for coredns in the virtual
	// cluster.
	//
	// +optional
	DNS *DNSConfig `json:"dns,omitempty"`
}

// DNSConfig allows customising DNS configuration for the virtual cluster.
type DNSConfig struct {
	CoreDNS *CoreDNS `json:"coreDNS,omitempty"`
}

// CoreDNS provides the configuration for coredns.
type CoreDNS struct {
	CustomConfigMap []CustomConfigMap `json:"customConfigMap"`
}

// CustomConfigMap is a name/value field that is written as a key to the coredns
// configuration in the virtual cluster.
type CustomConfigMap struct {
	Name  string `json:"name"`
	Value string `json:"value"`
}
```

enrichman (commented):
Thanks @bigkevmcd! Yes, this was what I was thinking about.

I left a small comment related to the naming of the structs, and I'm just not sure about what is going to happen in two cases:

1) what if the ConfigMap is updated (i.e. manually changed) in the **host** cluster?

IMHO we should have owership of the resource, retrigger a reconciliation to recreate it as defined in the Cluster spec (not sure if that's the case).

2) what if the ConfigMap is updated in the **virtual** cluster?

IMHO we should still overwrite it. If we own the resource probably this is going to happen automatically in the same way.

Issue #1125, opened 2026-08-07
Issue with Projected service account token


jp-gouin:
When trying to deploy the following app
```
helm  upgrade --install test gitlab/gitlab-agent \                               
    --namespace gitlab-agent-test \
    --create-namespace \
    --set image.tag=v18.11.0
```

It fails with `ProviderFailed`
```
026-08-07T13:26:37.818Z    ERROR    Unable to transform tokens for pod    {"cluster": "k3k-test-gitlab", "namespace":
  "gitlab-agent-test", "name": "test-gitlab-agent-v2-54f7c84c49-bxmqw", "pod": "test-gitlab-agent-v2-54f7c84c49-bxmqw-gitlab-agent-test-k-a6e21", "error": "Secret
  \"k3k-test-gitlab-agent-https://kubernetes.default.svc.clus-144f9/\" is invalid: metadata.name: Invalid value: \"k3k-test-gitlab-agent-https://kubernetes.default.svc.clus-144f9/\": a lowercase RFC 1123 subdomain
  must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is
  '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')"}
```

The translated name is wrongly formatted and `https://kubernetes.default.svc.clus` is happened to the name

PR #1128, opened 2026-08-08
Fix projected volume token translation and different service accounts

galal-hussein:
The fix does two things:

- Removes the `tokenRequest.Audiances` from the token name since it can contain characters such as `:` or `/` that is rejected by Kubernetes API.
- Removes the `serviceAccountName` from the synced pod's spec in both projected token translation and automount token translation.

I also added e2e test case testing the projected service account token translation, and fixed the unit test case to account of the removal of the Audiances from the secret name.

- #1125 

Related fix
- #1103

### DIFF
--- a/k3k-kubelet/provider/token_test.go
+++ b/k3k-kubelet/provider/token_test.go
@@ -216,12 +216,12 @@ func Test_addKubeAccessVolume(t *testing.T) {
 }
 
 func Test_virtualSecret(t *testing.T) {
-	s := virtualSecret("my-secret", "my-ns", "my-sa")
+	s := virtualSecret("my-secret", "my-ns", "my-serviceaccount")
 
 	assert.Equal(t, "my-secret", s.Name)
 	assert.Equal(t, "my-ns", s.Namespace)
 	assert.Equal(t, corev1.SecretTypeServiceAccountToken, s.Type)
-	assert.Equal(t, "my-sa", s.Annotations[corev1.ServiceAccountNameKey])
+	assert.Equal(t, "my-serviceaccount", s.Annotations[corev1.ServiceAccountNameKey])
 	assert.Equal(t, "Secret", s.Kind)
 	assert.Equal(t, "v1", s.APIVersion)
 }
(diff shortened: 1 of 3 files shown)


tests/e2e/cluster_pod_test.go
```diff
@@ -546,4 +548,107 @@ var _ = Context("In a shared cluster", Label(podTestsLabel), Ordered, func() {
...
+						},
+					},
+				},
+			}
+
+			pod, err = virtualCluster.Client.CoreV1().Pods(pod.Namespace).Create(ctx, pod, metav1.CreateOptions{})
+			Expect(err).To(Not(HaveOccurred()))
+		})
+
+		It("should have translated projected serviceaccount token to secret", func() {
```
bigkevmcd:
```suggestion
		It("should have translated projected serviceaccount token to secret", func(ctx context.Context) {
```

And remove the created context.

tests/e2e/cluster_pod_test.go
```diff
@@ -546,4 +548,107 @@ var _ = Context("In a shared cluster", Label(podTestsLabel), Ordered, func() {
...
+
+	When("creating a Pod with a projected serviceaccount token", Label("test"), func() {
+		var (
+			pod                     *corev1.Pod
+			serviceAccountName      = "sa-nginx"
+			serviceAccountTokenExp  = int64(3600)
+			serviceAccountTokenPath = "token"
+		)
+
+		BeforeAll(func() {
```
bigkevmcd:
```suggestion
		BeforeAll(func(ctx context.Context) {
```

Drop the `ctx := context.Background()`

k3k-kubelet/provider/token_test.go
```diff
@@ -236,92 +235,53 @@ func Test_generateTokenSecretName(t *testing.T) {
...
-				Spec: authv1.TokenRequestSpec{
-					Audiences:         []string{"api"},
-					ExpirationSeconds: ptr.To(int64(3600)),
-				},
-			},
-			want: "k3k-my-sa-api-3600-token",
-		},
-		{
-			name:               "with multiple audiences and expiration",
+			name:               "with audience and expiration, audiance ignored",
```
bigkevmcd:
```suggestion
			name:               "with audience and expiration, audience ignored",
```

bigkevmcd: I think you can add an audience into all of the test cases and it should not be generated in any...

An audience is common (and the fact that we don't output it is true across all test cases).

enrichman (commented):
Thanks for the fix! I had a look and I have one doubt about dropping the audience, let me know if I'm missing something.

As far as I understand, the audience changes the token itself (it ends up in the `aud` claim), so if we remove it from the name, two projected tokens with the same SA, expiration and path but different audiences end up sharing one secret, and the second `CreateOrUpdate` overwrites the first. Vault and AWS IRSA both use `path: token`, so a pod using both would silently get the wrong token. The path, on the other hand, isn't part of the token, it's just the key in `Data`.

So I'd rather sanitize the audience than drop it. It also looks less disruptive: dropping it renames the secret for everyone who has an audience set, while sanitizing leaves already-valid values like `vault` untouched and only changes the ones that were failing anyway.

Also, I think `path` has the same problem: a projected path can be `certs/token`, which would fail exactly like #1125.

Basically something like this:

```go
// token.go

// invalidNameChars matches any run of characters that cannot appear in a DNS-1123 subdomain.
var invalidNameChars = regexp.MustCompile(`[^a-z0-9.-]+`)

func sanitizeNameComponent(s string) string {
	sanitized := strings.Trim(invalidNameChars.ReplaceAllString(strings.ToLower(s), "-"), "-.")
	if sanitized == s {
		return sanitized
	}

	digest := sha256.Sum256([]byte(s))

	return sanitized + "-" + hex.EncodeToString(digest[:])[:6]
}

func generateTokenSecretName(serviceAccountName, tokenPath string, tokenReq *authv1.TokenRequest) string {
	nameComponents := []string{serviceAccountName}

	for _, audience := range tokenReq.Spec.Audiences {
		nameComponents = append(nameComponents, sanitizeNameComponent(audience))
	}

	if exp := tokenReq.Spec.ExpirationSeconds; exp != nil {
		nameComponents = append(nameComponents, strconv.FormatInt(*exp, 10))
	}

	if tokenPath != "" {
		nameComponents = append(nameComponents, sanitizeNameComponent(tokenPath))
	}

	return k3kcontroller.SafeConcatNameWithPrefix(nameComponents...)
}
```

What do you think? @bigkevmcd any thought about this?

enrichman (approved):

bigkevmcd (approved):
Looks good!

PR #1129, opened 2026-08-08
Fail cluster delete when the cluster name is ambiguous

### LINKED ISSUE #448: Prevent deleting wrong cluster when names are ambiguous

imantaba:
## Problem

`k3kcli cluster delete NAME` resolves a bare name to the `k3k-<name>` namespace. When clusters with the same name exist in several namespaces, the command can act on one the user didn't mean to target.

Fixes #448.

## Solution

When no namespace is given (no `-n`, no `namespace/name` arg), look the name up across the namespaces the user can list:

- one match → use it
- several matches → error, asking for `--namespace/-n`
- no match → `k3k-<name>` as before

If the list call fails the `k3k-<name>` convention is kept as a fallback, so users with access to only their own namespace are unaffected.

Note this also changes the single-match case: `delete foo` now finds `default/foo` instead of erroring on `k3k-foo`. Happy to make that stricter if you'd rather only handle the collision.

## Testing

Unit tests in `cli/cmds/cluster_delete_test.go`:

- ambiguous name errors and deletes nothing
- single match resolves to its namespace
- `-n` flag and `namespace/name` arg skip the lookup
- clusters with a different name are ignored

This change was prepared with AI assistance; all changes were reviewed and verified by me.

### DIFF
--- a/cli/cmds/cluster_delete.go
+++ b/cli/cmds/cluster_delete.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/sirupsen/logrus"
@@ -81,7 +82,7 @@ func delete(appCtx *AppContext) func(cmd *cobra.Command, args []string) error {
 			return errors.New("expected exactly one cluster name")
 		}
 
-		namespace, name, err := resolveClusterArg(appCtx, args[0])
+		namespace, name, err := resolveClusterArg(ctx, appCtx, args[0])
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


enrichman:
Hi @imantaba, thanks for the PR! It looks like I forgot to link/mention an existing PR ([#1080](https://github.com/rancher/k3k/pull/1080)) in the original issue. Since that one already addresses it, I've moved #448 to "To Test" and will be closing this PR.

Thanks again for contributing! We definitely need to update the `CONTRIBUTING.md` guide to simplify the contribution process and avoid this kind of misalignment in the future (e.g., requiring issues to be assigned before starting work). We appreciate your effort! :slightly_smiling_face:

PR #1141, opened 2026-08-12
Add e2e tests for external datastores

galal-hussein:
- https://github.com/rancher/k3k/issues/304

### DIFF
--- a/tests/e2e/cluster_datastore_test.go
+++ b/tests/e2e/cluster_datastore_test.go
@@ -1,0 +1,308 @@
+package k3k_test
+
+import (
+	"fmt"
+	"strconv"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/rancher/k3k/pkg/apis/k3k.io/v1beta1"
+	fwk3k "github.com/rancher/k3k/tests/framework/k3k"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = When("creating a shared mode cluster with postgres datastore via server args", Ordered, Label(datastoreTestsLabel), Label(slowTestsLabel), func() {
+	var virtualCluster *VirtualCluster
+
+	BeforeAll(func() {
+		namespace := fwk3k.CreateNamespace(k8s)
+
+		DeferCleanup(func() {
+			fwk3k.DeleteNamespaces(k8s, namespace.Name)
+		})
+		postgresEndpoint := fmt.Sprintf("postgres-k3k.%s.svc.cluster.local:5432", namespace.Name)
+
+		cluster := NewCluster(namespace.Name, func(c *v1beta1.Cluster) {
+			c.Spec.ServerArgs = []string{
+				fmt.Sprintf("--datastore-endpoint=postgres://%s:%s@%s/%s?sslmode=disable", postgresUser, postgresPassword, postgresEndpoint, postgresDatabase),
+				"--cluster-init=false",
+			}
+		})
+
+		deployPostgresInCluster(cluster)
+
+		CreateCluster(cluster)
+		client, restConfig := NewVirtualK8sClientAndConfig(cluster)
+
+		virtualCluster = &VirtualCluster{
+			Cluster:    cluster,
+			RestConfig: restConfig,
+			Client:     client,
+		}
+	})
+
+	It("creates and writes to a database for the cluster", func() {
+		Eventually(func(g Gomega) {
+			ctx := GinkgoT().Context()
+
+			count, err := queryPostgres(ctx, virtualCluster.Cluster.Namespace, "SELECT count(*) FROM kine WHERE name LIKE '/registry/namespaces/%'")
+			g.Expect(err).To(Not(HaveOccurred()))
+
+			g.Expect(strconv.Atoi(count)).To(BeNumerically(">", 0))
+		}).
+			WithTimeout(time.Minute).
+			WithPolling(time.Second).
+			Should(Succeed())
+	})
+	It("creates server pods with no etcd finalizers", func() {
+		Eventually(func(g Gomega) {
+			ctx := GinkgoT().Context()
+
+			serverPods := listServerPods(ctx, virtualCluster)
+			for _, s := range serverPods {
+				g.Expect(s.Finalizers).To(BeNil())
+			}
+		}).
+			WithTimeout(time.Minute).
+			WithPolling(time.Second).
+			Should(Succeed())
+	})
+	It("can create a nginx pod", func() {
+		_, _ = virtualCluster.NewNginxPod("")
+	})
+})
+
+var _ = When("creating a shared mode cluster with postgres datastore via drop-in config file", Ordered, Label(datastoreTestsLabel), Label(slowTestsLabel), func() {
+	var virtualCluster *VirtualCluster
+
+	BeforeAll(func() {
+		ctx := GinkgoT().Context()
+
+		namespace := fwk3k.CreateNamespace(k8s)
+
+		DeferCleanup(func() {
+			fwk3k.DeleteNamespaces(k8s, namespace.Name)
+		})
+
+		configSecretName := "datastore-config-secret"
+		configSecret := corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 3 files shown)


tests/e2e/cluster_datastore_test.go
```diff
@@ -0,0 +1,309 @@
...
+		})
+
+		postgresEndpoint := fmt.Sprintf("postgres-k3k.%s.svc.cluster.local:5432", namespace.Name)
+
+		cluster := NewCluster(namespace.Name, func(c *v1beta1.Cluster) {
+			c.Spec.ServerArgs = []string{
+				fmt.Sprintf("--datastore-endpoint=postgres://%s:%s@%s/%s?sslmode=disable", postgresUser, postgresPassword, postgresEndpoint, postgresDatabase),
+				"--cluster-init=false",
+			}
+		})
```
enrichman, line 44: I remember a bug I had in the past about complex passwords with `@` or `/` chars. This should work the same:

```suggestion
		postgresEndpoint := fmt.Sprintf("postgres-k3k.%s.svc.cluster.local:5432", namespace.Name)

		dsn := url.URL{
				Scheme:   "postgres",
				User:     url.UserPassword(postgresUser, postgresPassword),
				Host:     postgresEndpoint,
				Path:     postgresDatabase,
				RawQuery: "sslmode=disable",
		}

		cluster := NewCluster(namespace.Name, func(c *v1beta1.Cluster) {
			c.Spec.ServerArgs = []string{
				fmt.Sprintf("--datastore-endpoint=%s", dsn.String()),
				"--cluster-init=false",
			}
		})
```

tests/e2e/cluster_datastore_test.go
```diff
@@ -0,0 +1,309 @@
...
```
enrichman, line 199: I think we should set the `mode` as virtual here.

tests/e2e/cluster_datastore_test.go
```diff
@@ -0,0 +1,309 @@
...
+
+		DeferCleanup(func() {
+			fwk3k.DeleteNamespaces(k8s, namespace.Name)
+		})
+
+		postgresEndpoint := fmt.Sprintf("postgres-k3k.%s.svc.cluster.local:5432", namespace.Name)
+
+		cluster := NewCluster(namespace.Name, func(c *v1beta1.Cluster) {
+			c.Spec.ServerArgs = []string{
+				fmt.Sprintf("--datastore-endpoint=postgres://%s:%s@%s/%s?sslmode=disable", postgresUser, postgresPassword, postgresEndpoint, postgresDatabase),
```
enrichman: As before we should create the DSN in the same (safer) way.

tests/e2e/cluster_datastore_test.go
```diff
@@ -0,0 +1,309 @@
...
+			serverPods := listServerPods(ctx, virtualCluster)
+
+			for _, s := range serverPods {
+				g.Expect(s.Finalizers).To(BeNil())
+			}
+		}).
+			WithTimeout(time.Minute).
+			WithPolling(time.Second).
+			Should(Succeed())
+	})
```
enrichman, line 168: `Eventually` passes on the first poll, so this can't catch a finalizer that gets added shortly after the pod becomes Ready:

```suggestion
	It("creates server pods with no etcd finalizers", func() {
		Consistently(func(g Gomega) {
			serverPods := listServerPods(GinkgoT().Context(), virtualCluster)
			for _, s := range serverPods {
				g.Expect(s.Finalizers).To(BeEmpty())
			}
		}).
			WithTimeout(time.Second * 30).
			WithPolling(time.Second * 2).
			Should(Succeed())
	})
```

tests/e2e/cluster_datastore_test.go
```diff
@@ -0,0 +1,309 @@
...
+
+		configSecretName := "datastore-config-secret"
+		configSecret := corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      configSecretName,
+				Namespace: namespace.Name,
+			},
+			Data: map[string][]byte{
+				"config.yaml": []byte(fmt.Sprintf(
+					"datastore-endpoint: postgres://%s:%s@postgres-k3k.%s.svc.cluster.local:%d/%s?sslmode=disable\ncluster-init: false\n",
```
enrichman: A datastore endpoint could break the yaml if contains the `#` or `:` chars, so it's better to quote it. Having it as an external var would be easier:

```suggestion
					"datastore-endpoint: %q\ncluster-init: false\n",
```

enrichman (commented):
A few comments. We can probably have an helper func to create the DSN URI.

bigkevmcd (commented):
As @enrichman says this could be tricky with some passwords...

https://github.com/lib/pq/blob/main/connector.go#L1143 might give you some ideas on generating the string.

galal-hussein:
@enrichman @bigkevmcd I have added the first suggestion for using `url.URL` for generating the DSN, however I do not think we should not create a separate helper for generating the DSN since this should be tested by k3s->kine, I think the main functionality that we need to test from k3k side is that k3k cluster honors `--datastore-endpoint` whether its passed through serverArgs or config file and it skips the etcd member removal and thats guranateed by checking the finalizers

tests/e2e/cluster_datastore_test.go
```diff
@@ -0,0 +1,328 @@
...
+			Scheme:   "postgres",
+			User:     url.UserPassword(postgresUser, postgresPassword),
+			Host:     postgresEndpoint,
+			Path:     postgresDatabase,
+			RawQuery: "sslmode=disable",
+		}
+
+		cluster := NewCluster(namespace.Name, func(c *v1beta1.Cluster) {
+			c.Spec.ServerArgs = []string{
+				fmt.Sprintf("--datastore-endpoint=%s", dsn.String()),
```
enrichman:
```suggestion
				fmt.Sprintf("--datastore-endpoint=%q", dsn.String()),
```

tests/e2e/cluster_datastore_test.go
```diff
@@ -0,0 +1,328 @@
...
+		}
+
+		configSecretName := "datastore-config-secret"
+		configSecret := corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      configSecretName,
+				Namespace: namespace.Name,
+			},
+			Data: map[string][]byte{
+				"config.yaml": []byte(fmt.Sprintf("datastore-endpoint: %s\ncluster-init: false\n", dsn.String())),
```
enrichman:
```suggestion
				"config.yaml": []byte(fmt.Sprintf("datastore-endpoint: %q\ncluster-init: false\n", dsn.String())),
```

tests/e2e/cluster_datastore_test.go
```diff
@@ -0,0 +1,328 @@
...
+			Scheme:   "postgres",
+			User:     url.UserPassword(postgresUser, postgresPassword),
+			Host:     postgresEndpoint,
+			Path:     postgresDatabase,
+			RawQuery: "sslmode=disable",
+		}
+
+		cluster := NewCluster(namespace.Name, func(c *v1beta1.Cluster) {
+			c.Spec.ServerArgs = []string{
+				fmt.Sprintf("--datastore-endpoint=%s", dsn.String()),
```
enrichman:
```suggestion
				fmt.Sprintf("--datastore-endpoint=%q", dsn.String()),
```

enrichman:
```suggestion
				fmt.Sprintf("--datastore-endpoint=%q", dsn.String()),
```

tests/e2e/cluster_datastore_test.go
```diff
@@ -0,0 +1,328 @@
...
+		}
+
+		configSecretName := "datastore-config-secret"
+		configSecret := corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      configSecretName,
+				Namespace: namespace.Name,
+			},
+			Data: map[string][]byte{
+				"config.yaml": []byte(fmt.Sprintf("datastore-endpoint: %s\ncluster-init: false\n", dsn.String())),
```
enrichman:
```suggestion
				"config.yaml": []byte(fmt.Sprintf("datastore-endpoint: %q\ncluster-init: false\n", dsn.String())),
```

enrichman (commented):
A couple of small comments, then lgtm. 🙂

enrichman (approved):

PR #1147, opened 2026-08-13
Enhance ingress exposure validation

### LINKED ISSUE #1135: HCP mode cluster provisioning fails on expose: ingress when using Traefik
**Environmental Info:**
Host Cluster Configuration:
Downstream RKE2 cluster k8s v1.36.2+rke2r1 (AWS node provisioned) using Traefik (as RKE2 does by default).

K3K Cluster Configuration:
k3k chart 1.2.0-rc3
k3kcli 1.2.0-rc3

**Describe the bug:**
When configuring a cluster with `expose: ingress`, the K3k controller encounters two key issues:

**Scenario 1: Validation Failure on Missing Host Rule:** If `tlsSANs` is omitted, K3k generates an invalid `Ingress` object with empty `rules` and `defaultBackend`, causing the Kubernetes API server to reject the object with:
   `Ingress.networking.k8s.io "..." is invalid: spec: Invalid value: null: either defaultBackend or rules must be specified`

**Scenario 2: Missing TLS Passthrough Configuration:** When `tlsSANs` is provided, K3k successfully creates a standard `networking.k8s.io/v1` `Ingress` object, but leaves `Annotations` completely empty and defaults to HTTP/80 routing. Because the K3s API server requires raw mTLS client certificates, terminating TLS at the Ingress layer breaks client/agent authentication. Furthermore, standard `Ingress` resources cannot perform Layer 4 TLS passthrough on Traefik (the default Ingress controller in K3s/RKE2), which requires `IngressRouteTCP` CRDs.

**Steps To Reproduce:**
**For Scenario 1**
- Use the following yaml to create an HCP mode k3k cluster with expose: ingress omitting tlsSANs:
```
apiVersion: k3k.io/v1beta1
   kind: Cluster
   metadata:
     name: k3k-ingress-test
     namespace: k3k-system
   spec:
     mode: hcp
     servers: 1
     expose:
       ingress: {}
```
- Check the generated ingress resource on the host cluster and verify it failed with the following warning
```
kubectl describe cluster.k3k.io k3k-ingress-test -n k3k-system

Status:
  Cluster CIDR:  10.52.0.0/16
  Conditions:
    Last Transition Time:  2026-08-11T21:32:27Z
    Message:               Ingress.networking.k8s.io "k3k-k3k-ingress-test-ingress" is invalid: spec: Invalid value: null: either `defaultBackend` or `rules` must be specified
    Reason:                ProvisioningFailed
    Status:                False
    Type:                  Ready
  Host Version:            v1.36.3
  Phase:                   Failed
  Service CIDR:            10.53.0.0/16
  Tls SANs:
    10.43.77.144
    k3k-k3k-ingress-test-service
    k3k-k3k-ingress-test-service.k3k-system
Events:
  Type     Reason              Age                   From                    Message
  ----     ------              ----                  ----                    -------
  Warning  ProvisioningFailed  7m43s (x16 over 18m)  k3k-cluster-controller  Ingress.networking.k8s.io "k3k-k3k-ingress-test-ingress" is invalid: spec: Invalid value: null: either `defaultBackend` or `rules` must be specified
```

**For scenario 2**
- Use the following yaml to create an HCP mode k3k cluster with expose: ingress with tlsSANs set:
```
apiVersion: k3k.io/v1beta1
   kind: Cluster
   metadata:
     name: k3k-ingress-test
     namespace: k3k-system
   spec:
     mode: hcp
     servers: 1
     tlsSANs:
       - "k3k-api.test.local"
     expose:
       ingress: {}
```
- Check the generated ingress resource on the host cluster and observe `Spec.Port` is set to route traffic to port 80 and `Spec.Annotation` is empty:
```
kubectl describe ingress -n k3k-system k3k-k3k-ingress-test-ingress

Port: 80
Annotations: <none>
```

**Expected behavior:**
K3k should support injecting custom annotations into the generated Ingress object (e.g., nginx.ingress.kubernetes.io/ssl-passthrough: "true") to allow Layer 4 TLS passthrough where supported.

K3k should validate required host parameters prior to building the Ingress spec to avoid sending invalid manifests with null rules to the Kubernetes API server.

Documentation should clarify Ingress controller compatibility requirements (e.g., Nginx with SSL Passthrough enabled vs. Traefik limitations with standard Ingress resources).

**Actual behavior:**

The generated Ingress resource has no annotations (Annotations: <none>) and routes traffic via HTTP (port 80). 
If `tlsSANs` is omitted entirely from the spec, cluster creation fails instantly with a ProvisioningFailed warning due to invalid empty rules in the generated Ingress spec.

enrichman:
Fixes: https://github.com/rancher/k3k/issues/1135

If `tlsSANs` is omitted, K3k generates an invalid Ingress object. This PR adds a validation when the `ingress` expose is specified, filtering out IP addresses with a new `controller.FilterDNSNames` utility func.

It adds also some docs, that will need to be added in the [rancher/k3k-product-docs](https://github.com/rancher/k3k-product-docs) repo.

### DIFF
--- a/pkg/controller/cluster/cluster_test.go
+++ b/pkg/controller/cluster/cluster_test.go
@@ -1,0 +1,128 @@
+package cluster
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/rancher/k3k/pkg/apis/k3k.io/v1beta1"
+)
+
+func Test_validate(t *testing.T) {
+	tests := []struct {
+		name        string
+		clusterName string
+		mode        v1beta1.ClusterMode
+		tlsSANs     []string
+		expose      *v1beta1.ExposeConfig
+		policy      *v1beta1.VirtualClusterPolicy
+		wantErr     string
+	}{
+		{
+			name: "valid cluster without a policy",
+		},
+		{
+			name:   "valid cluster with a policy",
+			policy: newTestPolicy(v1beta1.SharedClusterMode),
+		},
+		{
+			// the name check does not depend on the policy, so it also runs
+			// in namespaces that are not bound to one
+			name:        "invalid cluster name without a policy",
+			clusterName: ClusterInvalidName,
+			wantErr:     "invalid cluster name",
+		},
+		{
+			name:        "invalid cluster name with a policy",
+			clusterName: ClusterInvalidName,
+			policy:      newTestPolicy(v1beta1.SharedClusterMode),
+			wantErr:     "invalid cluster name",
+		},
+		{
+			name:    "mode not allowed by the policy",
+			mode:    v1beta1.VirtualClusterMode,
+			policy:  newTestPolicy(v1beta1.SharedClusterMode),
+			wantErr: "is not allowed by the policy",
+		},
+		{
+			// without a policy there is no allowed mode to check against
+			name: "any mode is allowed without a policy",
+			mode: v1beta1.VirtualClusterMode,
+		},
+		{
+			name:   "expose without ingress",
+			expose: &v1beta1.ExposeConfig{NodePort: &v1beta1.NodePortConfig{}},
+		},
+		{
+			name:    "expose ingress without tlsSANs",
+			expose:  &v1beta1.ExposeConfig{Ingress: &v1beta1.IngressConfig{}},
+			wantErr: "spec.tlsSANs",
+		},
+		{
+			name:    "expose ingress with only IP tlsSANs",
+			tlsSANs: []string{"10.0.0.5", "::1"},
+			expose:  &v1beta1.ExposeConfig{Ingress: &v1beta1.IngressConfig{}},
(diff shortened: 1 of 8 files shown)


pkg/controller/cluster/cluster_test.go
```diff
@@ -0,0 +1,128 @@
...
+			name: "valid cluster without a policy",
+		},
+		{
+			name:   "valid cluster with a policy",
+			policy: newTestPolicy(v1beta1.SharedClusterMode),
+		},
+		{
+			// the name check does not depend on the policy, so it also runs
+			// in namespaces that are not bound to one
+			name:        "invalid cluster name without a policy",
```
galal-hussein, line 34: I am fine with the policy changes, but do we need it in the same PR?

galal-hussein (approved):
LGTM, one nit about the policy changes

pkg/controller/cluster/cluster_test.go
enrichman: There is no change about the functionality, but this test was generated automatically, and it's covering both the cases. We can keep it, or remove it. It doesn't really add much!

PR #1151, opened 2026-08-17
feat: import CI images directly into containerd instead of ttl.sh

enrichman:
Since we have intermittent issues with [ttl.sh](ttl.sh), this PR drops the external registry entirely, to improve reliability, privacy and speed of the tests.

Instead of pushing images and pulling them back, CI now builds them and imports them straight into each node's containerd (make save + k3s ctr images import).

### DIFF
--- a/.github/workflows/test-conformance-hcp.yaml
+++ b/.github/workflows/test-conformance-hcp.yaml
@@ -94,12 +94,16 @@ jobs:
     - name: Setup K3k (from source)
       if: inputs.k3k_version == ''
       run: |
-        export REPO=ttl.sh/$(uuidgen)
-        export VERSION=1h
+        export REPO=k3k.local
 
         make build
         make package
-        make push
+        make save
+
+        # Import straight into the node's containerd
+        sudo k3s ctr images import k3k-images.tar
+        sudo k3s crictl images | grep k3k
+
         make install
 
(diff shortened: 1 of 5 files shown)


.github/workflows/test-conformance-hcp.yaml
```diff
@@ -94,12 +94,16 @@ jobs:
...
+        export REPO=k3k.local
 
         make build
         make package
-        make push
+        make save
+
+        # Import straight into the node's containerd
+        sudo k3s ctr images import k3k-images.tar
+        sudo k3s crictl images | grep k3k
```
galal-hussein: I think we can avoid the list/grep of the images.

A suggestion also is to have this in the Makefile, this way we can reuse it locally, for example:

CTRBIN=<> NAMESPACE=<> RUNTIME_ENDPOINT=<> make import

This way it can work with k3s or rke2 locally or even any containerd supported distro:
```bash
sudo /var/lib/rancher/rke2/bin/ctr -n k8s.io -a /run/k3s/containerd/containerd.sock images import -
```

enrichman: As discussed removed grep, but keeping the rest! :slightly_smiling_face:

galal-hussein (approved):

PR #1158, opened 2026-08-19
feat: update kubeconfig generate command

### LINKED ISSUE #474: Add autocompletion for `k3kcli kubeconfig generate`
Similar to `k3kcli cluster delete`, the `kubeconfig generate` command should be enhanced with shell autocompletion. We will adopt the `namespace/cluster-name` format for specifying the target cluster, improving consistency across the CLI.

The existing `--namespace` and `--name` flags can be kept for backward compatibility but should/could be marked as deprecated in the help text.

```
# User types this and presses <TAB>
$ k3kcli kubeconfig generate myns/c

# CLI suggests completions
myns/cluster-1   myns/cluster-2
```

### Acceptance Criteria
- [ ] The kubeconfig generate command accepts a single argument in the format `<namespace>/<cluster-name>`.
- [ ] Shell autocompletion is implemented to suggest available clusters.
- [ ] The `--namespace` and `--cluster-name` flags are marked as deprecated (or removed?). The command logic prioritizes the new argument if both flags and an argument are provided.

enrichman:
This PR aligns the k3kcli kubeconfig generate command to the other commands we have.

It deprecates the `--name` flag, but keeping it, so it will still work. The user can use the plain NAME, and it will resolve to the `k3k-<CLUSTER_NAME>` namespace, or if provided it will use the right namespace. It also supports the autocompletion for namespaces, or arguments with the `namespace/cluster_name` format, so the behavior is now consistent.

```
# plain name -> resolved to the k3k-mycluster namespace
k3kcli kubeconfig generate mycluster

# with an explicit namespace
k3kcli kubeconfig generate -n mynamespace mycluster

# with the namespace/cluster_name format
k3kcli kubeconfig generate mynamespace/mycluster

# autocompletion
k3kcli kubeconfig generate <TAB>
k3k-mycluster/mycluster   mynamespace/othercluster

k3kcli kubeconfig generate -n mynamespace <TAB>
mynamespace/othercluster

# the deprecated flag still works
k3kcli kubeconfig generate --name mycluster
Flag --name has been deprecated, it will be removed in a future release, use the NAME argument instead
WARN[0000] the --name flag is deprecated and will be removed in a future release, use the NAME argument instead

# if both are provided the argument wins
k3kcli kubeconfig generate --name ignored mynamespace/mycluster
WARN[0000] the --name flag is deprecated and will be removed in a future release, ignoring it in favor of the 'mynamespace/mycluster' argument
```

### DIFF
--- a/cli/cmds/kubeconfig_test.go
+++ b/cli/cmds/kubeconfig_test.go
@@ -1,0 +1,87 @@
+package cmds
+
+import (
+	"testing"
+
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/rancher/k3k/pkg/apis/k3k.io/v1beta1"
+)
+
+// Test_generate_clusterKey checks how the cluster name and namespace are resolved from the
+// NAME argument and the deprecated --name flag. The cluster is always missing, so the
+// "not found" error reports the resolved key without going through the kubeconfig generation.
+func Test_generate_clusterKey(t *testing.T) {
+	tests := []struct {
+		name          string
+		flagNamespace string
+		flagName      string
+		args          []string
+		wantErr       string
+	}{
+		{
+			name:    "bare name argument defaults to k3k-<name>",
+			args:    []string{"mycluster"},
+			wantErr: `cluster "mycluster" not found in namespace "k3k-mycluster"`,
+		},
+		{
+			name:    "namespace/name argument is split",
+			args:    []string{"myns/mycluster"},
+			wantErr: `cluster "mycluster" not found in namespace "myns"`,
+		},
+		{
+			name:          "bare name argument respects the namespace flag",
+			flagNamespace: "myns",
+			args:          []string{"mycluster"},
+			wantErr:       `cluster "mycluster" not found in namespace "myns"`,
+		},
+		{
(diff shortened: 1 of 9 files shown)


cli/cmds/kubeconfig_test.go
```diff
@@ -0,0 +1,111 @@
...
+		flagNamespace string
+		flagName      string
+		args          []string
+		wantErr       string
+		wantWarn      string
+	}{
+		{
+			name:    "bare name argument defaults to k3k-<name>",
+			args:    []string{"mycluster"},
+			wantErr: `cluster "mycluster" not found in namespace "k3k-mycluster"`,
```
galal-hussein, line 32: nit: should we use the `metav1.StatusReasonNotFound` instead of plain text to indicate that the cluster  object is not found

enrichman: Hm, not sure how. The `metav1.StatusReasonNotFound` is a plain string:

```go
StatusReasonNotFound StatusReason = "NotFound"
```

and this error is the one that gets emitted from the cli:

```
➜ k3kcli kubeconfig generate notfound
Error: cluster "notfound" not found in namespace "k3k-notfound"
FATA[0000] cluster "notfound" not found in namespace "k3k-notfound"
```

galal-hussein: yeah apologies, I thought we just emitting the error, but we do look for the notfound error and just print a message which is odd:
```
				return fmt.Errorf("cluster %s not found in namespace %s", name, appCtx.namespace)
			}
```
We mayneed to fix that later, to allow the errors emmitted from the cli to be checkable by other tools

galal-hussein (approved):

PR #1161, opened 2026-08-19
Add gomodcheck script to compare `go.mod` dependencies

enrichman:
This PR adds a small script to check the common deps we have in our two `go.mod` files will not diverge. I.e. currently we had two different indirect:

```
➜ make check-gomod 
go run scripts/gomodcheck.go go.mod pkg/apis/go.mod
MISMATCH  github.com/go-logr/logr
  go.mod           v1.4.4
  pkg/apis/go.mod  v1.4.3

MISMATCH  k8s.io/utils
  go.mod           v0.0.0-20260319190234-28399d86e0b5
  pkg/apis/go.mod  v0.0.0-20260210185600-b8788abfbbc2

2 mismatches across 18 common modules
exit status 1
make: *** [Makefile:148: check-gomod] Error 1
```

Fixing also that.

```
➜ make check-gomod
go run scripts/gomodcheck.go go.mod pkg/apis/go.mod
all 18 common modules match
```


bigkevmcd (approved):
Looks ok to me, one question...what happens if there's a replace directive?

Won't it use that value instead of the parsed require version?

Would that lead to the wrong answer?

enrichman:
> Looks ok to me, one question...what happens if there's a replace directive?
> 
> Won't it use that value instead of the parsed require version?
> 
> Would that lead to the wrong answer?

Ah, that's a very good question. Yes, the script is just going to check the current go.mod requires, so it's not going to resolve the replaces. Currently we don't have any, but it's worth to keep this in mind. The main goal was to keep the same deps of the pkg/apis/go.mod file as the one we currently have in our main go.mod.

PR #1242, opened 2026-09-10
remove empty dir filtering on kata

bcraft:
This is no longer required since the instructions where updated to use block-plain emptydir volumes. The SRA build has excluded this filter for some time, and inclusion causes issues with CoCo.


enrichman (commented):
Having a look at the code it seems the FilterEmptyDirVolumes func is not used anymore. We can drop it completely.

galal-hussein (approved):

galal-hussein:
> Having a look at the code it seems the FilterEmptyDirVolumes func is not used anymore. We can drop it completely.

Yeah I agree I think we can remove the function itself

PR #1245, opened 2026-09-10
Removed some revive rules

enrichman:
This PR removes `govet` from the enabled list (since it's enabled by default) and removes some `revive` rules, which fixes the following issues:

The rules now enabled are:
- `early-return`
- `enforce-switch-style`
- `unnecessary-format`
- `unused-receiver`

The `cyclomatic` rule was also enabled, but with a higher threshold (30, compared to the default of 10).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,7 +6,6 @@ linters:
   - wsl_v5
   - importas
   - revive
-  - govet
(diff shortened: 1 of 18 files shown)


galal-hussein (approved):
LGTM, Thanks!

